PR #3, opened 2019-11-23
add ability to specify hue in all colorization commands

kyleclo:
shouldnt break anything; prespecify hue in all the coloriztion commands; defaults to generator if None


data-processing/explanations/colorize_tex.py
```diff
@@ -145,7 +146,8 @@ def colorize_citations(tex: str) -> Iterator[CitationColorizationBatch]:
             hue_generator = generate_hues()
 
             # And get the hue for the next entity.
-            hue = next(hue_generator)
+            if hue is None:
```
andrewhead: Feeling like a Python 3 newbie here. Can you verify for me that hue won't be 'None' when the generator yields StopIterator? I'm pretty sure that's not the case, but I just want to rule that potential bug.

kyleclo: nice catch; that would be a bug, check latest push

data-processing/explanations/colorize_tex.py
```diff
@@ -145,7 +148,10 @@ def colorize_citations(tex: str) -> Iterator[CitationColorizationBatch]:
             hue_generator = generate_hues()
 
             # And get the hue for the next entity.
-            hue = next(hue_generator)
+            if preset_hue is None:
+                hue = next(hue_generator)
+            else:
+                hue = preset_hue
```
andrewhead: Technically this condition isn't necessary as this exception will never be reached if preset_hue is reached. Just wanted to point that out but I'm good with it there if you want it there for explicitness

PR #4, opened 2019-12-01
missing dataclasses dependency

kyleclo:
found this when working with Dockerfile

### DIFF
--- a/data-processing/Dockerfile
+++ b/data-processing/Dockerfile
@@ -63,7 +63,7 @@ WORKDIR /data-processing
 COPY . .
 
 # Set default Python to Python 3
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 1
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
 
 # Set up path to include Python scripts
 ENV PYTHONPATH="/data-processing"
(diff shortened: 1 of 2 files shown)


andrewhead:
@kyleclo cool! Isn't dataclasses available by default with Python 3.7 without a pip install?

kyleclo:
Ah I didn't look too deeply into it; I just built the Docker image and tried running steps & this failed for me

data-processing/Dockerfile
```diff
@@ -63,7 +63,7 @@ WORKDIR /data-processing
 COPY . .
 
 # Set default Python to Python 3
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 1
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 1
```
andrewhead: @kyleclo can you try re-running the Docker build with this fix? It appears that Ubuntu had Python 3.6 installed by default, and the system Python was getting set to Python 3.6 instead of 3.7. This should fix that.

andrewhead:
Hey @kyleclo ! Is this solved, is it okay to close it?

andrewhead:
Closing as inactive

PR #6, opened 2019-12-11
Optimize 🚀

codeviking:
This modifies the UI so that we use production grade PDF.js
assets. To accomplish this I:

- Removed the submodule, in favor of committing the PDF.js
  build artifacts into the repository.
- Removed the steps we were taking in `Docker` to support
  the submodule mechanism.
- Add basic instructions for downloading the PDF.js fork
  we rely upon and executing the build.
- Remove the `pdfjs-dist` package, as it downloads a
  precompiled PDF.js distribution, in favor of
  `@types/pdfjs-dist`. This provides us with the
  typing information without downloading the prebuilt
  distribution.
- Update a few JavaScript dependencies to resolve a few
  security alerts flagged by `npm`.
- Update the build so that we properly cache from both
  build targets. This should speed up builds considerably.

In testing this locally, things seem *much* faster. The resources
are roughly 50% the size, and there's considerably less files
to download.  That said I didn't measure the precise speed up
(though we can do so if we think it'd be valuable).

The upsides of this change are that it simplifies the build considerably,
, removes the submodule mechanism, and speeds up the UI significantly. The 
downsides are that it's more complicated to make edits to the PDF.js fork
that we source the code from.  That said we haven't made a lot of edits to
that fork, so it feels like the right trade off. Let me know what you think.


codeviking:
I just noticed that this is net +27k lines.  This largely has to due with the static assets that are being added.

I'm going to investigate tomorrow if I can remove the localization related files (as well as the cmaps).  I tried removing them but ran into issues -- I think I'll have to modify the PDF.js source code to remove anything that relies upon them.

codeviking:
Closing in favor of an alternative approach to come shortly. Stay tuned!

PR #7, opened 2019-12-11
Use a precompiled, optimized version of PDF.js

codeviking:
This modifies the repository so that the build artifacts from
PDF.js are committed to `public/`. This simplifies the build
process, and makes it easy to use an optimized version of
PDF.js.

This should provide a significant performance boost. I've also
included instructions and a script for automating the process
of updating these files.


ui/README.md
```diff
@@ -34,19 +24,51 @@ PORT=3001 npm start  # 3001 is the port number. Save port 3000
...
+rendered. You can regenerate this code and overwrite the copies
+in `public/` by executing these commands:
+
+```
+git submodule init
+git submodule update
+bash bin/build_pdf.js.sh
+```
+
+To apply the effect, commit the resulting changes.
```
andrewhead: What do you mean by "apply the effect"?

ui/bin/build_pdf.js.sh
```diff
@@ -0,0 +1,25 @@
...
+cd "$pdfjs_dir"
+npm install
+
+ls
+npm run build
+
+rsync -rv --progress \
+    --exclude=locale/ \
+    --exclude=cmaps/ \
+    --exclude=compressed.tracemonkey-pldi-09.pdf \
```
andrewhead: good catch

ui/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "moody",
   "version": "0.1.0",
   "private": true,
-  "proxy": "http://localhost:3000",
+  "proxy": "https://s2-reader.apps.allenai.org",
```
andrewhead: What's the reason for this change? Will this impact our ability to test the dev version of the UI with the dev version of the API?

ui/README.md
codeviking: Running the script updates the files in `public/`. This is just noting that the user needs to commit them.

That said I think that's implicit. Once the operator executes `git status` the impact of `bin/build_pdf.js.sh` will be clear.

I'll just omit this from the docs.

ui/package.json
codeviking: Ah, this just makes things work "out of the box".  

If a user wants to run against a local API they simply need to edit this value (and remember not to commit it).

That said I think you're right that it's a bit kludgy as is. Let me see if I can make this a bit easier.  Ideally the default would be this (so that it works), but it'd be possible for a user to set an environment variable or config bit to override this for local execution (i.e. so they don't have to worry about not committing the effect).

andrewhead: If there's an easy solution with environment variables then yeah, that sounds like a nice convenience. Otherwise, I think this works, especially if you added a note to the README indicating what to change if you're developing with a dev version of the API.

codeviking: It was super easy. Thanks for the nudge :).

Here's the docs that explain the mechanics:
https://github.com/allenai/scholar-reader/blob/444c8b2ef2d68f90c65260dfea348f1362272397/ui/README.md#using-a-local-api

codeviking:
@andrewhead Let me know what you think of the tweak, it was very straightforward and a helpful nudge to do so!

andrewhead:
Looks great! Merge whenever you want.

PR #8, opened 2019-12-12
Add a UI for viewing all papers in the system.

codeviking:
This adds the `/papers/` endpoint to the interface, which displays
a list of papers the system knows about.

Given it's a standalone mechanism that is likely to be discarded,
I used standard Material UI components for quickly building the
interface.

There's no pagination currently. I figured we can add that later
when it's necessary.

As part of this I had to do some acrobatics with the webpack
build. I added another HTML file and entrypoint.

Here's a screenshot:
image


api/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "API for the Semantic Scholar Reader",
   "main": "index.js",
   "scripts": {
-    "start": "ts-node src/index.ts",
+    "start": "ts-node-dev src/index.ts",
```
codeviking: The `ts-node-dev` utility watches files for changes and automatically restarts the service whenever they occur.

A nice little utility for rapid feedback :).

api/src/queries.ts
```diff
@@ -78,6 +85,30 @@ export class Connection {
...
+                 WHERE citation.paper_id = paper.s2_id
+             ) AS citations,
+             (
+                SELECT COUNT(*)
+                  FROM symbol
+                 WHERE symbol.paper_id = paper.s2_id
+             ) AS symbols
+        FROM paper
+    ORDER BY symbols DESC, citations DESC
+    `);
```
codeviking: I couldn't get `Knex` and a subquery to play nice, so I had to just write good ole' SQL.  That said, I like writing good ole' SQL :).

ui/config/webpack.config.js
```diff
@@ -158,14 +167,12 @@ module.exports = function(webpackEnv) {
       // There will be one main bundle, and one file per asynchronous chunk.
       // In development, it does not produce real files.
       filename: isEnvProduction
-        ? 'static/js/[name].js'
-        : isEnvDevelopment && 'static/js/bundle.js',
+        ? 'static/js/[name].[hash:6].js'
+        : 'static/js/[name].js',
```
codeviking: The `hash` is a checksum, so it provides cache breaking.  I'm not sure why webpack doesn't include it by default (I thought it did, but who knows).

The `name` here is necessary as now we have separate bundles for each page.

ui/config/webpack.config.js
```diff
@@ -247,8 +254,7 @@ module.exports = function(webpackEnv) {
       // https://twitter.com/wSokra/status/969633336732905474
       // https://medium.com/webpack/webpack-4-code-splitting-chunk-graph-and-the-splitchunks-optimization-be739a861366
       splitChunks: {
-        chunks: 'all',
-        name: false,
```
codeviking: I had to remove this, so that chunks remove their name. This allows us to use `excludeChunks: [...]` below.

ui/config/webpack.config.js
```diff
@@ -495,6 +501,34 @@ module.exports = function(webpackEnv) {
           {
             inject: true,
             template: paths.appHtml,
+            excludeChunks: [ 'list' ]
```
codeviking: This is the include for the reader / main application, so we exclude the `list` bundle.

ui/config/webpack.config.js
```diff
@@ -495,6 +501,34 @@ module.exports = function(webpackEnv) {
...
+        )
+      ),
+      new HtmlWebpackPlugin(
+        Object.assign(
+          {},
+          {
+            inject: true,
+            template: paths.paperListHtml,
+            filename: 'papers/index.html',
+            excludeChunks: [ 'reader' ]
```
codeviking: ...and this is the list, so we exclude the reader bundle.

ui/config/webpack.config.js
```diff
@@ -577,22 +611,6 @@ module.exports = function(webpackEnv) {
...
-          importWorkboxFrom: 'cdn',
-          navigateFallback: publicUrl + '/index.html',
-          navigateFallbackBlacklist: [
-            // Exclude URLs starting with /_, as they're likely an API call
-            new RegExp('^/_'),
-            // Exclude URLs containing a dot, as they're likely a resource in
-            // public/ and not a SPA route
-            new RegExp('/[^/]+\\.[^/]+$'),
-          ],
-        }),
```
codeviking: We don't use service workers right now, nor should we (they just make debugging harder).  If / when our application is widely used we can introduce this complexity -- but for now we avoid another piece of machinery that might break :).

ui/src/api.ts
```diff
@@ -34,7 +38,7 @@ export async function mathMlForArxivId(arxivId: string) {
 /**
  * 'get' is a Promise returned by 'axios.get()'
  */
-async function doGet(get: Promise<AxiosResponse<any>>) {
+async function doGet<T>(get: Promise<AxiosResponse<T>>) {
```
andrewhead: Nice.

ui/src/papers/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    year: string;
+}
+
+interface S2ApiAuthor {
+    authorId: string;
+    name: string;
+    url: string;
+}
+
+// We only retain a the fields from S2ApiPaper that we need, as otherwise
```
andrewhead:
```suggestion
// We only retain the fields from S2ApiPaper that we need, as otherwise
```

ui/src/papers/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    year: string;
+}
+
+interface CachedPaperWithMeta {
+    expires: number;
+    paper: PaperWithMeta;
+}
+
+/**
+ * This wrapper returns paper metadata from a local cache for up to 24 hours.a
```
andrewhead:
```suggestion
 * This wrapper returns paper metadata from a local cache for up to 24 hours.
```

ui/src/papers/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+                                        {paper.authors.map((author, idx) => (
+                                            <>
+                                                <a key={author.url} href={author.url}>
+                                                    {author.name}
+                                                </a>
+                                                {idx !== paper.authors.length - 1 ? ', ' : null}
+                                            </>
+                                        ))}
+                                        {" "}&bull; {paper.venue}
+                                        {" "}&bull; {paper.year}
```
andrewhead: In some cases, I think venue and year might be null. Can we add a check for those here?

api/src/api.ts
```diff
@@ -215,5 +215,15 @@ export const plugin = {
         }
       }
     });
+
+    server.route({
+      method: "GET",
+      path: "papers/list",
+      handler: async (request, h) => {
+        const papers = await dbConnection.getAllPapers();
+        return h.response(papers).code(200);
```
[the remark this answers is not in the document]
codeviking: Awesome!

ui/README.md
```diff
@@ -20,8 +20,7 @@ npm install
 Then you can launch the app as follows:
 
 ```bash
-PORT=3001 npm start  # 3001 is the port number. Save port 3000
-                     # for the development API server.
+npm start
```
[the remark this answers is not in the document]
codeviking: Ah, I just updated the default to be `3001` in `scripts/start.js` (so the user doesn't have to do anything).

ui/config/webpack.config.js
```diff
@@ -495,6 +501,34 @@ module.exports = function(webpackEnv) {
           {
             inject: true,
             template: paths.appHtml,
+            excludeChunks: [ 'list' ]
+          },
+          isEnvProduction
+            ? {
+                minify: {
+                  removeComments: true,
```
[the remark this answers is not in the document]
codeviking: Good catch. This will just remove comments in inline script and/or HTML. I'll toggle this to false!

ui/src/papers/index.tsx
codeviking: Good catch, will do.

api/package.json
codeviking: Nope! Good catch.

ui/config/webpack.config.js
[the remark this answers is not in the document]
codeviking: Fixed!

ui/src/papers/index.tsx
codeviking: Fixed!

api/package.json
codeviking: Removed!

PR #9, opened 2019-12-12
Annotations

andrewhead:
Added support for annotating symbols and citations within the interactive reader.

Annotation Interface

To enter annotation mode, a user types 'Ctrl+Shift+A'. (They can later disable annotation mode, and hide all manual annotations, by typing 'Ctrl+Shift+A' a second time).

To annotate an entity, a user clicks once to start rectangular selection, and then clicks a second time to finish that selection. Then an entity annotation is created! Annotations can be created on any page.

Each entity has a tag that indicates whether it's a citation of a symbol. To change the tag for an entity, click on that entity, and then choose "Citation" or "Symbol" from the tooltip that appears, or type "c" for citation or "s" for symbol. Citations are colored green and symbols are colored purple.

To delete an annotation, click on that annotation and type "Delete", or click the delete button in the tooltip that appears when the annotation is selected.

Annotations are stored in the database as soon as they're created. They are therefore persistent (you can close and re-open the reader and they will still be there) and shared across users.

Feedback requested:
@codeviking: Did I commit an egregious sins in writing the React code? While most users won't use the annotations, annotation functionality will probably be kept in the code for the foreseeable future.
@kyleclo: You and I are going to use this tool starting possibly tomorrow. Got any feedback for the interaction design to make annotation easier for us?

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -137,7 +137,7 @@ export const plugin = {
       path: "papers/arxiv:{arxivId}/annotations",
       handler: async (request, h) => {
         const arxivId = request.params.arxivId;
-        const id = dbConnection.postAnnotationForArxivId(
+        const id = await dbConnection.postAnnotationForArxivId(
           arxivId,
           request.payload as AnnotationData
         );
@@ -152,6 +152,8 @@ export const plugin = {
     });
 
     /**
+     * This PUT route only lets you update an annotation; it does not let you create a new
+     * annotation. For the time being, the only way to create a new annotation is with POST.
      * Example usage:
      * requests.put(
      *   "http://localhost:3000/api/v0/papers/arxiv:1508.07252/annotation/2",
@@ -172,12 +174,12 @@ export const plugin = {
       path: "papers/arxiv:{arxivId}/annotation/{id}",
       handler: async (request, h) => {
         const { arxivId, id } = request.params;
-        const { created, annotation } = await dbConnection.putAnnotation(
+        const annotation = await dbConnection.putAnnotation(
           arxivId,
           Number(id),
           request.payload as AnnotationData
         );
-        return h.response(annotation).code(created ? 201 : 200);
+        return h.response(annotation).code(200);
       },
       options: {
         validate: {
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -152,8 +152,6 @@ export const plugin = {
     });
 
     /**
-     * This PUT route only lets you update an annotation; it does not let you create a new
-     * annotation. For the time being, the only way to create a new annotation is with POST.
      * Example usage:
      * requests.put(
      *   "http://localhost:3000/api/v0/papers/arxiv:1508.07252/annotation/2",
@@ -226,6 +224,5 @@ export const plugin = {
         return papers;
       }
     });
-
   }
 };
--- a/ui/src/Annotation.tsx
+++ b/ui/src/Annotation.tsx
@@ -1,5 +1,6 @@
 import ClickAwayListener from "@material-ui/core/ClickAwayListener";
 import MuiTooltip from "@material-ui/core/Tooltip";
+import classNames from "classnames";
 import React from "react";
 import * as selectors from "./selectors";
 import { ScholarReaderContext } from "./state";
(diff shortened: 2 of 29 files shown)


api/src/api.ts
```diff
@@ -152,6 +152,8 @@ export const plugin = {
     });
 
     /**
+     * This PUT route only lets you update an annotation; it does not let you create a new
+     * annotation. For the time being, the only way to create a new annotation is with POST.
```
codeviking: I think those are standard REST semantics, so `For the time being` can be omitted, as I think this is correct!

ui/src/Annotation.tsx
```diff
@@ -20,33 +26,71 @@ export class Annotation extends React.Component<
...
+  deselect() {
+    this.setState({ selected: false });
+    if (this.props.onDeselected) {
+      this.props.onDeselected();
+    }
+  }
+
+  isSelected() {
+    return this.props.selected || this.state.selected;
+  }
```
codeviking: Could you just use `state.selected` as the source of truth, and use [React's getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) to set `state.selected` from `props.selected`?

ui/src/SelectionLayer.tsx
```diff
@@ -0,0 +1,72 @@
...
+          }
+        }}
+      >
+        {this.state.anchor !== null && this.state.active !== null && (
+          <Selection anchor={this.state.anchor} active={this.state.active} />
+        )}
+      </div>
+    );
+  }
+}
```
codeviking: Have you tested what happens if the user blurs / defocuses the window?  Things might work as-is, or your mind need to bind to the associated window events to handle them as desired.

ui/src/UserAnnotation.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  render() {
+    let typeClass = "";
+    if (this.props.annotation.type === "citation") {
+      typeClass = "citation-user-annotation";
+    } else if (this.props.annotation.type === "symbol") {
+      typeClass = "symbol-user-annotation";
+    }
```
codeviking: I like using the [classnames](https://www.npmjs.com/package/classnames) package for this.  It'd allow you do do something like:

```TypeScript
const className = classNames("user-annottation", { 
    "citation-user-annotation": this.props.annotation.type === "citation",
    "symbol-user-annotation": this.props.annotation.type === "symbol"
});
```

I don't think you need to change this now though.  Just a potential `TODO` for down the road, particularly if you find yourself doing this kind of thing often.

andrewhead: Nice! I'll integrate this. The existing code was feeling a little clunky.

ui/src/Annotation.tsx
andrewhead: Good idea! Will do.

api/src/api.ts
andrewhead: Cool! I saw some examples online of REST servers that create on PUT when the ID is new, and [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) also kind of hints that sometimes PUT creates a new object. That said, this sounds like optional behavior so I'll update the comment.

ui/src/SelectionLayer.tsx
andrewhead: For the time being, nothing too annoying ^_^

I think the desired behavior is to cancel any selection in-progress when the window is blurred or defocused, and I'll add that in.

A trickier case of undesired behavior is when someone clicks out of a tooltip and it immediately starts a new selection. This feels a bit annoying and I'm trying to think of how to fix it.  One option might be to add a condition to the selection layer that a selection can only start if the selection layer has focus.

In order for that to feel smooth, I would also want to add some indicator that the selection layer has focus, so that on their first click an annotator can see that something has changed, even if their selection doesn't start until they click a second time.

andrewhead:
Notes from @kyleclo:
* Can we add annotator ID
* Can we add record creation and update dates?

ui/src/Annotation.tsx
andrewhead: I updated Annotation to be a controlled component so that it no longer needs to merge together props and state 🎉

andrewhead:
> Notes from @kyleclo:
> 
> * Can we add annotator ID

In theory yes, it should be pretty simple. Hopefully I get a chance to do this before we start annotating papers. I'll save that for a future PR.

> * Can we add record creation and update dates?

Yup! Done.

PR #10, opened 2019-12-13
Pipeline: Upload data to timestamped schemas

andrewhead:
Sometimes, I want to create a "dev" copy of the database to test changes to the data pipeline in the UI, without impacting a clean master set of data in the dataset that everyone else is using for development and demos.

In this change, every time the data pipeline is run, it will upload data to a different schema (like a different namespace in Postgres), named by the time at which the pipeline was run (e.g., `schema_20191213_024444`). In the future, once we have validated that the data in a schema is high quality enough, it can be renamed to `public` to make it live in the UI.

This PR isn't _quite_ complete. I also want to make the schema name that data is uploaded to from the data processing scripts configurable (e.g., I can choose to always upload results to the "dev" schema). Though it is ready for high-level design feedback.

### DIFF
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -10,14 +10,15 @@ nconf
       host: "scholar-reader.c5tvjmptvzlz.us-west-2.rds.amazonaws.com",
       port: 5432,
       database: "scholar-reader",
-      user: "api"
+      user: "api",
+      schema: "public"
     }
   });
 
 export const init = async (config: nconf.Provider) => {
   const server = new Server({
     port: 3000,
-    host: "0.0.0.0", 
+    host: "0.0.0.0",
     debug: {
       request: ["error"]
     }
@@ -31,7 +32,7 @@ export const init = async (config: nconf.Provider) => {
     }
   });
 
-  server.route({ method: 'GET', path: '/health', handler: () => '👍' })
+  server.route({ method: "GET", path: "/health", handler: () => "👍" });
 
   await server.start();
   console.log("Server running on %s", server.info.uri);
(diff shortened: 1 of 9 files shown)


data-processing/models/models.py
```diff
@@ -175,5 +206,19 @@ def create_tables() -> None:
             EntityBoundingBox,
             Annotation,
         ],
+        # Don't create the tables if they're already created.
         safe=True,
     )
+
+    output_database.execute_sql(
```
andrewhead: @codeviking do you have any pointers on best practices for programmatically granting privileges on automatically-created tables? Will this do or are there gotchas to it?

I also tried commands like 

```sql
ALTER DEFAULT PRIVILEGES
    GRANT SELECT ON TABLES 
TO api;
```

So that the commands would only have to be run once, in a management console. However, that doesn't seem to have worked (maybe I just need to dig deeper into the causes of failure).

api/src/index.ts
```diff
@@ -10,14 +10,15 @@ nconf
       host: "scholar-reader.c5tvjmptvzlz.us-west-2.rds.amazonaws.com",
       port: 5432,
       database: "scholar-reader",
-      user: "api"
+      user: "api",
+      schema: "public"
```
codeviking: Nice! Glad this exists as a connection setting.

data-processing/models/models.py
codeviking: Nope, the explicit route I think is the best one.

The gotcha is the user executing these will have to have permission to do so.  Provided they created the schema, this should work (I think).

In this case given the data isn't terribly sensitive, you might be able to just do:

```
GRANT ALL ON SCHEMA %s TO %s
```

It's certainly not following the principle of least privilege, but it's ok given the low data sensitivity.

api/src/index.ts
andrewhead: Me too!!

PR #11, opened 2019-12-14
Add the ability to submit feedback.

codeviking:
This adds the ability to submit feedback. The user can submit feedback
from the following contexts:

1. Globally
2. From the citation tooltip
3. From the drawer, after clicking a citation
4. From the symbol tooltip
5. From the drawer, after clicking a symbol

Clicking the button takes the user to a Google form, and prepopulates
the `context` field with information from the user interface. This
mechanism allows us to decorate the feedback with metadata that helps
us examine the user's state upon preparing feedback.

## Here's the global button:
image

## In the citation tooltip:
image

## In the side drawer, after clicking a paper in the citation tooltip:
image

## In the symbol tooltip:
image

## In the symbol drawer:
image


codeviking:
@andrewhead Conflict resolved, so this is ready to go once you're able to take a quick look!

ui/src/ScholarReader.tsx
```diff
@@ -38,6 +40,7 @@ class ScholarReader extends React.Component<ScholarReaderProps, State> {
      * are called from outside ScholarReader.
      */
     this.state = {
+      paperId: props.paperId,
```
andrewhead: Not that I expect the paper ID to change during an application session, though if it does, would this transfer the paper ID from the props to the state?

ui/src/FeedbackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { makeStyles } from "@material-ui/core/styles"
+import { ScholarReaderContext, PaperId } from "./state";
+import queryString from "querystring";
+
+function mkFeedbackLink(paperId?: PaperId, extraContext?: Object) {
+    console.log(extraContext);
+    const baseUrl = "https://docs.google.com/forms/d/e/1FAIpQLSdnTn4ng-3SsNqwr6M7yF54IhABNAw9_KIjPdWC746fIe546w/viewform";
+    const params = {
+        // This ID is produced by Google Forms. If the form changes we
+        // may need to change this value
```
andrewhead: Does this mean if edits are made to the form the value might change, or if we start using another form?

ui/src/FeedbackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+        "entry.331961046": JSON.stringify(Object.assign(
+            {},
+            extraContext || {},
+            { paperId }
+        ))
+    };
+    return `${baseUrl}?${queryString.stringify(params)}`;
+}
+
+const useStyles = makeStyles({
```
andrewhead: Generally I've been trying to use Less rather than styled components as I think it makes live-reloading easier. Do you think we should consider using styled components?

ui/src/FeedbackButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+                    </button>
+                )}</ScholarReaderContext.Consumer>
+            )
+        }
+        case "default":
+        default: {
+            return (
+                <ScholarReaderContext.Consumer>{({ paperId }) => (
+                    <Button
+                        onClick={() => window.open(mkFeedbackLink(paperId, extraContext), '_blank' )}
```
andrewhead: Does this open an entirely new window, or a new tab?

ui/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "moody",
   "version": "0.1.0",
   "private": true,
-  "proxy": "https://s2-reader.apps.allenai.org",
+  "proxy": "http://localhost:3000",
```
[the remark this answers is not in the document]
codeviking: Oops! Good catch. This was a side-effect of me being lazy and not using the `ENV_VAR` approach!

ui/src/CitationTooltipBody.tsx
```diff
@@ -14,25 +19,39 @@ export class CitationTooltipBody extends React.Component<
   render() {
     return (
       <div className="tooltip-body citation-tooltip-body">
-        <div className="tooltip-body__section tooltip-body__label citation-tooltip-body__header">
-          This citation was matched to {this.props.paperIds.length}
-          {this.props.paperIds.length > 1 ? " papers" : " paper"}:
-        </div>
+        <Box padding="8px 8px 0">
```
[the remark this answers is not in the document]
codeviking: I think so.

* `Box` is a `@material-ui` specific component, a sort of catch-all for setting padding, margin and other layout-esque css that's component specific.  Given we're using `@material-ui` I'd suggest opting into their APIs. That said if we think we're likely to abandon `@material-ui` in the future, then we should likely err away from using it.  
*  I love using `display: grid` over `display: flex` lately, as it feels like it's more precise.  Here I'm using `@material-ui`'s `Grid`, which ironically uses `display: flex`.  Again I think it's just a matter of choosing whether we want to "opt-in" to `@material-ui`'s general layout mechanism or not.

The components provided by `@material-ui/` seem quite powerful, and saved me a fair amount of time in composing this PR.  I think it'd be a fine choice for a component library for the project, and seems theme-able to the extent that we can make sure it doesn't end up looking like a Google product.

Let me know what you think!

ui/src/FeedbackButton.tsx
codeviking: Ah, good question.

The ID is auto-generated by Google, and accordingly the mechanism behind it is opaque to us.  So it could change from any action in theory.

I'll update the docs to clarify.

ui/src/FeedbackButton.tsx
codeviking: Ah, interesting!

I went this route as it was `@material-ui/`s canonical mechanism for styling.  That said I'll go the `less` route, given it's more consistent with the existing code-base and works better for live-reload purposes!

ui/src/FeedbackButton.tsx
codeviking: New tab.  Do you think a new (smaller) window would be preferable?

ui/src/ScholarReader.tsx
codeviking: Ah, good question. 

It depends on whether the component is remounted. I might need to use `getDerivedStateFromProps` to make sure the value is updated.

I'll test this locally and make sure it works properly before merging.

codeviking:
> For the feedback button that appears in the toolbar, could we also add the text "Give Feedback" or some other string, in case users don't recognize the icon?

Good idea! I tried this, but the buttons rendered by Mozilla's viewer by default didn't handle text well.  That said I can definitely decorate this with additional styles that fix the issues.  Your ask confirms the gut intuition that this is worth the time investment, so I'll go ahead and do that today!

ui/src/CitationTooltipBody.tsx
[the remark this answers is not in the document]
codeviking: Continuing my thought here a bit.

It seems we're leaning away from `@material-ui` in some respects.  The conversation around using `less` over `makeStyles`, for instance, shows that we've opted out of the canonical styling mechanism provided by `@material-ui`.   

I'd personally advise against this -- as I think there's some gotchas to going this route:

- Having two ways to style things creates a bit of confusion for implementers (as illustrated here). Should one use `less` or `makeStyles`.  Is the answer always `less`? It's easy to opt into `makeStyles` given it's well documented in `@material-ui`s docs.  Are we sure `less` is providing enough value here that it's worth using here? If it is, how do we clearly document when to use one or the other?

- Using both will mean we'll need to do some acrobatics to keep certain things in sync, particularly if / when we create a theme for `@material-ui`'s components.  Hopefully that theme can be something we source in both the `less` files and the `@material-ui` theme, otherwise we'll have two sources of truth for our "layout parameters" (padding values, colors, font-sizes, etc).  

That said, this is all stuff to think about post this week's deadline.  I'll refactor to remain consistent with where we're at now, and maybe you can carve out some time to assess this in early 2020!

ui/src/FeedbackButton.tsx
codeviking: Updated!

ui/src/CitationTooltipBody.tsx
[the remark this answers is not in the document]
codeviking: Updated to use `less`.  Let's chat about this in the new year! 🎉

ui/src/FeedbackButton.tsx
codeviking: Fixed. I removed `makeStyles` and ported the styles into less.

ui/src/ScholarReader.tsx
codeviking: It appeared to update w/o a change.  I think that's because updating the URL refreshes the page and re-executes the entry point.

That said I still adopted the method as I think it's good practice, and ensures the value *does* update if we ever do something fancy and change the `paperId` without refreshing the page.

ui/src/FeedbackButton.tsx
codeviking: I tried in a new (smaller) window. It feels much more contiguous.  The prior treatment made giving feedback feel like leaving the application.

Good call!

PR #13, opened 2019-12-15
UI: Symbol-only highlighting

zkirby:
Clicking on a symbol now highlights (changes the underline to red) for all other matching symbols. 

Open to discussing different ways of highlighting the symbols, just wanted to get something that could be demoed. Also, thinking about switching over to the annotation id's now that those exist (might take a performance hit from doing that tho).

Additionally, wanted to talk about the performance hit that we take from implementing this since we can't update the underline without also re-rending every single onscreen annotation which can be an issue with symbol heavy pages. 

ezgif com-video-to-gif

### DIFF
--- a/ui/src/Annotation.tsx
+++ b/ui/src/Annotation.tsx
@@ -16,6 +16,7 @@ interface AnnotationProps {
   location: BoundingBox;
   tooltipContent: React.ReactNode;
   onKeyDown?: (e: React.KeyboardEvent<HTMLDivElement>) => void;
+  highlighted?: Boolean | false;
 }
 
 interface AnnotationState {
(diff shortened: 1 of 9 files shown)


zkirby:
> Looks great @zkirby, and you got this done real fast!
> 
> What's the source of the performance hit is? (Have you used the React profiling tool before? https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) Even with a few thousand annotations, I would expect us to be able to get the restyling down to near-inperceivable time. Once we get it to interactive time, I think it will be ready to roll in.
> 
> One change that might make the code a bit more straightforward (and maybe it would improve performance?) is if each SymbolAnnotation determined whether it should be highlighted by just comparing itself to the currently selected symbol. That could save us from needing the `highlightedSymbols` state variable and needing to keep it consistent as the selected symbol changes.
> 
> It may also be worth rebasing on top of origin if it's not too tricky, as I made a couple of small performance improvements that may or may not help here.

Good idea! I'll investigate implementing that, as well as rebasing :)

andrewhead:
Closing as this feature development moved from this PR to #20

PR #14, opened 2019-12-16
Show a small snippet of the abstract in the citation tooltip.

codeviking:
This modifies the citation tooltip so that a small snippet from the
abstract id displayed. The limit I set (110 characters) is roughly
2 lines of text, which is what felt natural to me.

In doing this I ported over Semantic Scholar's text truncation
mechanism, as it tries to avoid truncating mid-word and uses the
`utf-8` ellipsis character.

image

### DIFF
--- a/ui/src/style/tooltip.less
+++ b/ui/src/style/tooltip.less
@@ -36,6 +36,17 @@
     grid-gap: @medium * 0.5;
     align-items: center;
   }
+
+
+  /**
+   * Adding a small amount of horizontal padding to citations so that there's
+   * some separation between the hover state background color and the left and
+   * right edges of the text.
+   */
+  .tooltip-body__citation {
+    padding-left: 4px;
+    padding-right: 4px;
+  }
 }
 
 /**
--- a/ui/src/ui-utils.ts
+++ b/ui/src/ui-utils.ts
@@ -29,6 +29,9 @@ const ELLIPSIS = '…';
  * ellipsis upon truncation by default.  If the text is shorter than the provided limit, the full
  * text is returned.
  *
+ * This method was ported from Semantic Scholar's UI codebase. It's a UI
+ * utility.
+ *
  * @param {string} text The text to truncate.
  * @param {number} limit The maximum number of characters to show.
  * @param {boolean} whether to include an ellipsis after the truncation, defaults to true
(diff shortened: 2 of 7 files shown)


ui/src/style/tooltip.less
```diff
@@ -36,6 +36,17 @@
...
   }
+
+
+  /**
+   * Adding a small amount of horizontal padding to citations so that there's
+   * some separation between the hover state background color and the left and
+   * right edges of the text.
+   */
+  .tooltip-body__citation {
+    padding-left: 4px;
```
andrewhead: Style question: Does it make more sense to do units as 'px', 'em', 'rem'?

ui/src/ui-utils.ts
```diff
@@ -19,3 +19,45 @@ export function isKeypressEscape(event: React.KeyboardEvent) {
...
 }
+
+const PATTERN_NON_WORD_CHAR = /\W/;
+const PATTERN_WORD_CHAR = /\w/;
+const ELLIPSIS = '…';
+
+/**
+ * Truncates the provided text such that no more than limit characters are rendered and adds an
+ * ellipsis upon truncation by default.  If the text is shorter than the provided limit, the full
+ * text is returned.
```
andrewhead: Could we throw in a link back to the S2 code this is based on? Just to make it easier for me to find the implementation and keep it updated while we're waiting for standardized S2 components.

ui/src/style/tooltip.less
codeviking: Ah, good question.

The answer is it depends.

If you want the value to adjust as the user adjusts their base font-size (i.e. because they like fonts to be bigger or smaller in general) then you should use `em` or `rem`.

If we want the value to be the same regardless of the user's default font-size, we should use `px`.

I used pixels here to be consistent with the padding set by `@material-ui`, which is using `px` in this case:

image

ui/src/ui-utils.ts
codeviking: Yes, good call!

ui/src/style/tooltip.less
andrewhead: Works for me, it's fine to me if we keep consistent with @material-ui

codeviking: 👍

ui/src/ui-utils.ts
codeviking: I added a sufficiently opaque indication of where this was derived from!

PR #16, opened 2019-12-16
Close the drawer on escape.

codeviking:
This adds code that closes the drawer if the escape key is pressed.
This felt like a nice convenience, and conveniently required very
little code.

I also revised the annotation event handler to use the same mechanism
for binding and removing events. We don't actively mount / remount
the top level `<ScholarReader />`, so there's no active motivator --
but in case we ever do it'll ensure things get added and removed properly.

### DIFF
--- a/ui/src/ScholarReader.tsx
+++ b/ui/src/ScholarReader.tsx
@@ -203,16 +203,30 @@ class ScholarReader extends React.Component<ScholarReaderProps, State> {
     this.setState({ userAnnotations: annotations });
   }
 
+  closeDrawerOnEscape = (event: KeyboardEvent) => {
+    if (event.key === "Escape") {
+      this.setDrawerState("closed");
+    }
+  }
+
+  toggleUserAnnotationState = (event: KeyboardEvent) => {
+    if (event.ctrlKey && event.shiftKey && event.key !== "a") {
+      this.setUserAnnotationsEnabled(!this.state.userAnnotationsEnabled);
+    }
+  }
+
   async componentDidMount() {
     waitForPDFViewerInitialization().then(application => {
       this.subscribeToPDFViewerStateChanges(application);
     });
     this.loadDataFromApi();
-    document.onkeypress = (e: KeyboardEvent) => {
-      if (e.ctrlKey && e.shiftKey && e.key !== "a") {
-        this.setUserAnnotationsEnabled(!this.state.userAnnotationsEnabled);
-      }
-    };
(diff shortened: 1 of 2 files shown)


ui/src/ScholarReader.tsx
```diff
@@ -203,16 +203,30 @@ class ScholarReader extends React.Component<ScholarReaderProps, State> {
     this.setState({ userAnnotations: annotations });
   }
 
+  closeDrawerOnEscape = (event: KeyboardEvent) => {
+    if (event.key === "Escape") {
```
andrewhead: There's a new helper for this! Check this out: https://github.com/allenai/scholar-reader/blob/ec635d0e07b8c04565d2e65cbfbe68b8ba852705/ui/src/SelectionLayer.tsx#L58

The level of carefulness of checking for the Escape key may, however, be unnecessary.

ui/src/ScholarReader.tsx
```diff
@@ -203,16 +203,30 @@ class ScholarReader extends React.Component<ScholarReaderProps, State> {
...
-        this.setUserAnnotationsEnabled(!this.state.userAnnotationsEnabled);
-      }
-    };
+    window.addEventListener('keypress', this.toggleUserAnnotationState);
+    window.addEventListener('keydown', this.closeDrawerOnEscape);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('keypress', this.toggleUserAnnotationState);
+    window.removeEventListener('keydown', this.closeDrawerOnEscape);
```
andrewhead: Great!!

ui/src/ScholarReader.tsx
codeviking: Awesome!

I'll update.

codeviking: Ah, interesting. The utility expects a `React.KeyboardEvent`, but in this case we're handling a native, `KeyboardEvent`.

I wonder if I can somehow create a `React.KeyboardEvent` from a raw `KeyboardEvent`, or overload the method to accept either.

codeviking: FYI, here's the change:
c4e4d984bc2cab52a1b4eed423a0386503e5aea4

andrewhead: Oh, I overlooked that! Either approach sounds fine to me. I could also be persuaded that the isKeypressEscape utility is unnecessarily complex.

andrewhead: I was late to the party with my last comment XD Looks good!

PR #17, opened 2019-12-17
UI : Re-style citation and symbol underlines

nuwandavek:
I've made the following changes

- Recolour the underline to "#333"
- Restyle the underline to "dotted" from "solid"
- Correct a bug in the ui Readme file
- Add height and width corrections to the bounding boxes for annotations

### DIFF
--- a/ui/src/selectors/annotation.ts
+++ b/ui/src/selectors/annotation.ts
@@ -11,6 +11,8 @@ export function divDimensionStyles(pageView: PDFPageView, box: BoundingBox) {
    * location detection in the data processing scripts?
    */
   const SCALE_CORRECTION = 0.9975;
+  const HEIGHT_CORRECTION = -2;
+  const WIDTH_CORRECTION = -2;
   const viewport = pageView.pdfPage.getViewport({
     scale: pageView.viewport.scale * SCALE_CORRECTION
   });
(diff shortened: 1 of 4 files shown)


ui/src/selectors/annotation.ts
```diff
@@ -23,10 +25,10 @@ export function divDimensionStyles(pageView: PDFPageView, box: BoundingBox) {
    * https://github.com/wojtekmaj/react-pdf/blob/73f505eca1bf1ae243a2b7068fce1e86b98b408a/src/Page/AnnotationLayer.jsx#L104
    */
   return {
-    left: viewportBox[0],
+    left: viewportBox[0] - WIDTH_CORRECTION,
     top: viewportBox[1],
-    width: viewportBox[2] - viewportBox[0],
-    height: viewportBox[1] - viewportBox[3]
+    width: viewportBox[2] - viewportBox[0] + (2 * WIDTH_CORRECTION),
+    height: viewportBox[1] - viewportBox[3] + HEIGHT_CORRECTION
```
andrewhead: Can we do these height and width corrections using margin and padding, instead of dynamically setting width and height here? That way the hit-boxes can be swapped out dynamically. Hit boxes sizing is currently implemented here, with the `@highlight-hitbox-padding` parameter: https://github.com/allenai/scholar-reader/blob/23a1fd4c3748e08c7e95ca694595ff437cd36181/ui/src/style/annotation.less#L11-L14

andrewhead:
Merged in at https://github.com/allenai/scholar-reader/commit/8470e17dbd4943a1808b59caf7bc5d7e946f57cf! Thanks @nuwandavek !

PR #20, opened 2019-12-25
Faster version of Symbol Highlights

zkirby:
This is an optimization using strategies @andrewhead and I talked about to speed up symbol annotation highlighting. It now adds no perceivable costs to the annotation tooltip opening. The highlighting of in the video is just a placeholder so it's easier to see on the gif. Haven't really decided on a highlighting for the actual implementation (right now it's just the s2-influential-citation-orange but opening to suggestions.
ezgif com-video-to-gif (1)

### DIFF
--- a/ui/src/Annotation.tsx
+++ b/ui/src/Annotation.tsx
@@ -21,7 +21,7 @@ interface AnnotationProps {
   source?: string;
   className?: string;
   location: BoundingBox;
-  shouldHighlight?: boolean | false;
+  shouldHighlight?: boolean;
   /**
    * Correction factor to apply to bounding box coordinates before rendering the annotation.
    * You normally should not need to set this and should be able to trust the defaults.
@@ -39,6 +39,9 @@ export class Annotation extends React.PureComponent<
   AnnotationProps,
   AnnotationState
 > {
+  static defaultProps = {
+    shouldHighlight: false,
+  }
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
--- a/ui/src/Annotation.tsx
+++ b/ui/src/Annotation.tsx
@@ -41,11 +41,6 @@ export class Annotation extends React.PureComponent<
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
-  constructor(props: AnnotationProps) {
-    super(props);
-    this.state = { selected: false };
-  }
-
   onKeyDown(e: React.KeyboardEvent<HTMLDivElement>) {
     if (uiUtils.isKeypressEscape(e)) {
       this.deselectIfSelected();
@@ -61,6 +56,20 @@ export class Annotation extends React.PureComponent<
 
   select() {
     this.context.setSelectedAnnotationId(this.props.id);
+  }
+
+  shouldHighlight() {
+    if (!this.context.selectedAnnotationId) { return false; }
+    const [typeSelected, idSelected,] = this.context.selectedAnnotationId.split('-');
+    const [typeSelf, idSelf,] = this.props.id.split('-');
+    if (typeSelf !== typeSelected) { return false; }
+    
+    // We are only focusing on symbols for right now;
+    if (typeSelf === 'symbol') {
+      const highlightedIds = this.context.symbolMatchSet[Number(idSelected)];
+      return this.isSelected() || highlightedIds.has(Number(idSelf));
+    }
+    return false;
   }
 
   deselectIfSelected() {
@@ -110,7 +119,8 @@ export class Annotation extends React.PureComponent<
                       selected: this.isSelected(),
                       "source-tex-pipeline":
                         this.props.source === "tex-pipeline",
-                      "source-other": this.props.source === "other"
+                      "source-other": this.props.source === "other",
+                      "annotation-highlight": this.shouldHighlight(),
                     }
                   )}
                   tabIndex={0}
--- a/ui/src/Annotation.tsx
+++ b/ui/src/Annotation.tsx
@@ -21,6 +21,7 @@ interface AnnotationProps {
   source?: string;
   className?: string;
   location: BoundingBox;
+  shouldHighlight?: boolean | false;
   /**
    * Correction factor to apply to bounding box coordinates before rendering the annotation.
    * You normally should not need to set this and should be able to trust the defaults.
@@ -56,20 +57,6 @@ export class Annotation extends React.PureComponent<
 
   select() {
     this.context.setSelectedAnnotationId(this.props.id);
-  }
-
-  shouldHighlight() {
-    if (!this.context.selectedAnnotationId) { return false; }
-    const [typeSelected, idSelected,] = this.context.selectedAnnotationId.split('-');
-    const [typeSelf, idSelf,] = this.props.id.split('-');
-    if (typeSelf !== typeSelected) { return false; }
-    
-    // We are only focusing on symbols for right now;
-    if (typeSelf === 'symbol') {
-      const highlightedIds = this.context.symbolMatchSet[Number(idSelected)];
-      return this.isSelected() || highlightedIds.has(Number(idSelf));
-    }
-    return false;
   }
 
   deselectIfSelected() {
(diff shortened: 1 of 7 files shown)


ui/src/ScholarReader.tsx
```diff
@@ -313,6 +321,28 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
         if (symbols.length >= 1) {
           const mathMl = await api.mathMlForArxivId(this.props.paperId.id);
           this.setMathMl(mathMl);
+
+          /**
+           * Build a mapping from symbol to all possible parent matches
+           */
+          const symbolMatchSet: SymbolMatchSet = {};
+          symbols.map(s => {
```
andrewhead: should this just be a forEach instead of a map, given that the return value isn't used?

ui/src/ScholarReader.tsx
```diff
@@ -313,6 +321,28 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+            symbolMatchSet[s.id] = new Set(
+              selectors.matchingSymbols(s, symbols, mathMl)
+              .map(m => {
+                // Need to get the top most parent as this is the 
+                // symbol that determines the annotation
```
andrewhead: If you just get the parents, doesn't that only match each symbol to itself? If so, that's probably okay, though we wouldn't need to create a symbol match set (just do a literal comparison of the symbol within the shouldHighlight method).

ui/src/Annotation.tsx
```diff
@@ -56,6 +51,20 @@ export class Annotation extends React.PureComponent<
...
+    const [typeSelf, idSelf,] = this.props.id.split('-');
+    if (typeSelf !== typeSelected) { return false; }
+    
+    // We are only focusing on symbols for right now;
+    if (typeSelf === 'symbol') {
+      const highlightedIds = this.context.symbolMatchSet[Number(idSelected)];
+      return this.isSelected() || highlightedIds.has(Number(idSelf));
+    }
+    return false;
+  }
```
andrewhead: Can this code go in `SymbolAnnotation`? It feels like it might belong as logic for the symbol annotation component, rather than code that all annotations execute.

ui/src/Annotation.tsx
```diff
@@ -100,7 +109,8 @@ export class Annotation extends React.PureComponent<
                     "scholar-reader-annotation",
                     this.props.className,
                     {
-                      selected: this.isSelected()
+                      selected: this.isSelected(),
+                      "annotation-highlight": this.shouldHighlight(),
```
andrewhead: I feel like `annotation-highlight` is a bit vague here, I wonder if we can give it a more semantic class name. Say, something like `matching-symbol-annotation`?

ui/src/style/variables.less
```diff
@@ -33,14 +32,13 @@
 @highlight-border-color: #88f;
 @highlight-hover-opacity: 0.2;
 @highlight-hitbox-padding: .1em;
-
+@annotation-highlight-color: @s2-influential-citation-orange;
```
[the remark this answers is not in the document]
zkirby: I'll use @highlight so as to not limit these variables to annotations

ui/src/ScholarReader.tsx
zkirby: yes good catch!

ui/src/ScholarReader.tsx
zkirby: I am a little confused by what you mean by "only match each symbol to itself"? Since it's matching against selectors.matchingSymbols I believe it will match against every exact match but since the annotations are only at the top most parent, we need that id in order to do the actual highlighting I believe. However, I might be misunderstanding hahaha

ui/src/Annotation.tsx
zkirby: I like matching-symbol-annotation!

ui/src/Annotation.tsx
zkirby: Yes agreed that also makes more sense from an abstraction perspective as all annotations don't need to know how to highlight specific annotation types

ui/src/style/variables.less
```diff
@@ -33,14 +32,13 @@
 @highlight-border-color: #88f;
 @highlight-hover-opacity: 0.2;
 @highlight-hitbox-padding: .1em;
-
+@highlight: @s2-influential-citation-orange;
```
andrewhead:
```suggestion
@highlight-color: @s2-influential-citation-orange;
```

ui/src/ScholarReader.tsx
```diff
@@ -313,6 +321,28 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
           this.setMathMl(mathMl);
+
+          /**
+           * Build a mapping from symbol to all possible parent matches
+           */
+          const symbolMatchSet: SymbolMatchSet = {};
+          symbols.forEach(s => {
+            symbolMatchSet[s.id] = new Set(
+              selectors.matchingSymbols(s, symbols, mathMl)
+              .map(m => {
```
andrewhead:
```suggestion
              .map(s => {
```

Or `symbol` or something that conveys the meaning of the list item

ui/src/ScholarReader.tsx
```diff
@@ -313,6 +321,28 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
         if (symbols.length >= 1) {
           const mathMl = await api.mathMlForArxivId(this.props.paperId.id);
           this.setMathMl(mathMl);
+
+          /**
+           * Build a mapping from symbol to all possible parent matches
+           */
+          const symbolMatchSet: SymbolMatchSet = {};
```
andrewhead: Can this be named something like `symbolMatches` or `symbolMatchSets`? The data structure is a dictionary of sets, though the name makes me wonder if it's a set of matches.

ui/src/ScholarReader.tsx
andrewhead: Sorry, it's my mistake. I think I get what's happening here!

First the call to `matchingSymbols()` gets all other symbols with a matching subtree:

```
Symbol (x_i) -> {All other symbols with matching a subtree like: x, i, x_i }
```

And then the subsequent `.map` gets the highest ancestor of those symbols, which will be the "clickable" symbol entity in the document:

```
{All symbols with (x, i, x_i) -> {All clickable symbols containing (x, i, x_i), including: e_x, x^2, ij}
```

Is that the way it works? If so, can we update the comment above with this clarification?

ui/src/Annotation.tsx
```diff
@@ -14,6 +14,7 @@ interface AnnotationProps {
   id: string;
   className?: string;
   location: BoundingBox;
+  shouldHighlight?: boolean | false;
```
andrewhead: Is this supposed to be `boolean | undefined` if `shouldHighlight` is optional?

ui/src/SymbolAnnotation.tsx
```diff
@@ -11,6 +12,18 @@ interface SymbolAnnotationProps {
 }
 
 export class SymbolAnnotation extends React.PureComponent<SymbolAnnotationProps> {
+  static contextType = ScholarReaderContext;
+  context!: React.ContextType<typeof ScholarReaderContext>;
+
+  shouldHighlight() {
+    if (!this.context.selectedAnnotationId) { return false; }
+    const [typeSelected, idSelected,] = this.context.selectedAnnotationId.split('-');
```
andrewhead:
```suggestion
    const [typeSelected, idSelected] = this.context.selectedAnnotationId.split('-');
```

ui/src/SymbolAnnotation.tsx
```diff
@@ -11,6 +12,18 @@ interface SymbolAnnotationProps {
...
+    if (typeSelected !== 'symbol') { return false; }
+    
+    const matchingSymbolIds = this.context.symbolMatchSet[Number(idSelected)];
+    return matchingSymbolIds.has(this.props.symbol.id);
+  }
```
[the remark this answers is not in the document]
andrewhead: I think the current implementation is fine.

ui/src/Annotation.tsx
andrewhead: I mean, just `shouldHighlight?: boolean`?

zkirby: yes this should be shouldHighlight?: boolean?

ui/src/ScholarReader.tsx
zkirby: Yes I will rename it for clarity :)

ui/src/ScholarReader.tsx
zkirby: Yes that is correct, I will update the comment :)!

PR #21, opened 2020-01-30
Save to library functionality

buraczyn:
The review adds functionality and api calls to S2 to fetch and modify library entries.

When signed in on S2:
* Fetches the user library entries from S2 and saves them in local storage for an hour
* Shows "In Your Library" or "Save To Library" depending on if its in your library already
* Clicking on "Save To Library" changes the button text and sends a request to add the entry to the user's library
* Clicking on "In Your Library" opens a new tab directed to the user's library in S2

When signed out:
* Clicking on the "Save To Library" button does what it does now, pops open an info box saying it isn't implemented yet. This is because we don't support sign ins from the reader at this time

readersave (2)

This is what the reader link looks like on the S2 side for more context:
signin redirect

### DIFF
--- a/ui/src/PaperSummary.tsx
+++ b/ui/src/PaperSummary.tsx
@@ -10,6 +10,7 @@ import InfluentialCitationIcon from "./icon/InfluentialCitationIcon";
 import S2Link from "./S2Link";
 import { ScholarReaderContext } from "./state";
 import { truncateText } from "./ui-utils";
+import {userLibraryUrl} from "./s2-url";
 
 function warnOfUnimplementedActionAndTrack(actionType: string) {
   alert("Sorry, that feature isn't implemented yet. Clicking it tells us " +
@@ -24,8 +25,16 @@ function warnOfUnimplementedActionAndTrack(actionType: string) {
 }
 
 function goToLibrary() {
-  const s2LibraryUrl = `https://www.semanticscholar.org/me/library`;
-  window.open(s2LibraryUrl, '_blank');
+  window.open(userLibraryUrl, '_blank');
+}
+
+function trackLibrarySave() {
+  if (window.heap) {
+    window.heap.track(
+        "Save to Library",
+        { actionType: 'save' }
+    );
+  }
 }
 
 interface PaperSummaryProps {
@@ -34,6 +43,7 @@ interface PaperSummaryProps {
 
 interface PaperSummaryState {
   showFullAbstract: boolean;
+  errorMessage: string
 }
 
 export class PaperSummary extends React.PureComponent<
@@ -48,14 +58,26 @@ export class PaperSummary extends React.PureComponent<
     super(props);
 
     this.state = {
-      showFullAbstract: false
+      showFullAbstract: false,
+      errorMessage: ''
     };
+  }
+
+  renderLibraryButton(label: string, onClick: () => void) {
+    return (
+      <Button
+          startIcon={<SaveIcon />}
+          className="paper-summary__action"
+          onClick={onClick}>
+        {label}
+      </Button>
+    )
   }
 
   render() {
     return (
       <ScholarReaderContext.Consumer>
-        {({ papers, jumpPaperId, setJumpPaperId, userLibrary, addToLibrary, userId }) => {
+        {({ papers, jumpPaperId, setJumpPaperId, userLibrary, addToLibrary }) => {
           const paper = papers[this.props.paperId];
           const hasMetrics = paper.citationVelocity !== 0 || paper.influentialCitationCount !== 0;
           const truncatedAbstract = paper.abstract ? truncateText(paper.abstract, 300) : null;
@@ -151,25 +173,25 @@ export class PaperSummary extends React.PureComponent<
                   >
                     Cite
                   </Button>
-                  {inLibrary ?
-                    <Button
-                      startIcon={<SaveIcon />}
-                      className="paper-summary__action"
-                      onClick={() => goToLibrary()}>
-                      In Your Library
-                    </Button> :
-                    <Button
-                        startIcon={<SaveIcon />}
-                        className="paper-summary__action"
-                        onClick={() => {
-                          if (!userId) {
-                            warnOfUnimplementedActionAndTrack('save');
-                          } else {
-                            addToLibrary(this.props.paperId, paper.title);
+                  {inLibrary ? this.renderLibraryButton('In Your Library', () => goToLibrary())
+                      : this.renderLibraryButton('Save To Library', async() => {
+                        if (!userLibrary) {
+                          warnOfUnimplementedActionAndTrack('save');
+                        } else {
+                          try {
+                            await addToLibrary(this.props.paperId, paper.title);
+                            trackLibrarySave();
+                            this.setState({
+                              errorMessage: ''
+                            })
+                          } catch (error) {
+                            this.setState({
+                              errorMessage: 'Cannot save to library at this time.'
+                            })
                           }
-                        }}>
-                      Save To Library
-                    </Button>}
+                        }
+                      })
+                  }
               </div>
               <FavoriteButton
                 favoritableId={{
@@ -177,6 +199,9 @@ export class PaperSummary extends React.PureComponent<
                   entityId: this.props.paperId
                 }}
               />
+              <div className="paper-summary__library-error">
+                {this.state.errorMessage && this.state.errorMessage}
+              </div>
             </div>
           );
         }}
--- a/ui/src/PaperSummary.tsx
+++ b/ui/src/PaperSummary.tsx
@@ -23,6 +23,11 @@ function warnOfUnimplementedActionAndTrack(actionType: string) {
   }
 }
 
+function goToLibrary() {
+  const s2LibraryUrl = `https://www.semanticscholar.org/me/library`;
+  window.open(s2LibraryUrl, '_blank');
+}
+
 interface PaperSummaryProps {
   paperId: string;
 }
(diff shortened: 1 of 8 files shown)


ui/src/PaperSummary.tsx
```diff
@@ -23,6 +23,11 @@ function warnOfUnimplementedActionAndTrack(actionType: string) {
   }
 }
 
+function goToLibrary() {
+  const s2LibraryUrl = `https://www.semanticscholar.org/me/library`;
```
cecile-nguyen: we should add a utm source here so we know where traffic is coming from to S2's library

ui/src/PaperSummary.tsx
```diff
@@ -140,13 +151,25 @@ export class PaperSummary extends React.PureComponent<
...
                   </Button>
-                  <Button
-                    startIcon={<SaveIcon />}
-                    className="paper-summary__action"
-                    onClick={() => warnOfUnimplementedActionAndTrack("save")}
-                  >
-                    Save
-                  </Button>
+                  {inLibrary ?
+                    <Button
```
cecile-nguyen: we could add some heap props here to track interaction with this button

ui/src/PaperSummary.tsx
```diff
@@ -140,13 +151,25 @@ export class PaperSummary extends React.PureComponent<
...
+                      startIcon={<SaveIcon />}
+                      className="paper-summary__action"
+                      onClick={() => goToLibrary()}>
+                      In Your Library
+                    </Button> :
+                    <Button
```
cecile-nguyen: These 2 buttons are pretty similar besides the onclick and label, consider a function that takes in these 2 args and renders the component

ui/src/ScholarReader.tsx
```diff
@@ -31,6 +31,7 @@ import { isKeypressEscape } from "./ui-utils";
 
 interface ScholarReaderProps {
   paperId?: PaperId;
+  userId?: number
```
cecile-nguyen: userId?: ?number since it can be undefined

ui/src/ScholarReader.tsx
```diff
@@ -52,6 +53,11 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
      * are called from outside ScholarReader.
      */
     this.state = {
+      userId: undefined,
```
cecile-nguyen: maybe set to null so it's explicit that we wanted it to not have a value and not oversight that it wasn't set?

ui/src/ScholarReader.tsx
```diff
@@ -52,6 +53,11 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      userLibrary: null,
+      setUserId: this.setUserId.bind(this),
```
cecile-nguyen: don't know if the codebase as the babel plugin for arrow functions to avoid .bind() but if not it's fine as is!

ui/src/ScholarReader.tsx
```diff
@@ -182,6 +196,20 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
     }
   }
 
+  async addToLibrary(paperId: string, paperTitle: string) {
+    if (this.props.paperId !== undefined) {
```
cecile-nguyen: could just do if (this.props.paperId) and it'll understand if its not null/undefined

ui/src/ScholarReader.tsx
```diff
@@ -182,6 +196,20 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      await api.addLibraryEntry(
```
cecile-nguyen: we should do a try/catch to show an error message if for some reason this call fails

ui/src/ScholarReader.tsx
```diff
@@ -182,6 +196,20 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+          paperId,
+          paperTitle
+      );
+      const userLibrary = this.state.userLibrary;
+      if (userLibrary) {
+        const paperIds = userLibrary.paperIds.concat(paperId);
```
cecile-nguyen: any reason for concat vs push here?

ui/src/api.ts
```diff
@@ -78,6 +78,57 @@ export async function deleteAnnotation(arxivId: string, id: number) {
   return await axios.delete(`/api/v0/papers/arxiv:${arxivId}/annotation/${id}`);
 }
 
+export async function addLibraryEntry(paperId: string, paperTitle: string) {
+  const s2ApiUrl = `https://www.semanticscholar.org/api/1/library/entries`;
```
cecile-nguyen: I wonder if we should have an S2 const urls file or obj of some sort. Might be helpful as we add more calls to have all the urls in a central place

ui/src/state.ts
```diff
@@ -18,6 +26,12 @@ export interface State {
   papers: Readonly<Papers>;
   setPapers(papers: Papers): void;
 
+  /*
+   * USER DATA
+   */
+  userId?: number,
```
cecile-nguyen: userId?: ?number

ui/src/state.ts
```diff
@@ -18,6 +26,12 @@ export interface State {
...
+  userLibrary: UserLibrary | null;
```
cecile-nguyen: I think you could do ?UserLibrary which would accept UserLibrary | null | undefined. Not sure if you're strict w/ undefined though

cecile-nguyen:
That reminds me we should add an origin header facet in datadog to filter api calls from reader domain

ui/src/PaperSummary.tsx
andrewhead: small style comment: could be single quotes rather than backticks as there's no variable substitution.

ui/src/api.ts
```diff
@@ -78,6 +78,57 @@ export async function deleteAnnotation(arxivId: string, id: number) {
...
+        console.debug(`Using cached library entries for:${userId}`);
+        return Promise.resolve(parsedItem.userLibrary);
+      } else {
+        console.debug(`Library entries for ${userId} are cached but have expired.`);
+      }
+    }
+  }
+
+  const s2ApiUrl = `https://www.semanticscholar.org/api/1/user`;
+  const data = await doGet(axios.get<UserInfo>(s2ApiUrl, {withCredentials: true}));
```
andrewhead: If I understand right, the S2 API doesn't take a user ID as a parameter, but instead probably does some magic with cookies or session variables to detect the user. Is that the right interpretation? And in that case, the user ID is just used as a key for the cache?

ui/src/ScholarReader.tsx
```diff
@@ -319,6 +347,12 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
           this.props.paperId.id
         );
         this.setUserAnnotations(annotations);
+
+        const userLibrary = await api.getUserLibraryInfo(this.state.userId);
+        if (userLibrary) {
+          this.setUserId(userLibrary.userId);
```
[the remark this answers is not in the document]
andrewhead: I'll resolve this comment, as I just realized that the user ID is initially unset, and updated only based on this call to getUserLibraryInfo.

ui/src/state.ts
buraczyn: This setup doesn't allow for ? before any types, it complains saying "JsDoc types can only be used inside documentation comments". I followed the pattern of other type declarations that specify | null to go around it

ui/src/api.ts
```diff
@@ -78,6 +78,57 @@ export async function deleteAnnotation(arxivId: string, id: number) {
...
 }
 
+export async function addLibraryEntry(paperId: string, paperTitle: string) {
+  const s2ApiUrl = `https://www.semanticscholar.org/api/1/library/entries`;
+  const tags: string[] = [];
+  const response = await axios.post(
+      s2ApiUrl,
+      {
+        paperId,
+        paperTitle,
```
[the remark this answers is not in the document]
buraczyn: We require the paper title for the request in order to maintain backwards compatibility with an table that used to hold the entries, so its an artifact of that

ui/src/api.ts
buraczyn: Yes, I am using the userId as the cache key because I figured it was the most unique thing I could use.

ui/src/api.ts
```diff
@@ -78,6 +78,57 @@ export async function deleteAnnotation(arxivId: string, id: number) {
...
+        tags
+      },
+      {withCredentials: true}
+  );
+  return response.data;
+}
+
+export async function getUserLibraryInfo(userId?: number) {
```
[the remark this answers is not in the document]
buraczyn: I initially enabled caching so we don't need to make calls to get the library every time, but only on an hour basis. But as I was developing, I actually modify those values to show what the current state is anyway. I was thinking the cache would save calls if the user opens different papers in the reader, so it would use the same cache then, but its really not a big issue. I think i'll remove it then, and then we don't need to have reference to the user id anymore!

ui/src/api.ts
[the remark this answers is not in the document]
buraczyn: There's no real API reference for the internal APIs, but here is the link to the resource code for this url: https://github.com/allenai/scholar/blob/master/online/service/src/main/scala/org/allenai/s2/service/api/user/UserResource.scala#L50

And here it is for adding library entries:
https://github.com/allenai/scholar/blob/master/online/service/src/main/scala/org/allenai/s2/service/api/library/LibraryResource.scala#L78

cecile-nguyen:
Looks good!

PR #38, opened 2020-02-11
Move PDF on Drawer Open

zkirby:
The PDF now moves to the side on small resolutions and will be horizontally scrollable on large resolutions. 

**edit**: It will now move a symbol into view if it would otherwise have been obscured by the side drawer.
ezgif com-video-to-gif

### DIFF
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -16,46 +16,11 @@ export class Drawer extends React.PureComponent {
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
-  getNeedMoveLeft() {
-    const symbol = this.context.selectedSymbol;
-    let l = 0;
-    let t = 0;
-    if (symbol) {
-      const bb = symbol.bounding_boxes[0];
-      const { left, top } = selectors.divDimensionStyles(
-        this.context.pages[bb.page + 1].view,
-        bb,
-      );
-      l = left;
-      t = top;
-    }
-    
-    if (this.context.selectedAnnotationId) {
-      let [,id] = this.context.selectedAnnotationId.split('-');
-      if (id === `${this.context.selectedSymbol?.id}`) {
-        console.log('we gone scroll');
-        return [l, t];
-      } else if (id === `${this.context.selectedCitation?.id}`) {
-        return [10, 20];
-      }
-    }
-
-    // 
-    return [10, 20];
-  }
-
   addChildClass(pdfViewerContainer: HTMLElement) {
     // Creating padding for scroll
     Array.from(pdfViewerContainer.children).forEach(page => {
       page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)  
     })
-
-    // Scroll the Annotation into View
-    // const [l, t] = this.getNeedMoveLeft();
-    // pdfViewerContainer.parentElement?.scroll(
-    //   l, t
-    // );
-    console.log(pdfViewerContainer.parentElement?.clientWidth);
   } 
 
   removeChildClass(pdfViewerContainer: HTMLElement) {
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -15,10 +15,22 @@ export class Drawer extends React.PureComponent {
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
+  addChildClass(pdfViewerContainer: HTMLElement) {
+    Array.from(pdfViewerContainer.children).forEach(page => {
+      page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)
+    })
+  } 
+
+  removeChildClass(pdfViewerContainer: HTMLElement) {
+    Array.from(pdfViewerContainer.children).forEach(page => {
+      page.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS)
+    })
+  }
+
   componentWillUnmount() {
     const { pdfViewer } = this.context;
-    if (pdfViewer != null) {
-      pdfViewer.container.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS);
+    if (pdfViewer !== undefined && pdfViewer !== null) {
+      this.removeChildClass(pdfViewer.viewer);
     }
   }
 
@@ -74,9 +86,9 @@ export class Drawer extends React.PureComponent {
     const drawerState = this.drawerState();
     if (pdfViewer != null) {
       if (drawerState !== "closed") {
-        pdfViewer.container.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS);
+        this.addChildClass(pdfViewer.viewer);
       } else {
-        pdfViewer.container.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS);
+        this.removeChildClass(pdfViewer.viewer);
       }
     }
 
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -20,6 +20,10 @@ export class Drawer extends React.PureComponent {
     Array.from(pdfViewerContainer.children).forEach(page => {
       page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)  
     })
+
+    if (this.drawerState() === "show-symbols") {
+      this.context.scrollSymbolIntoView();
+    }
   } 
 
   removeChildClass(pdfViewerContainer: HTMLElement) {
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -18,6 +18,8 @@ export class Drawer extends React.PureComponent {
   positionPdfForDrawerOpen(pdfViewerContainer: HTMLElement) {
     // Creating padding for scroll
     Array.from(pdfViewerContainer.children).forEach(page => {
+      // XXX(@zkirby, @andrewhead) per our discussion at https://github.com/allenai/scholar-reader/pull/38/files#r388514946 
+      // this is 'safe' as pages are not deleted when scrolled out of view (just their inner content).
       page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)  
     })
 
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -34,8 +34,8 @@ export class Drawer extends React.PureComponent {
 
   componentWillUnmount() {
     const { pdfViewer } = this.context;
-    if (pdfViewer !== undefined && pdfViewer !== null) {
-      this.removeChildClass(pdfViewer.viewer);
+    if (pdfViewer !== null) {
+      this.removeChildClass(pdfViewer.container);
     }
   }
 
@@ -91,9 +91,9 @@ export class Drawer extends React.PureComponent {
     const drawerState = this.drawerState();
     if (pdfViewer != null) {
       if (drawerState !== "closed") {
-        this.addChildClass(pdfViewer.viewer);
+        this.addChildClass(pdfViewer.container);
       } else {
-        this.removeChildClass(pdfViewer.viewer);
+        this.removeChildClass(pdfViewer.container);
       }
     }
 
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -7,6 +7,7 @@ import { Favorites } from "./Favorites";
 import FeedbackButton from "./FeedbackButton";
 import PaperList from "./PaperList";
 import SearchResults from "./SearchResults";
+import * as selectors from "./selectors";
 import { ScholarReaderContext } from "./state";
 
 const PDF_VIEWER_DRAWER_OPEN_CLASS = "drawer-open";
@@ -15,10 +16,46 @@ export class Drawer extends React.PureComponent {
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
+  getNeedMoveLeft() {
+    const symbol = this.context.selectedSymbol;
+    let l = 0;
+    let t = 0;
+    if (symbol) {
+      const bb = symbol.bounding_boxes[0];
+      const { left, top } = selectors.divDimensionStyles(
+        this.context.pages[bb.page + 1].view,
+        bb,
+      );
+      l = left;
+      t = top;
+    }
+    
+    if (this.context.selectedAnnotationId) {
+      let [,id] = this.context.selectedAnnotationId.split('-');
+      if (id === `${this.context.selectedSymbol?.id}`) {
+        console.log('we gone scroll');
+        return [l, t];
+      } else if (id === `${this.context.selectedCitation?.id}`) {
+        return [10, 20];
+      }
+    }
+
+    // 
+    return [10, 20];
+  }
+
   addChildClass(pdfViewerContainer: HTMLElement) {
+    // Creating padding for scroll
     Array.from(pdfViewerContainer.children).forEach(page => {
-      page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)
+      page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)  
     })
+
+    // Scroll the Annotation into View
+    // const [l, t] = this.getNeedMoveLeft();
+    // pdfViewerContainer.parentElement?.scroll(
+    //   l, t
+    // );
+    console.log(pdfViewerContainer.parentElement?.clientWidth);
   } 
 
   removeChildClass(pdfViewerContainer: HTMLElement) {
--- a/ui/src/Drawer.tsx
+++ b/ui/src/Drawer.tsx
@@ -7,7 +7,6 @@ import { Favorites } from "./Favorites";
 import FeedbackButton from "./FeedbackButton";
 import PaperList from "./PaperList";
 import SearchResults from "./SearchResults";
-import * as selectors from "./selectors";
 import { ScholarReaderContext } from "./state";
 
 const PDF_VIEWER_DRAWER_OPEN_CLASS = "drawer-open";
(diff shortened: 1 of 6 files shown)


ui/src/PageOverlay.tsx
```diff
@@ -52,6 +52,7 @@ class PageOverlay extends React.PureComponent<PageProps, {}> {
     ) {
       this.props.view.div.removeChild(this._element);
     }
+    this.props.view.div.classList.remove("scholar-reader-overlay-2");
```
zkirby: whoops ill remove this

ui/src/types/pdfjs-viewer.ts
```diff
@@ -38,6 +38,7 @@ export interface DocumentLoadedEvent {
 export interface PDFViewer {
   scrollPageIntoView: (params: ScrollPageIntoViewParameters) => void;
   container: HTMLDivElement;
+  viewer: HTMLDivElement;
```
zkirby: and this!

ui/src/style/pdf-viewer.less
```diff
@@ -2,11 +2,14 @@
  * Move view of PDF page inward when the drawer is open.
  * TODO(andrewhead): Find a way to animate drawer even when left sidebar is open.
  */
-#viewerContainer {
+.page {
   transition-property: "padding-right";
 
   &.drawer-open {
-    padding-right: @drawer-paper-clipping-width + @margin * 2;
+    padding-right: @drawer-paper-clipping-width + @margin * 4;
```
zkirby: I found * 4 looks slightly better

zkirby: If I have time, I might try to come up with a more verbose/mathy solution to this (instead of just @margin * 4)

zkirby:
Note: Updating the PR to include what we discussed during the group meeting - (i.e. that we should move the symbol into the visible part of the paper if it is not visible)

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
-          // + some padding to make it a little bit away from the drawer. 
-          pdfViewer.container.scrollLeft += Math.max((relativeSymbolLeftPosition - viewableViewportWidth) + 50, 0);
-        }
-      } 
-    }, 0);
+    const { selectedSymbol, pdfViewer, pages } = this.state;
+    if (pdfViewer && selectedSymbol) {
+      const symBounds = selectedSymbol.bounding_boxes[0];
+      const { left } = selectors.divDimensionStyles(
+        pages[symBounds.page + 1].view, symBounds
```
andrewhead: It might be a good idea to check that `pages[symBounds.page + 1]` is defined before using its value here

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      );
+      /*
+      * Each component of the calculation: 
+      * Left = left position on the pdf page of the selected symbol
```
andrewhead: I don't entirely understand this. Is it the x-coordinate of the left side of the symbol?

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      * scrollLeft = how much the pdf has been scrolled left already
+      * 200 = the extra padding from the pdf's side panel
```
andrewhead: Is this a default value set by pdf.js, or is this an amount we enforce?

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      * ----------------
+      * innerWidth = possible visible area of the viewport
+      * 470 = width of the drawer that is now obscuring the view
+      */
+      /*
+      * TODO(@zkirby) need to account for if the side panel is not open.
+      */
+      const relativeSymbolLeftPosition = left - pdfViewer.container.scrollLeft + 200;
```
andrewhead: Can we move these magic numbers into named constants?

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      * Left = left position on the pdf page of the selected symbol
+      * scrollLeft = how much the pdf has been scrolled left already
+      * 200 = the extra padding from the pdf's side panel
+      * ----------------
+      * innerWidth = possible visible area of the viewport
+      * 470 = width of the drawer that is now obscuring the view
+      */
+      /*
+      * TODO(@zkirby) need to account for if the side panel is not open.
+      */
```
andrewhead: @zkirby is this something that can make it into this PR? If not, we can do another one, just open a new issue for this. Wondering if it would be helpful to discuss alternatives for implementation here?

ui/src/ScholarReader.tsx
```diff
@@ -412,28 +413,31 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      const { left } = selectors.divDimensionStyles(
+        pages[symBounds.page + 1].view, symBounds
+      );
+      /*
+      * Each component of the calculation: 
+      * Left = left position on the pdf page of the selected symbol
+      * scrollLeft = how much the pdf has been scrolled left already
+      * 200 = the extra padding from the pdf's side panel
+      * ----------------
+      * innerWidth = possible visible area of the viewport
```
andrewhead: Which viewport? Is it specifically the area between the side panel on the left and the drawer on the right? Or does it include the side panel?

ui/src/ScholarReader.tsx
andrewhead: And can we make the `470` the same variable as wherever else it's set in the code?

ui/src/ScholarReader.tsx
zkirby: It's hardcoded in the pdf.js code

ui/src/ScholarReader.tsx
zkirby: I can try to look a little deeper in the PDF code and see if there is an easy way to tell without massive changes.

ui/src/ScholarReader.tsx
zkirby: it includes the viewport of the entire website so any and everything

ui/src/ScholarReader.tsx
zkirby: I believe divDimensionStyles returns the x-coordinate of the left side of the symbol?

zkirby: I see now it might be more appropriate for this to be the 'right' side of the symbol (left + width)

ui/src/ScholarReader.tsx
andrewhead: Great! Can you note that in the comment?

ui/src/ScholarReader.tsx
andrewhead: @zkirby take a look at [these events](https://github.com/mozilla/pdf.js/blob/52749d1f0dd04d344310c5ac599d30e0e0a4b390/web/app.js#L1601) that get fired in pdf.js, do you think any of them are events we can track to detect the state of the sidebar?

We already listen for some events from pdf.js [here](https://github.com/allenai/scholar-reader/blob/3187c2f51f6363eac89cb07fe659f08b93d94346/ui/src/ScholarReader.tsx#L316), so could do the same for other sidebar-related events.

ui/src/ScholarReader.tsx
andrewhead: roger! can you update the comment to clarify?

ui/src/ScholarReader.tsx
zkirby: Should these be moved to a constants.js file?

andrewhead: So far, I've just been assigning them meaningful names, putting them in a file they feel they most belong in, and moving them to the headers of the file. Would that work?

zkirby: Also the 470 comes from some less variables, is there an easy way to import these into JS? Everything online requires installing npm packages

andrewhead: I'm not sure! If it requires an extra npm package, then maybe we should skip it, and just note in the comments that the two values are the same and should be changed together?

zkirby: I'll point out where it comes from in the comments and if we run into this issue in the future maybe we can consider adding the package?

ui/src/ScholarReader.tsx
zkirby: Looks like that worked - thanks!!

ui/src/types/pdfjs-viewer.ts
zkirby: @andrewhead not sure why this was removed, but I need it to add the padding. Hope it's not an issue to add back.

ui/src/ScholarReader.tsx
andrewhead: Sounds good!

ui/src/ScholarReader.tsx
```diff
@@ -404,12 +409,41 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+        /*
+        * Each component of the calculation: 
+        * left + width = right position on the pdf page of the selected symbol
+        * scrollLeft = how much the pdf has been scrolled left already
+        * PDF_SIDE_PANEL_WIDTH = the extra padding from the pdf's side panel (hard coded in pdf.js code)
+        * ----------------
+        * innerWidth = possible visible area of the viewport for the entire website
+        * 470 = width of the drawer that is now obscuring the view
+        */
+        const relativeSymbolRightPosition = (left + width) - pdfViewer.container.scrollLeft + (pdfSideBarIsOpen ? PDF_SIDE_PANEL_WIDTH : 0);
```
andrewhead: @zkirby the `left` position of the symbol is relative to the left-hand side of the page (not relative to the left-hand size of the PDF viewer). Does that mean you need to add in the page `left` coordinate to this computation of the relativeSymbolRightPosition?

ui/src/ScholarReader.tsx
```diff
@@ -404,12 +409,41 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+        const viewableViewportWidth = window.innerWidth - 470;
+        if (relativeSymbolRightPosition > viewableViewportWidth) {
+          // Add 50px padding to make the symbol close to the drawer but not hidden by it.
+          pdfViewer.container.scrollLeft += Math.max((relativeSymbolRightPosition - viewableViewportWidth) + 50, 0);
```
andrewhead: @zkirby nitpick: can you make named constants for `470` and `50`?

ui/src/ScholarReader.tsx
```diff
@@ -404,12 +409,41 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
           box.left + SCROLL_OFFSET_X,
-          box.top + SCROLL_OFFSET_Y
+          box.top + SCROLL_OFFSET_Y,
+          null,
         ]
       });
     }
   }
 
+  scrollSymbolIntoView() {
```
andrewhead: Can you provide a clarifying function-level comment, or rename this function, to indicate that this function will scroll it horizontally into view if it's occluded by a drawer? The current name suggests that this might also do vertical scrolling, or scrolling it rightward into view.

ui/src/Drawer.tsx
```diff
@@ -15,10 +15,27 @@ export class Drawer extends React.PureComponent {
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
+  addChildClass(pdfViewerContainer: HTMLElement) {
+    // Creating padding for scroll
+    Array.from(pdfViewerContainer.children).forEach(page => {
+      page.classList.add(PDF_VIEWER_DRAWER_OPEN_CLASS)  
+    })
```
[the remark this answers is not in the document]
zkirby: Hmmm can you point me to the code that suggests they are deleted as they are scrolled out of view? From the behavior I've observed the pages themselves appear to remain and the annotationLayer is deleted so I believe this should still work :). 
Screen Shot 2020-03-05 at 1 25 01 PM

ui/src/style/pdf-viewer.less
```diff
@@ -2,11 +2,14 @@
...
+    // A very unfortunate side effect of using padding is that it 
+    // will distort the border image if the user shrinks the page too much.
+    border-image: none;
```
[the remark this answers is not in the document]
zkirby: Hahah I don't quite understand why, but the PDF's themselves don't actually have borders with a drop shadow, but just a border-image *of* a drop shadow... it's quite strange. But adding padding will distort border-images as the CSS box model states that the border is outside of the padding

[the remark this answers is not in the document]
zkirby: It's not really noticeable in my opinion so I just removed it, but we can perhaps look into trying to keep the image? It might be a difficult solution, not sure.

ui/src/ScholarReader.tsx
```diff
@@ -308,6 +310,9 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
     eventBus.on("documentloaded", (eventData: DocumentLoadedEvent) => {
       this.setState({ pdfDocument: eventData.source });
```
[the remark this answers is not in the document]
zkirby: I'm not sure why, but this event also appears to fire on document load (or at least PDF load) hahaha

ui/src/ScholarReader.tsx
zkirby: Ah hahaha, I am realizing now - I need to do more testing - but I think I've found a solution based on your above comment to remove the pdfSideBar event hahaha

ui/src/Drawer.tsx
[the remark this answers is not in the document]
andrewhead: Got it! It's been a while since I've looked closely at page refresh behavior, so I trust your recent observations from your recent investigations 👍  Can you add a comment to this effect just clarifying that page `div`s won't be deleted when pages are re-rendered?

ui/src/style/pdf-viewer.less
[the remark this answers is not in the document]
andrewhead: @zkirby would any of these fixes work?
* setting the `margin` property instead of the `padding` property?
* move the padding to a child element or parent element?
Those have sometimes fixed issues for me where the trying to set spacing between elements has affected those elements' bounding boxes.

ui/src/ScholarReader.tsx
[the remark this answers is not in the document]
andrewhead: The `sidebarchanged` event?

Can you find evidence in the `pdf.js` code that the event will always fire before `documentloaded` is called (and that "before" is important, because it's once the document is loaded that all of our annotation layers will be loaded)? If so, just make a note of it in the code and we'll consider this one resolved.

ui/src/ScholarReader.tsx
andrewhead: Sweet, I look forward to seeing it!

ui/src/style/pdf-viewer.less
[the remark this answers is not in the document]
zkirby: 1. Margin will unfortunately not work (I've tried). 
2. I've tried moving it to a child and doing so will unfortunately do the same thing, but I might be able to try moving it to a parent?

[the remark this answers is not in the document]
zkirby: Although I suspect that it will not work which is why I moved it to the page in the first place hahaha. Although it's been long enough I have forgotten hahaha

zkirby:
@andrewhead I've tried adding the padding to different parts of the pdfviewer and unfortunately only adding it to the pages themselves seems to work.

ui/src/Drawer.tsx
```diff
@@ -15,10 +15,29 @@ export class Drawer extends React.PureComponent {
   static contextType = ScholarReaderContext;
   context!: React.ContextType<typeof ScholarReaderContext>;
 
+  positionPdfForDrawerOpen(pdfViewerContainer: HTMLElement) {
+    // Creating padding for scroll
+    Array.from(pdfViewerContainer.children).forEach(page => {
+      // XXX(@zkirby, @andrewhead) per our discussion at https://github.com/allenai/scholar-reader/pull/38/files#r388514946
```
andrewhead:
```suggestion
      // XXX(zkirby, andrewhead) per our discussion at https://github.com/allenai/scholar-reader/pull/38/files#r388514946 
```

andrewhead:
Got it! Thanks for looking into this 👍

ui/src/ScholarReader.tsx
```diff
@@ -404,12 +405,46 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+   * Will scroll a symbol horizontally into view when the drawer opens
+   * if it is now obscured by the drawer.
+   */
+  scrollSymbolHorizontallyIntoView() {
+    const { selectedSymbol, pdfViewer, pages } = this.state;
+    const DRAWER_WIDTH = 470;
+    const SYMBOL_VIEW_PADDING = 50;
+    if (pdfViewer && selectedSymbol) {
+      const symBounds = selectedSymbol.bounding_boxes[0];
+      const { x } = pdfViewer.container.getBoundingClientRect() as DOMRect;
```
andrewhead:
```suggestion
      const { pageLeft } = pdfViewer.container.getBoundingClientRect() as DOMRect;
```

Nit pick: just a suggestion for a more descriptive variable name.

zkirby: Unfortunately can't commit this as it's an object de-structure. But I'll rename this and recommit :)

zkirby:
@andrewhead I think this should be good to go, so if you could merge in when you get a chance (I don't have access hahah) - that'd be awesome!

Issue #26, opened 2020-02-03
Easier way to close the side bar


andrewhead:
Users want an easier way to close the side bar (i.e. the one that contains lists of paper abstracts).

Currently, the side bar can be closed by scrolling to the top and clicking the left chevron, or by typing Escape. The first one feels annoying, and the second option is unknown to most users. Perhaps we should make the chevron "sticky" so that it can be used at any time to dismiss the side bar.

We may also might want to make it possible to close the sidebar by clicking outside of it.

zkirby:
Thoughts on just clicking outside of the sidebar to close it? This feels the most natural to me personally and I was a little confused when the side bar didn't behave that way.

andrewhead:
Agreed, that sounds natural to me! If we went with that, I think we should also include some button that's always visible in the sidebar in case someone is looking for a "close" button.

zkirby:
Also just to clarify, I can work on this issue as well :)

zkirby:
Diving into this a little deeper, should the draw close when you select another symbol? I am unsure about what that interaction should be

andrewhead:
Good question! I think that sounds like a suitable behavior. Let's implement that and see how it feels.

PR #45, opened 2020-02-13
Easier way to close the side bar

zkirby:
Solving issue #26 

**Note:**
- Closes when clicking on *anything* outside of the drawer.
- Styling on the chevron is open to change, was just looking for something that felt better as a fixed symbol. 
ezgif com-video-to-gif (2)


ui/src/Drawer.tsx
```diff
@@ -55,28 +56,32 @@ export class Drawer extends React.PureComponent {
...
+            }}>
+              <MuiDrawer
+                className="drawer"
+                variant="persistent"
+                anchor="right"
+                open={drawerState !== "closed"}
+              >
+                <div className="drawer__header">
+                  <div className="drawer__close_icon MuiButton-contained">
+                    <IconButton onClick={() => setDrawerState("closed")}>
```
andrewhead: Small style comment: I notice "closed" and 'closed' on Lines 70 and 60 have different quotation marks. Do you have Typescript formatting turned on in your VSCode? If not, can you turn it on so we have a consistent auto-formatter? One helpful feature is VSCode's save-on-format feature which will take care of this on save.

ui/src/ScholarReader.tsx
```diff
@@ -130,7 +130,14 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
   }
 
   setDrawerState(state: DrawerState) {
-    this.setState({ drawerState: state });
+    if (state === 'show-symbols') {
+      // There are a few edge cases where a close and show-symbol event 
+      // happen at the same time, thus we need to ensure the show-symbol is always 
+      // run last.
+      setTimeout(() => this.setState({ drawerState: state }), 0);
```
andrewhead: I feel kind of uneasy about scheduling code to run with timers. Sometimes this is an indication that we should do a small code design elsewhere to make sure we get the intended behavior reliably with synchronous execution.

What are the edge cases? Can we refactor the code so they don't happen?

ui/src/style/drawer.less
```diff
@@ -26,3 +23,9 @@
 .drawer__content {
   padding: 0 @padding;
 }
+
+.drawer__close_icon {
+  position: fixed;
+  border-radius: 100px;
```
andrewhead: Don't icon buttons already have rounded corners?

ui/src/style/variables.less
```diff
@@ -73,6 +73,11 @@
 @drawer-paper-clipping-width: @paper-summary-width + (@paper-padding * 2);
 @drawer-paper-clipping-height: 200px;
 
+/**
+* Z INDEXS
+*/
+@drawer-close-button-index: 10;
```
andrewhead:
```suggestion
@drawer-close-button-z-index: 10;
```

andrewhead: Also, why 10? Can you set using one of the values from the Material UI guidelines? https://material-ui.com/customization/z-index/

zkirby: Yup, setting it to the 'drawer' index (1200)

ui/src/style/drawer.less
zkirby: Yes it appears there is a border-radius: 50% on the root of all MUI buttons. For some reason it is not actually working. I will investigate :/

ui/src/Drawer.tsx
zkirby: I have it turned on but for some reason the format on save feature is not working.... will check it out :/

ui/src/ScholarReader.tsx
zkirby: Yes, I agree! Definitely doesn't sit right having to use setTimeout. The issue is that the click away listener listens to *any* clicks outside of the drawer container. This means that buttons that actually open the drawer end up sending close signals (due to the click away). I think this can be prevented by stoping bubbling within the button but I feel that's not very robust as it relies on any developer to know that if they wish to implement anything that opens the drawer.

andrewhead: Ahhhh I see. Great point! What about adding an `onclick` listener to the document or page overlay that closes the drawer? Or, refactoring the `Drawer` so that instead of an explicit `open` property, it opens based on whether there is a symbol of citation selected? That way, once a symbol is deselected and the selected symbol is set to `null`, the drawer would automatically close?

zkirby: Ah I played a around a little bit with the onclick on the document but unfortunately end up running into the same issue about what should/shouldn't send close signals. The second idea sounds really cool tho - I'll take a crack at that!

zkirby: Hmmm so I've implemented this and realized that it unfortunately won't solve the issue :(. The problem is that the click away listener will listen to *any* clicks outside of the drawer. So in any system that has an external open mechanism to the drawer (regardless of the system) will always have this issue. This is because the sequence of event triggers will always be: opensignal (via some button) ==> close signal (picked up by a bubble to the document on the click away).... Not quite sure how to proceed

andrewhead: I see, that makes sense. One workaround is to have the ClickAwayListener only attempt to dismiss the Drawer if a click is made on something other than an annotation, otherwise it keeps it open. Would that be too hacky?

zkirby: I see, I think this kind of circle's back to my other point that it's not very robust to start hard coding specific events that the drawer should know not to close for as then if someone wants to add another button that open's the drawer in the future, they need to also hard code that. It feels almost more hacky to me personally, but I am interested in what you think :o!

andrewhead: Agreed. At this moment, I feel like whatever we choose will feel a _liiiitle_ hacky (which seems super weird because I feel like this interaction pattern is really common! It should be better supported by our programming tools.)

I'm thinking the alternative that sounds most appealing to me so far is to update the ClickAwayListener to not dismiss when other drawer-launching components are clicked, and to put clear comments both before the listener and at the top of the Drawer warning members of our team that if they want to add other drawer-launching components, they need to update the listener (e.g., by adding a new class to a list of classes that will be checked when determining whether to dismiss the drawer). I could be convinced that one of the other approaches we've discussed is cleaner however, though am coming to terms with the fact that there doesn't seem to be a fantastic solution to this XD

zkirby: Hahah okay, I'll move forward with this approach - but yeah it's a shame there isn't a cleaner solution for what would seem like a common interaction

zkirby: Okay just pushed a new solution that uses e.preventDefault(). Trying to see if a click came from a tooltip ended up being a little messy. Wondering what you think of this solution?

andrewhead: Sorry to belabor this 😬 . What does the option for checking to see if a click came from a tooltip look like? The e.preventDefault() approach could work, though it does fragment click-away the logic into several different classes.

zkirby: It was really a combo of using typescript + trying to dig into the event. Since someone can click on the text within the button within the tooltip, I need to check if any parent in the DOM hierarchy of what was clicked has the class 'tooltip'. Which in and of itself is not a very robust solution imo bc now we are making the assumption that all tooltips will contain buttons that open the drawer.

zkirby: Maybe I can play around with that implementation to see if I can get the code to be a little cleaner? Although to be completely honest I think all of these solutions are bad and I'm starting to feel as though this may not be worth the time of re-implementing all these slightly better but also bad solutions hahahahah

andrewhead: Fair enough. In that spirit, why we don't we go with your initial fix (e.preventDefault) for now and just clean it up. I'm thinking adding better breadcrumbs to describe the fix that will make the whole issue more easily understandable to anyone who has to maintain this in the future.

First, a hack-comment above the e.preventDefault() wherever it appears with a more thorough description of the design choice:

```typescript
/*
 * XXX(zkirby): ....
 */
```

And then another within the ClickAwayListener, emphasizing how the future code will have to be written to maintain the current behavior (i.e. that future annotations will also need a call to `e.preventDefault()`. How's that sound?

andrewhead: @zkirby have you gotten a chance to incorporate this documentation into your PR?

zkirby: Yes! So sorry, committing now hahaha

zkirby: Well I found another really hairy bug while trying to finish the implementation on this :/. For some reason preventing event propagation in any part of the annotation (including it's own clickaway listener) prevents the annotation from closing which shouldn't happen. Not sure how to proceed from here, if we are committed to not using setTimeouts then it appears I'll have to implement your last suggestion with traversing the event chain and searching for a tooltip class :((((((

andrewhead: Wow, what a bug! Can we do the traversing the even chain and searching for a tooltip class? I think it's pretty robust all things considered, and probably pretty easy to read and maintain as long as we put it in a nicely-named helper function.

ui/src/Drawer.tsx
```diff
@@ -9,23 +10,67 @@ import SearchResults from "./SearchResults";
...
+    if (pdfViewer != null) {
       pdfViewer.container.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS);
     }
   }
 
+  /**
+   * XXX(zkirby): This is an unfortunate byproduct of the clickaway listener
+   * listening to *all* clicks outside of the drawer. Any developer wishing to 
+   * add a button that opens the drawer will need to add the class name of that element
+   * to the black listed class names list.
```
andrewhead: This is great! Task 1: Update this comment to describe briefly what this code does (as the intent is not clear from the code alone)---specifically, it only closes the drawer if the click does not land on another annotation, because the drawer should be left open if a user has selected another annotation.

andrewhead: I know I'm really leaning on you to update these code comments, and I'll keep pushing. The rule of thumb is that comments are necessary---and the times I'll always ask for them---is when the rationale behind the code isn't clear from the code itself. So I'll always ask for a comment if either of these are true:

1. The code required some **nuanced implementation**, and it would be a tragedy if someone saw this code and decided to replace it with something simpler (like here)
2. The code **isn't self-describing**: Someone won't be able to tell what the code is doing, or why it's doing it, after 30 seconds of looking at the code

Without these clarifications, people will either not want to touch the code, or try to delete it and replace it with something simpler, falling into pitfalls we have already solved. I'm hoping this code base will be both maintainable and robust for years to come and many collaborators, so let's get this right!

ui/src/Drawer.tsx
```diff
@@ -9,23 +10,67 @@ import SearchResults from "./SearchResults";
...
-
+const BLACK_LISTED_CLASS_NAME = "MuiTooltip-tooltip";
 export class Drawer extends React.PureComponent {
+  static contextType = ScholarReaderContext;
+  context!: React.ContextType<typeof ScholarReaderContext>;
+
   componentWillUnmount() {
     const { pdfViewer } = this.context;
-    if (pdfViewer !== undefined && pdfViewer !== null) {
+    if (pdfViewer != null) {
```
andrewhead: Why was the check for `undefined` removed?

ui/src/Drawer.tsx
```diff
@@ -9,23 +10,67 @@ import SearchResults from "./SearchResults";
...
+    switch(this.drawerState()) {
+      case 'show-symbols': {
+        this.context.setSelectedSymbol(null);
+        break;
+      }
+      case 'show-citations': {
+        this.context.setSelectedCitation(null);
+        break;
+      }
+    }
```
andrewhead: Can we combine this with the `closeDrawerOnEscape` functionality in the ScholarReader?

ui/src/style/variables.less
```diff
@@ -73,6 +73,11 @@
 @drawer-paper-clipping-width: @paper-summary-width + (@paper-padding * 2);
 @drawer-paper-clipping-height: 200px;
 
+/**
+* Z INDEXS
```
andrewhead:
```suggestion
* Z INDEXES
```

ui/src/Drawer.tsx
zkirby: I think so! Just to be clear, you mean move the closeDrawerOnEscape into the drawer correct?

ui/src/Drawer.tsx
zkirby: != null is equivalent to !== undefined && !== null due to JS type coercion for single equals. It was used at many of my previous internships so I thought I would introduce it here hahah.

ui/src/Drawer.tsx
zkirby: Gotcha :)! No problem, I can update this comment to better reflect the functionality. Just to be clear, this code *does* close the drawer if the user clicks on another annotation as that is the implementation we discussed. Should that not be the case? If not there will need to be some changes that are made.

ui/src/Drawer.tsx
andrewhead: Roger! Okay, this is fine.

ui/src/Drawer.tsx
andrewhead: Yup! Either that, or making some helper function in the ScholarReader or the Drawer that does the setting of the selected citation and symbol to null.

ui/src/Drawer.tsx
```diff
@@ -9,23 +10,68 @@ import SearchResults from "./SearchResults";
...
-    if (pdfViewer !== undefined && pdfViewer !== null) {
+    if (pdfViewer != null) {
       pdfViewer.container.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS);
     }
   }
 
+  /**
+   * XXX(zkirby): Since the clickaway listener listens to *all* clicks outside of the 
+   * drawer, if we do not have the code below it will close after a button is clicked that 
+   * is meant to open the drawer. The code below simple gets the element that the click that is intending
```
andrewhead:
```suggestion
   * is meant to open the drawer. The code below simply gets the element that the click that is intending 
```

ui/src/Drawer.tsx
```diff
@@ -9,23 +10,68 @@ import SearchResults from "./SearchResults";
...
       pdfViewer.container.classList.remove(PDF_VIEWER_DRAWER_OPEN_CLASS);
     }
   }
 
+  /**
+   * XXX(zkirby): Since the clickaway listener listens to *all* clicks outside of the 
+   * drawer, if we do not have the code below it will close after a button is clicked that 
+   * is meant to open the drawer. The code below simple gets the element that the click that is intending 
+   * to close the drawer originated from and traverses the class list and class list of all 
+   * parent elements looking for if this click happened from within a tooltip.
```
andrewhead:
```suggestion
   * parent elements looking for if this click happened from within a tooltip. 
   * Only close the drawer if the click is not within the tooltip.
```

PR #49, opened 2020-02-18
Symbol highlight in faint blue background instead of orange underline

jocelynzsun:


ui/src/style/variables.less
```diff
@@ -23,7 +23,7 @@
 @s2-text-default: #2e414f;
 @s2-text-light: #8c9296;
 @s2-border-light: #e0e0e0;
-@s2-influential-citation-orange: #dd913f;
+@s2-influential-citation-orange: #3f91dd48;
```
zkirby: Curious where this hex code comes from? I thought hex codes were 6 digits max? Also I think it might be best to not change s2-influential-citation-orange and instead create a new variable. Thoughts?

andrewhead: Agreed with @zkriby's point, I think semantically it makes sense to create a new variable, and leave the `@s2-influential-citation-orange` be so that it can be continue to be used by other components.

andrewhead:
@jocelynzsun Can you edit the description of this PR to include a brief summary of the change---what it is, how you did it---and a screenshot of the updated UI?

ui/src/style/variables.less
```diff
@@ -33,7 +33,7 @@
 @highlight-border-color: #88f;
 @highlight-hover-opacity: 0.2;
 @highlight-hitbox-padding: .1em;
-@highlight-annotation-color: @s2-influential-citation-orange;
+@highlight-matching-entity-color: rgba(70, 130, 180, 0.3);
```
zkirby: I think it might be best to be consistent and use hex codes for all color variables - you can convert it using something like this website (https://www.rgbtohex.net/)! :)

PR #59, opened 2020-02-28
Selectively disabled link for citations

jocelynzsun:
As discussed during the meeting, the css selector successfully identified the links specifically for paper citations and disabled them. When the mouse hovers a citation, only the purple highlight is displayed, and when a citation is clicked, the citation tooltip appears.

Screenshot:


ui/src/style/pdf-viewer.less
```diff
@@ -14,5 +14,12 @@
  * Hide default annotations to avoid clutter and confusion
  */
 .annotationLayer {
-  display: none;
+  display: inline;
 }
```
andrewhead: Can we just delete this segment?

ui/src/style/pdf-viewer.less
```diff
@@ -14,5 +14,12 @@
...
+
+/**
+  * Hide default annotation specifically for citations to avoid unnecessary jumps
+  */
+a[href^="#cite"] {
```
andrewhead: @jocelynzsun can you make the selector here more specific? One possible format could be:

```css
.class-for-default-pdf.js-annotations a[href^="#cite"]
```

Where `.class-for-default-pdf.js-annotations` could be determined by looking at the annotations PDF.js makes using the web inspector.

Right now, this will match any link that is not to a citation, though we have other links in the document we need to preserve.

ui/src/style/pdf-viewer.less
```diff
@@ -14,5 +14,12 @@
  * Hide default annotations to avoid clutter and confusion
  */
 .annotationLayer {
-  display: none;
+  display: inline;
 }
+
+/**
+  * Hide default annotation specifically for citations to avoid unnecessary jumps
+  */
```
andrewhead: Would a more accurate comment here specify that we're hiding annotations for citations from PDF.js to make ours the default?

andrewhead:
@jocelynzsun Can you also add a GIF showing that the links for equations and figures still work?

jocelynzsun:
GIF for links other than citations:
record1

GIF for symbol annotations (a bit slow):
record2

PR #62, opened 2020-03-07
Consistent highlight between pdf.js annotation and custom annotation

jocelynzsun:
I used the :not() selector because we previously disabled behavior for any links with "#cite", this is just to avoid any unexpected behavior. 

All modified properties were found in the original pdf.js stylesheet, I only changed the properties that have to do with colors. 

Screenshots:

(to be compared with our custom highlights)

### DIFF
--- a/ui/src/style/pdf-viewer.less
+++ b/ui/src/style/pdf-viewer.less
@@ -22,7 +22,7 @@
   * and custom purple highlight for citation annotations.
   */
 .linkAnnotation a:not([href^="#cite"]):hover {
-  background: rgb(0, 0, 255);
-  opacity: 0.2;
-  box-shadow: 0px 2px 10px rgb(0, 0, 255)
+  background: @highlight-hover-background;
+  opacity: @highlight-hover-opacity;
+  box-shadow: 0px 2px 10px @highlight-hover-background;
 }
--- a/ui/src/style/pdf-viewer.less
+++ b/ui/src/style/pdf-viewer.less
@@ -17,12 +17,17 @@
   display: none;
 }
 
-/** 
-  * Consistent highlight between default pdf.js highlight for tables, figures, etc. 
+/**
+  * Consistent default behavior between pdf.js highlight for tables, figures, etc.
   * and custom purple highlight for citation annotations.
   */
-.annotationLayer .linkAnnotation > a:hover {
-  background: @highlight-hover-background;
-  opacity: @highlight-hover-opacity;
-  box-shadow: 0px 2px 10px @highlight-hover-background;
+.annotationLayer .linkAnnotation > a {
+  border-bottom: @underline-width @underline-color dotted;
+
+  &:hover, 
+  &.selected {
+    background: @highlight-hover-background;
+    opacity: @highlight-hover-opacity;
+    box-shadow: 0px 2px 10px @highlight-hover-background;
+  }
 }
--- a/ui/src/style/pdf-viewer.less
+++ b/ui/src/style/pdf-viewer.less
@@ -16,3 +16,13 @@
 .linkAnnotation a[href^="#cite"] {
   display: none;
 }
+
+/** 
+  * Consistent highlight between default pdf.js highlight for tables, figures, etc. 
+  * and custom purple highlight for citation annotations.
+  */
+.linkAnnotation a:not([href^="#cite"]):hover {
+  background: rgb(0, 0, 255);
+  opacity: 0.2;
+  box-shadow: 0px 2px 10px rgb(0, 0, 255)
+}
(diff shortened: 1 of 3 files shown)


ui/src/style/pdf-viewer.less
```diff
@@ -15,4 +15,14 @@
...
 .linkAnnotation a[href^="#cite"] {
   display: none;
+}
+
+/** 
+  * Consistent highlight between default pdf.js highlight for tables, figures, etc. 
+  * and custom purple highlight for citation annotations.
+  */
+.linkAnnotation a:not([href^="#cite"]):hover {
+  background: rgb(0, 0, 255);
```
andrewhead: @jocelynzsun Can this use variables for background, opacity, and box-shadow, so that colors and opacity of annotations only need to be updated in one place when we change them? The steps would be:

1. Define new variables in `variables.less`
2. Refer to those variables here, and in the styles in `annotation.less`

ui/src/style/pdf-viewer.less
```diff
@@ -15,4 +15,14 @@
   */
 .linkAnnotation a[href^="#cite"] {
   display: none;
+}
+
+/** 
+  * Consistent highlight between default pdf.js highlight for tables, figures, etc. 
+  * and custom purple highlight for citation annotations.
+  */
+.linkAnnotation a:not([href^="#cite"]):hover {
```
andrewhead:
```suggestion
.linkAnnotation a:hover {
```

What do you think of making this selector apply to all PDF.js links? The advantage of this approach is that, if in the future we decide to disable citation annotations, then our styling will apply to PDF.js's links without us needing to make additional modifications.

jocelynzsun: Apparently if we only use a:hover as our selector, this is less specific than the original pdf.js styling and it gets overwritten by the default yellow highlight. If we want to keep the purple highlight even if we disable our citation annotations, it would need to have the selector look something like this:


just wondering if this is something we want to have? will we ever decide to disable citation annotations in the future?

ui/src/style/pdf-viewer.less
jocelynzsun: I don't think any of these variables are used in `annotation.less` at this point?

ui/src/style/pdf-viewer.less
andrewhead: > just wondering if this is something we want to have?

sure! I think the verbose selector you show above is okay.

> will we ever decide to disable citation annotations in the future?

I might want to turn it on and off for upcoming studies, though there are no concrete plans right now.

ui/src/style/pdf-viewer.less
```diff
@@ -21,7 +21,8 @@
   * Consistent highlight between default pdf.js highlight for tables, figures, etc. 
   * and custom purple highlight for citation annotations.
   */
-.linkAnnotation a:not([href^="#cite"]):hover {
+.annotationLayer .linkAnnotation > a:hover,
+.annotationLayer .buttonWidgetAnnotation.pushBotton > a:hover {
```
andrewhead: @jocelynzsun When does this second selector (`.annotationLayer .buttonWidgetAnnotation.pushBotton > a:hover`) become necessary? Can we just use `.annotationLayer .linkAnnotation > a:hover`?

jocelynzsun: We are probably not using the feature it applies to seeing that everything still behaves the same way when I delete it. it seems like `.annotationLayer .linkAnnotation > a:hover` is sufficient

ui/src/style/pdf-viewer.less
andrewhead: Can we use the same variables and CSS properties to style both these link annotations and our annotations? I'm looking at these styles from `annotation.less`. It would be great if they could both use the same properties (e.g., either `background` or `background-color`) and had the same hover and click behavior too!

https://github.com/allenai/scholar-reader/blob/5019fa9f446bb2062d9f37f61630c1ecb78d09d0/ui/src/style/annotation.less#L16-L41

jocelynzsun:
> This is looking good, Jocelyn! Have you gotten a chance to loop in the interactive behavior for the pdf.js annotations (i.e. make it underline by default, light blue highlight on hover, and darker blue highlight on click)?

do you mean the behaviors specifically for citation annotations just in case we disable our annotations?

andrewhead:
Nope, sorry for the ambiguity. I mean for all of the pdf.js annotations
(including for figure, equations, etc.). Ideally all pdf.js annotations (I
think? Though perhaps this is something we should discuss!) should be
styled to have the same underlines as our annotations, have a light blue
highlight on hover, and a darker blue highlight on click, to look like our
annotations.

jocelynzsun:
> darker blue highlight on click

You did the same shade of purple with same opacity value for hover and onclick in `annotation.less`. I guess I'll follow that?

andrewhead:
Looks great! I'll merge this :-)

PR #63, opened 2020-03-09
Merge Consecutive <mi> Tags

zkirby:
Merge consecutive <mi> tags that are also direct descendants of <mrow> tags as this is a good indicator that the consecutive <mi> tags should really have been considered one word. 

**Before:**
Screen Shot 2020-03-09 at 7 47 04 PM

**After:**
Screen Shot 2020-03-09 at 7 13 20 PM

### DIFF
--- a/data-processing/common/parse_equation.py
+++ b/data-processing/common/parse_equation.py
@@ -51,9 +51,11 @@ def get_characters(mathml: str) -> List[Character]:
 def get_symbols(mathml: str) -> List[Symbol]:
     soup = BeautifulSoup(mathml, "lxml")
     node_symbols = []
-    consecutive_chars: List[Symbol] = []
-    all_indices: List[int] = []
-    for symbol_node in soup.find_all_next(SYMBOL_TAGS):
+    consecutive_char_sequence: List[BeautifulSoup] = [] # The symbol_nodes of consecutive characters
+    consecutive_char_indices: List[int] = [] # Indices of the characters of a consecutive character sequence
+    # It's 'safe' to call soup.find_all as the underlying bs4 implementation 
+    # appears to preserve order https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py
+    for symbol_node in soup.find_all(SYMBOL_TAGS):
 
         # Collect indexes of all characters that belong to this symbol
         characters = []
@@ -69,18 +71,20 @@ def get_symbols(mathml: str) -> List[Symbol]:
         # Group consecutive character symbols only if they are a child of the mrow tag.
         # The .has_attr('s2:end') is necessary as not every symbol is annotated by katex
         if symbol_node.has_attr('s2:end') and symbol_node.parent.name in MERGABLE_PARENT_TAGS and symbol_node.name in CHARACTER_TAGS:
-            if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
-                node_symbols.append(_merge_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
-                consecutive_chars = []
-                all_indices = []
-            consecutive_chars.append(symbol_node)
-            all_indices.extend(characters)
-        elif len(consecutive_chars) > 0:
-            node_symbols.append(_merge_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
-            consecutive_chars = []
-            all_indices = []
-
-        if len(consecutive_chars) == 0:
+            # This symbol_node is part of consecutive_char_sequence but not the current sequence.
+            if len(consecutive_char_sequence) > 0 and not consecutive_char_sequence[-1]['s2:end'] == symbol_node['s2:start']:
+                new_symbol = _merge_consecutive_symbols(consecutive_char_sequence, consecutive_char_indices, soup.new_tag(symbol_node.name))
+                node_symbols.append(new_symbol)
+                consecutive_char_sequence = []
+                consecutive_char_indices = []
+            consecutive_char_sequence.append(symbol_node)
+            consecutive_char_indices.extend(characters)
+        else:
+            if len(consecutive_char_sequence) > 0:
+                new_symbol = _merge_consecutive_symbols(consecutive_char_sequence, consecutive_char_indices, soup.new_tag(symbol_node.name))
+                node_symbols.append(new_symbol)
+                consecutive_char_sequence = []
+                consecutive_char_indices = []
             # Clean the MathML of our annotations. Do this cleaning on a clone of the node, so that the
             # annotations are available for processing later nodes.
             node_clone = _clean_node_of_annotations(symbol_node)
@@ -91,29 +95,29 @@ def get_symbols(mathml: str) -> List[Symbol]:
             )
 
     # Possibility that the last symbol was part of a continuing symbol
-    if len(consecutive_chars) > 0:
-        node_symbols.append(_merge_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
+    if len(consecutive_char_sequence) > 0:
+        node_symbols.append(_merge_consecutive_symbols(consecutive_char_sequence, consecutive_char_indices, soup.new_tag(consecutive_char_sequence[0].name)))
 
     # Set the children of each symbols to descendant symbols beneath it in the tree.
     symbols = _set_children(node_symbols)
     return symbols
 
 # Removes s2 annotations from node and all descendants
-def _clean_node_of_annotations(symbol_node: Symbol) -> Symbol:
+def _clean_node_of_annotations(symbol_node: BeautifulSoup) -> BeautifulSoup:
     node_clone = copy.copy(symbol_node)
     _remove_s2_annotations(node_clone)
     for descendant in node_clone.descendants:
         _remove_s2_annotations(descendant)
     return node_clone
 
-# Combines all symbols in consecutive_chars into one NodeSymbol
-def _merge_consecutive_symbols(consecutive_chars: List[Symbol], all_indices: List[int], base_tag: Symbol) -> NodeSymbol:
-    base_tag['s2:start'] = consecutive_chars[0]['s2:start']
-    base_tag['s2:end'] = consecutive_chars[-1]['s2:end']
-    base_tag['s2:index'] = consecutive_chars[0]['s2:index']
-    base_tag.string  = ''.join(list(map(lambda node: node.string, consecutive_chars)))
+# Combines all symbols in consecutive_char_sequence into one NodeSymbol
+def _merge_consecutive_symbols(consecutive_char_sequence: List[BeautifulSoup], consecutive_char_indices: List[int], base_tag: BeautifulSoup) -> NodeSymbol:
+    base_tag['s2:start'] = consecutive_char_sequence[0]['s2:start']
+    base_tag['s2:end'] = consecutive_char_sequence[-1]['s2:end']
+    base_tag['s2:index'] = consecutive_char_sequence[0]['s2:index']
+    base_tag.string = ''.join(list(map(lambda node: node.string, consecutive_char_sequence)))
     node_clone = _clean_node_of_annotations(base_tag)
-    return NodeSymbol(characters=all_indices, mathml=str(node_clone), node=base_tag)
+    return NodeSymbol(characters=consecutive_char_indices, mathml=str(node_clone), node=base_tag)
 
 def _set_children(node_symbols: List[NodeSymbol]) -> List[Symbol]:
 
--- a/data-processing/common/parse_equation.py
+++ b/data-processing/common/parse_equation.py
@@ -67,7 +67,8 @@ def get_symbols(mathml: str) -> List[Symbol]:
                 characters.append(index)
 
         # Group consecutive character symbols only if they are a child of the mrow tag.
-        if symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katext
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
             if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
                 node_symbols.append(_build_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
                 consecutive_chars = []
--- a/data-processing/common/parse_equation.py
+++ b/data-processing/common/parse_equation.py
@@ -80,11 +80,10 @@ def get_symbols(mathml: str) -> List[Symbol]:
             consecutive_chars = []
             all_indices = []
 
-        # Clean the MathML of our annotations. Do this cleaning on a clone of the node, so that the
-        # annotations are available for processing later nodes.
-        node_clone = _clean_node_of_annotations(symbol_node)
-
         if len(consecutive_chars) == 0:
+            # Clean the MathML of our annotations. Do this cleaning on a clone of the node, so that the
+            # annotations are available for processing later nodes.
+            node_clone = _clean_node_of_annotations(symbol_node)
             node_symbols.append(
                 # set node to 'symbol_node', not clone, as the node must have a parent for the
                 # 'set_children' method. The clone will have had its parent removed.
--- a/data-processing/common/parse_equation.py
+++ b/data-processing/common/parse_equation.py
@@ -21,7 +21,7 @@ invisible and we wouldn't want to detect it anyway.
 KATEX_ERROR_COLOR = "#ffffff"
 CHARACTER_TAGS = ["mi"]
 SYMBOL_TAGS = ["msubsup", "msub", "msup", "mi"]
-VALID_PARENT_TAG = ["mrow"]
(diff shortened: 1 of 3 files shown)


data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,55 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
 
+        # Group consecutive character symbols only if they are a child of the mrow tag.
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katext
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
+            if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
+                node_symbols.append(_build_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
+                consecutive_chars = []
+                all_indices = []
+            consecutive_chars.append(symbol_node)
+            all_indices.extend(characters)
```
andrewhead: Does this handle the pathological case of, say,

```xml
<mrow>
s
u
<msup>
p
e
</msup>
r
h
u
m
a
n
</mrow>
```

i.e., would it succeed at grouping together `su` and `rhuman`?

zkirby: Assuming each is wrapped in an individual <mi> tag, then yes hahaha

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,55 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+# Removes s2 annotations from node and all descendants
+def _clean_node_of_annotations(symbol_node: Symbol) -> Symbol:
+    node_clone = copy.copy(symbol_node)
+    _remove_s2_annotations(node_clone)
+    for descendant in node_clone.descendants:
+         _remove_s2_annotations(descendant)
+    return node_clone
+
+# Combines all symbols in consecutive_chars into one NodeSymbol
+def _build_consecutive_symbols(consecutive_chars: List[Symbol], all_indices, base_tag: Symbol) -> NodeSymbol:
```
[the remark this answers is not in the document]
zkirby: I see, is Symbol not the BeatifulSoup symbol? haha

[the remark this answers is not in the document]
zkirby: Oh cool, had no idea!

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,55 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+    base_tag['s2:start'] = consecutive_chars[0]['s2:start']
+    base_tag['s2:end'] = consecutive_chars[-1]['s2:end']
+    base_tag['s2:index'] = base_tag['s2:start']
```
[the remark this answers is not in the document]
zkirby: Yes I wanted to talk about this bc I'm not quite sure what index represents. What should it be in the case of something like ReLU? Is it just the first symbol's index?

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,55 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+    base_tag.string  = ''.join(list(map(lambda node: node.string, consecutive_chars)))
+    node_clone = _clean_node_of_annotations(base_tag)
```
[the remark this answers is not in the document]
zkirby: No, but line 85 should be moved inside the if statement!

data-processing/tests/test_parse_equation.py
```diff
@@ -85,3 +85,15 @@ def test_get_symbol_children():
     assert len(t_sub_i.children) == 2
     assert t in t_sub_i.children
     assert i in t_sub_i.children
+
+def test_parse_consecutive_mi():
+  with open(get_test_path(os.path.join("mathml", "relu.xml"))) as mathml_file:
```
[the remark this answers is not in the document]
zkirby: Ahh I did not, I'll do that now!

data-processing/common/parse_equation.py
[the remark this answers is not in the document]
andrewhead: The first symbol's index is fine 👍  The index is just a unique identifier for each MathML node. No two nodes will have the same index. These indexes are assigned counting up as KaTeX creates each MathML node; though they aren't contiguous---i.e. sometimes I see sequences of indexes that are `0, 1, 3, 6, 7`.

The most important characteristic to preserve for indexes is that they are unique and never repeated between symbols---later stages of the pipeline use this index as an identifier for the symbol.

data-processing/common/parse_equation.py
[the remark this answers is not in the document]
andrewhead: Correct! Lol, I didn't know that BeautifulSoup had a Symbol 😆

data-processing/tests/mathml/relu.xml
```diff
@@ -0,0 +1,13 @@
+<span class="katex">
+  <math xmlns="http://www.w3.org/1998/Math/MathML">
+    <semantics>
```
andrewhead: This example looks Fantastic! Clean, complete, awesome.

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
 
-        # Clean the MathML of our annotations. Do this cleaning on a clone of the node, so that the
-        # annotations are available for processing later nodes.
-        node_clone = copy.copy(symbol_node)
-        _remove_s2_annotations(node_clone)
-        for descendant in node_clone.descendants:
-            _remove_s2_annotations(descendant)
+        # Group consecutive character symbols only if they are a child of the mrow tag.
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katex
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
```
andrewhead: nit pick: wrap line at 100 characters. Are you using black and pylint?

andrewhead: Ah, I never shared with you the settings.json for the project. Try loading these options into your settings.json in VSCode. It should catch the small formatting and typing things.

```json
{
  "python.linting.enabled": true,
  "python.linting.pylintEnabled": true,
  "python.testing.pytestArgs": [
    "tests"
  ],
  "python.linting.pylintArgs": [
    "--extension-pkg-whitelist=cv2",
    "--disable=missing-docstring,len-as-condition,fixme,invalid-name,anomalous-backslash-in-string,line-too-long,bad-continuation,no-self-use,too-many-locals,too-few-public-methods,ungrouped-imports,pointless-string-statement,too-many-arguments"
  ],
  "python.testing.unittestEnabled": false,
  "python.testing.nosetestsEnabled": false,
  "python.testing.pytestEnabled": true,
  "python.linting.mypyEnabled": true,
  "python.linting.mypyArgs": [
    "--ignore-missing-imports",
    "--follow-imports=silent",
    "--show-column-numbers",
    "--strict"
  ]
}
```

andrewhead: I just opened issue #64 for myself to add the settings.json file in the next couple of days. In the meantime, go ahead and incorporate this into your project directory!

andrewhead: nit pick: Is there a more descriptive variable name than `VALID_PARENT_TAGS`, something that conveys that it's tags for parents of clusters of characters or spans of mergeable characters?

data-processing/common/parse_equation.py
```diff
@@ -51,6 +51,8 @@ def get_characters(mathml: str) -> List[Character]:
 def get_symbols(mathml: str) -> List[Symbol]:
     soup = BeautifulSoup(mathml, "lxml")
     node_symbols = []
+    consecutive_chars = []
+    all_indices = []
     for symbol_node in soup.find_all(SYMBOL_TAGS):
```
andrewhead: @zkirby does BeautifulSoup guarantee that nodes will be visited in order? If so, can you write a comment confirming that here? If not, this loop may need to be rewritten so that we can make sure that nodes are visited in the left-to-right order that the code below assumes they are for merging mi's.

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
-        # annotations are available for processing later nodes.
-        node_clone = copy.copy(symbol_node)
-        _remove_s2_annotations(node_clone)
-        for descendant in node_clone.descendants:
-            _remove_s2_annotations(descendant)
+        # Group consecutive character symbols only if they are a child of the mrow tag.
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katex
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
+            if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
+                node_symbols.append(_build_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
```
andrewhead: Maybe `_merge_consecutive_symbols`?

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
-        # Clean the MathML of our annotations. Do this cleaning on a clone of the node, so that the
-        # annotations are available for processing later nodes.
-        node_clone = copy.copy(symbol_node)
-        _remove_s2_annotations(node_clone)
-        for descendant in node_clone.descendants:
-            _remove_s2_annotations(descendant)
+        # Group consecutive character symbols only if they are a child of the mrow tag.
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katex
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in VALID_PARENT_TAG and symbol_node.name in CHARACTER_TAGS:
+            if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
```
andrewhead: Do we need this constraint that checks that `s2:end` doesn't align with the previous character's `s2:start`? Can we just look for `mi` nodes that follow one after the other?

As an example, I wonder if this condition would prevent characters from getting grouped for the following LaTeX:

```latex
$R e L U$        % spaces in between characters
${R}{e}{L}{U}$   % would appear almost if not exactly the same as $ReLU$
```

where R, e, L, and U might (though I don't know for sure) have non-overlapping start and end indices. I don't know enough about how KaTeX parses these to know whether the MathML would have `start` and `end` numbers compatible with your algorithm. It's worth a test!

data-processing/common/parse_equation.py
zkirby: I believe it does, but to be safe I'll change it to `find_all_next` which does.

data-processing/common/parse_equation.py
zkirby: Gotcha, let me test it out. I originally ran into a lot of trouble trying to use .sibling_next and what not but maybe I can take another crack at it.

zkirby: Looks like KaTeX spits out the same thing for all three so we should be good.

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+# Removes s2 annotations from node and all descendants
+def _clean_node_of_annotations(symbol_node: Symbol) -> Symbol:
+    node_clone = copy.copy(symbol_node)
+    _remove_s2_annotations(node_clone)
+    for descendant in node_clone.descendants:
+        _remove_s2_annotations(descendant)
+    return node_clone
+
+# Combines all symbols in consecutive_chars into one NodeSymbol
+def _merge_consecutive_symbols(consecutive_chars: List[Symbol], all_indices: List[int], base_tag: Symbol) -> NodeSymbol:
```
andrewhead: nit pick: it looks like `base_tag` always has the same input (`soup.new_tag('mi')`). Does the method need this parameter?

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+        # Group consecutive character symbols only if they are a child of the mrow tag.
+        # The .has_attr('s2:end') is necessary as not every symbol is annotated by katex
+        if symbol_node.has_attr('s2:end') and symbol_node.parent.name in MERGABLE_PARENT_TAGS and symbol_node.name in CHARACTER_TAGS:
+            if len(consecutive_chars) > 0 and not consecutive_chars[-1]['s2:end'] == symbol_node['s2:start']:
+                node_symbols.append(_merge_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
+                consecutive_chars = []
+                all_indices = []
+            consecutive_chars.append(symbol_node)
+            all_indices.extend(characters)
+        elif len(consecutive_chars) > 0:
```
andrewhead: I'm a little confused about when this condition will get triggered. Is it specifically if symbol_node doesn't have an `s2:end` attribute?

data-processing/common/parse_equation.py
```diff
@@ -51,7 +51,9 @@ def get_characters(mathml: str) -> List[Character]:
 def get_symbols(mathml: str) -> List[Symbol]:
     soup = BeautifulSoup(mathml, "lxml")
     node_symbols = []
-    for symbol_node in soup.find_all(SYMBOL_TAGS):
+    consecutive_chars: List[Symbol] = []
+    all_indices: List[int] = []
```
andrewhead: nit pick: I was a little confused by what all_indices and what consecutive_chars are; can we rename these to something where it's clear that consecutive_chars are symbols (aren't they? I think even things that aren't `<mi>` get added to consecutive_chars?) And can we indicate that `all_indices` are the characters for consecutive characters?

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+                consecutive_chars = []
+                all_indices = []
+            consecutive_chars.append(symbol_node)
+            all_indices.extend(characters)
+        elif len(consecutive_chars) > 0:
+            node_symbols.append(_merge_consecutive_symbols(consecutive_chars, all_indices, soup.new_tag('mi')))
+            consecutive_chars = []
+            all_indices = []
+
+        if len(consecutive_chars) == 0:
```
andrewhead: I'm also confused about when this condition gets triggered. Can you give the condition some type of descriptive variable name, or provide some comment?

data-processing/common/parse_equation.py
```diff
@@ -64,23 +66,54 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
+                NodeSymbol(characters=characters, mathml=str(node_clone), node=symbol_node)
+            )
 
-        node_symbols.append(
-            # set node to 'symbol_node', not clone, as the node must have a parent for the
-            # 'set_children' method. The clone will have had its parent removed.
-            NodeSymbol(characters=characters, mathml=str(node_clone), node=symbol_node)
-        )
+    # Possibility that the last symbol was part of a continuing symbol
+    if len(consecutive_chars) > 0:
```
andrewhead: @zkirby isn't this redundant with Line 78?

data-processing/common/parse_equation.py
zkirby: consecutive chars can be anything in CHARACTER_TAGS :)

data-processing/common/parse_equation.py
zkirby: There is the possibility that the last symbol was also included in consecutive_chars (i.e. went through the if statement at line 71). In that case we need to clean up consecutive_chars and build anything that could be in it.

data-processing/common/parse_equation.py
zkirby: Ah, I was debating passing in the soup object itself or the soup.new_tag('mi'). I felt that the latter made more sense with regards to what the function needed to know.

data-processing/common/parse_equation.py
zkirby: Yeah I'm going to rewrite this statement as a whole. Basically there are four cases that need to be addressed (realizing now that this can just be put into two cases):
1. The symbol we are one is part of a consecutive sequence of symbols 
    1a) that consecutive sequence is our current consecutive sequence.
     1b) we had a consecutive sequence, but this one is part of a different consecutive sequence. 
2. The symbol is not part of a consecutive sequence
   2a) We were building a consecutive sequence and thus now need to add it in before we build our current symbol
   2b) We were not building a consecutive sequence, business as usual.

data-processing/common/parse_equation.py
zkirby: see comment below. I'm going to rewrite this statement for clarity :)

data-processing/common/parse_equation.py
```diff
@@ -51,6 +51,10 @@ def get_characters(mathml: str) -> List[Character]:
 def get_symbols(mathml: str) -> List[Symbol]:
     soup = BeautifulSoup(mathml, "lxml")
     node_symbols = []
+    consecutive_char_sequence: List[BeautifulSoup] = [] # The symbol_nodes of consecutive characters
+    consecutive_char_indices: List[int] = [] # Indices of the characters of a consecutive character sequence
+    # It's 'safe' to call soup.find_all as the underlying bs4 implementation 
+    # appears to preserve order https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py
```
andrewhead: Thanks for this!! 👍 ⚡️

Issue #65, opened 2020-03-13
Fix up sentence splitting


andrewhead:
Our TeX sentence detector returns the wrong character offsets for sentences. This impacts the user experience, as we cannot detect the locations of sentences (and therefore make those sentences interactive) if we don't have accurate offsets for them in the TeX.

The cause of the issue is that when we call the `pysbd` for sentence splitting, the character offsets for the `start` of sentences are sometimes incorrect. Specifically, the `start` index is sometimes erroneously set to 0. See [this issue](https://github.com/nipunsadvilkar/pySBD/issues/49) for details about when and why this occurs.

**Spec**. In a single example, here's the desired behavior. When we run the following code:

```python
from entities.sentences.extractor import SentenceExtractor
extractor = SentenceExtractor()
sentences = extractor.parse(
    "filename.tex",
    "This is a sentence. Next we describe two items. 1) The first item. 2) The second item."
    )

for s in sentences:
    print(s)
```

We _don't_ want this behavior (the current, undesired, output):

```
Sentence(start=0, end=19, tex_path='filename.tex', id_='0', tex='This is a sentence.', context_tex='This is a sentence.', text='This is a sentence.')
Sentence(start=20, end=47, tex_path='filename.tex', id_='1', tex='', context_tex='', text='Next we describe two items.')
Sentence(start=0, end=18, tex_path='filename.tex', id_='2', tex='', context_tex='', text='1) The first item.')
Sentence(start=0, end=19, tex_path='filename.tex', id_='3', tex='', context_tex='', text='2) The second item.')
```

The _desired output_ is (note the corrected `start` and `end` indices of the 3rd and 4th sentences):

```
Sentence(start=0, end=19, tex_path='filename.tex', id_='0', tex='This is a sentence.', context_tex='This is a sentence.', text='This is a sentence.')
Sentence(start=20, end=47, tex_path='filename.tex', id_='1', tex='', context_tex='', text='Next we describe two items.')
Sentence(start=48, end=66, tex_path='filename.tex', id_='2', tex='', context_tex='', text='1) The first item.')
Sentence(start=67, end=86, tex_path='filename.tex', id_='3', tex='', context_tex='', text='2) The second item.')
```

**Implementation options**. The only requirements for this issue are that:

- [ ] We should still use pysbd, as it is a high-quality sentence splitter
- [ ] The sentence extractor returns the desired output above

[...]

PR #66, opened 2020-03-16
Fix up sentence splitting

zkirby:
Please see #65 for further discussion. This PR fixes the issues with the sentence extractor start and end but continues the assumptions made in #65.

### DIFF
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -47,14 +47,15 @@ class SentenceExtractor(EntityExtractor):
             plaintext_to_tex_offset_map[len(plaintext)] = last_segment.tex_end
         # Segment the plaintext. Return offsets for each setence relative to the TeX input
         segmenter = pysbd.Segmenter(language="en", clean=False, char_span=True)
-        # Record the previous sentence end to account for extractor bug
-        pre_sentence_end = -1
+        # Record the current length of the plain text so account for the extractor bug
+        length_so_far_in_plain_text = -1
         for i, sentence in enumerate(segmenter.segment(plaintext)):
-            length = plaintext_to_tex_offset_map[sentence.end] - plaintext_to_tex_offset_map[sentence.start]
             # Since the sentence extractor has several bugs related to finding start and end indicies
             # we will simply set them ourselves.
-            start = pre_sentence_end + 1
-            end = start + length
+            start = plaintext_to_tex_offset_map[length_so_far_in_plain_text + 1]
+            end = plaintext_to_tex_offset_map[sentence.end + (length_so_far_in_plain_text + 1 - sentence.start)]
+            length_so_far_in_plain_text += sentence.end - sentence.start + 1
+
             tex = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
             pre_sentence_end = end
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -9,18 +9,18 @@ from common.types import SerializableEntity
 # These are 'reserved characters' by the pysbd module and can potentially
 # cause issues if they are present in a string. This list was compiled from the
 # psybd source code as of 3/23/20. locations:
-# ∯: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/abbreviation_replacer.py, https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/lists_item_replacer.py
-# ȸ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/processor.py
-# ♨: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/lists_item_replacer.py
-# ☝: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/lists_item_replacer.py
-# ✂: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/lists_item_replacer.py
-# ⎋: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓰ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓱ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓳ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓴ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/processor.py, https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓷ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/cleaner.py, https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
-# ᓸ: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/processor.py, https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/punctuation_replacer.py
+# ∯: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/abbreviation_replacer.py, https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/lists_item_replacer.py
+# ȸ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/processor.py
+# ♨: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/lists_item_replacer.py
+# ☝: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/lists_item_replacer.py
+# ✂: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/lists_item_replacer.py
+# ⎋: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓰ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓱ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓳ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓴ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/processor.py, https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓷ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/cleaner.py, https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
+# ᓸ: https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/processor.py, https://github.com/nipunsadvilkar/pySBD/blob/master/pysbd/punctuation_replacer.py
 PYSBD_RESERVED_CHARACTERS: List[str] = ["∯", "ȸ", "♨", "☝", "✂", "⎋", "ᓰ", "ᓱ", "ᓳ", "ᓴ", "ᓷ", "ᓸ"]
 
 @dataclass(frozen=True)
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -58,7 +58,7 @@ class SentenceExtractor(EntityExtractor):
             end = plaintext_to_tex_offset_map[real_end]
             length_so_far_in_plain_text = real_end
 
-            tex = tex[start:end]
+            tex_sub = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
             yield Sentence(
                 text=sentence.sent,
@@ -66,6 +66,6 @@ class SentenceExtractor(EntityExtractor):
                 end=end,
                 id_=str(i),
                 tex_path=tex_path,
-                tex=tex,
+                tex=tex_sub,
                 context_tex=context_tex,
             )
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -45,16 +45,19 @@ class SentenceExtractor(EntityExtractor):
 
         if last_segment is not None:
             plaintext_to_tex_offset_map[len(plaintext)] = last_segment.tex_end
-
         # Segment the plaintext. Return offsets for each setence relative to the TeX input
         segmenter = pysbd.Segmenter(language="en", clean=False, char_span=True)
+        # Record the previous sentence end to account for extractor bug
+        pre_sentence_end = -1
         for i, sentence in enumerate(segmenter.segment(plaintext)):
-
-            start = plaintext_to_tex_offset_map[sentence.start]
-            end = plaintext_to_tex_offset_map[sentence.end]
+            length = plaintext_to_tex_offset_map[sentence.end] - plaintext_to_tex_offset_map[sentence.start]
+            # Since the sentence extractor has several bugs related to finding start and end indicies
+            # we will simply set them ourselves.
+            start = pre_sentence_end + 1
+            end = start + length
             tex = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
-
+            pre_sentence_end = end
             yield Sentence(
                 text=sentence.sent,
                 start=start,
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -9,7 +9,7 @@ from common.types import SerializableEntity
 # These are 'reserved characters' by the pysbd module and can potentially
 # cause issues if they are present in a string. This list was compiled from the
 # psybd source code as of 3/23/20
-PSYDB_RESERVED_CHARACTERS: List[str] = ["∯", "ȸ", "♨", "☝", "✂", "⎋", "ᓰ", "ᓱ", "ᓳ", "ᓴ", "ᓷ", "ᓸ"]
+PYSBD_RESERVED_CHARACTERS: List[str] = ["∯", "ȸ", "♨", "☝", "✂", "⎋", "ᓰ", "ᓱ", "ᓳ", "ᓴ", "ᓷ", "ᓸ"]
 
 @dataclass(frozen=True)
 class Sentence(SerializableEntity):
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -58,7 +58,6 @@ class SentenceExtractor(EntityExtractor):
 
             tex = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
-            pre_sentence_end = end
             yield Sentence(
                 text=sentence.sent,
                 start=start,
--- a/data-processing/entities/sentences/extractor.py
+++ b/data-processing/entities/sentences/extractor.py
@@ -52,9 +52,11 @@ class SentenceExtractor(EntityExtractor):
         for i, sentence in enumerate(segmenter.segment(plaintext)):
             # Since the sentence extractor has several bugs related to finding start and end indicies
             # we will simply set them ourselves.
-            start = plaintext_to_tex_offset_map[length_so_far_in_plain_text + 1]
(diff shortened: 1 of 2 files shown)


data-processing/entities/sentences/extractor.py
```diff
@@ -45,16 +45,19 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
-
         # Segment the plaintext. Return offsets for each setence relative to the TeX input
         segmenter = pysbd.Segmenter(language="en", clean=False, char_span=True)
+        # Record the previous sentence end to account for extractor bug
+        pre_sentence_end = -1
         for i, sentence in enumerate(segmenter.segment(plaintext)):
-
-            start = plaintext_to_tex_offset_map[sentence.start]
-            end = plaintext_to_tex_offset_map[sentence.end]
+            length = plaintext_to_tex_offset_map[sentence.end] - plaintext_to_tex_offset_map[sentence.start]
```
andrewhead: @zkirby this `length` computation doesn't seem quite right.

As an example, consider the two sentences `hi \m. 123) hi.` Because plaintext strips out macros (like `\m`), the characters, their indexes in the TeX, and their indexes in the plain text will be:

```
character   TeX index    Plaintext index
=========   =========    ===============
h           0            0
i           1            1
            2            2
\           3            -
m           4            -
.           5            3
            6            4
1           7            5
2           8            6
3           9            7
)           10           8
            11           9
h           12           10
i           13           11
.           14           12
```

Say we're looking for the `length` of the second sentence, but `pysbd` incorrectly sets `sentence.start, sentence.end` to (0, 8) (wrong start offset, correct length). This would get mapped back to TeX positions 0-10 (follow the map from the "Plaintext index" to the "TeX index" column above). Based on this, the length of `123) hi` would be estimated to be "10 - 0" = 10, which is two more than it's supposed to be. Do I have the right interpretation of how length is getting calculated?

zkirby: Hmm I think I might be a tad confused, would this be solved by just using the sentence.start and sentence.end instead? Or am I misunderstanding

zkirby: Or I am confused by the current tests, which appear to include the macros in the length calculation? Take `test_extract_sentences` for example: 
```
extractor = SentenceExtractor()
 sentences = list(
    extractor.parse(
            "main.tex",
            "This is the first \\macro[arg]{sentence}. This is the second sentence.",
        )
    )
sentence1 = sentences[0]
assert sentence1.start == 0
assert sentence1.end == 40
assert sentences[0].text == "This is the first argsentence."
```
Which shows that the macros are taken into account for the end calculation

andrewhead: Yup, that's right! The sentence positions returned by the SentenceExtractor should include the macros. The only thing is, if `pysbd` returns the wrong `sentence.start` and `setnence.end` position (i.e. returns a `start` of 0, as in the example above), if there are macros in the TeX, `plaintext_to_tex_offset_map[sentence.end] - plaintext_to_tex_offset_map[sentence.start]` might return the wrong sentence length.

zkirby: I am realizing now - unless I am misunderstanding - we can't actually do a full tex-string-search right? Bc we don't actually split on the sentences with tex in them so we don't actually have that sentence and in some cases, it will be impossible to find that sentence

zkirby:
we'll that ended up being slightly conceptually more difficult than I had envisioned but fun to work on! 

I ended up not being able to use the string search solution because I wouldn't be able to tell the tex distinctions in the sentences (i.e. the first word in the sentence of the tex could be different than that of the plaintext). But I believe this new push is the solution

data-processing/entities/sentences/extractor.py
```diff
@@ -45,16 +45,20 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
         # Segment the plaintext. Return offsets for each setence relative to the TeX input
         segmenter = pysbd.Segmenter(language="en", clean=False, char_span=True)
+        # Record the current length of the plain text so account for the extractor bug
+        length_so_far_in_plain_text = -1
         for i, sentence in enumerate(segmenter.segment(plaintext)):
+            # Since the sentence extractor has several bugs related to finding start and end indicies
+            # we will simply set them ourselves.
+            start = plaintext_to_tex_offset_map[length_so_far_in_plain_text + 1]
+            end = plaintext_to_tex_offset_map[sentence.end + (length_so_far_in_plain_text + 1 - sentence.start)]
+            length_so_far_in_plain_text += sentence.end - sentence.start + 1
```
andrewhead: @zkirby I like the simplicity of the solution. Can you guarantee that these length computations will be correct? A couple cases that I expect could confuse this computation are:

* What if a sentence ends with a bunch of spaces (`Hello there     .`), does `.end` still end at the period?
* What if there is punctuation between sentences that's more than one character long (`Hello there!?`)

Can you take a look at the `pysbd` code to the extent that we know this is a robust solution?

data-processing/entities/sentences/extractor.py
```diff
@@ -45,16 +45,20 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
 
-            start = plaintext_to_tex_offset_map[sentence.start]
-            end = plaintext_to_tex_offset_map[sentence.end]
             tex = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
-
+            pre_sentence_end = end
```
andrewhead: nit pick: dead code?

data-processing/entities/sentences/extractor.py
zkirby: Hmmm there are definitely still sentences that break pysbd. In particular having multiple punctuation will still break it pysbd but if that is something we want to address it should probably be done in another PR since it's an issue with pysbd and not this particular solution.

I have noticed however that there is definitely a white space bug with this solution that I will investigate.

zkirby:
Okay I believe this solution is as optimal as possible. It works for all whitespace variants as well as for the list variant. It should be noted that this still does not work for multiple punctuation (but that is bc psybd/plain_text_mapping break so maybe we should open an issue for that

data-processing/entities/sentences/extractor.py
andrewhead: Roger, thanks for looking into this!

andrewhead:
> I ended up not being able to use the string search solution because I wouldn't be able to tell the tex distinctions in the sentences (i.e. the first word in the sentence of the tex could be different than that of the plaintext). But I believe this new push is the solution

Just to give this a bit more thought, would a search-based solution be possible if we did the search on the plain text string instead? Like this:

```
for sentence in [sentences extracted with pysbd]

  if (sentence is not the first sentence)

    if (sentence.start is 0)
-     real-start := (search  full-tex-string, starting at last-sentence.end)
+     real-start := (search  plain-tex-string, starting at last-sentence.end)
      real-end := (real-start) + (sentence.end - sentence.start)
      update sentence start and end to real-start and real-end

  last-sentence := sentence

+ for each sentence, map its real-start and real-end to positions in the TeX string
```

One potential advantage of this approach is that it would be a little more resistant to error; with the current solution if we do get the length of one of the sentences wrong, wouldn't that result in getting the positions of all following sentences incorrect?

zkirby:
Gotcha, switched over to that version :)

data-processing/entities/sentences/extractor.py
```diff
@@ -45,22 +45,27 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+            real_start = plaintext.find(sentence.sent, length_so_far_in_plain_text)
+            start = plaintext_to_tex_offset_map[real_start]
+            real_end = real_start + (sentence.end - sentence.start)
+            end = plaintext_to_tex_offset_map[real_end]
+            length_so_far_in_plain_text = real_end
 
-            start = plaintext_to_tex_offset_map[sentence.start]
-            end = plaintext_to_tex_offset_map[sentence.end]
-            tex = tex[start:end]
+            tex_sub = tex[start:end]
```
zkirby: also I was pretty sure this was a bug so I went ahead and changed it, but the tex/tex_context were blank for everything that wasn't the first string, was this intentional?

andrewhead: Oh yikes! Yes that was definitely a bug 🐞 . Great catch!

data-processing/entities/sentences/extractor.py
```diff
@@ -45,22 +45,27 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
             plaintext_to_tex_offset_map[len(plaintext)] = last_segment.tex_end
-
         # Segment the plaintext. Return offsets for each setence relative to the TeX input
         segmenter = pysbd.Segmenter(language="en", clean=False, char_span=True)
+        # Record the current length of the plain text so account for the extractor bug
+        length_so_far_in_plain_text = 0
         for i, sentence in enumerate(segmenter.segment(plaintext)):
+            # Since the sentence extractor has several bugs related to finding start and end indicies
+            # we will simply set them ourselves.
+            real_start = plaintext.find(sentence.sent, length_so_far_in_plain_text)
```
andrewhead: This looks great! A couple of questions to make this code super robust:

1. Can you add a check for the return value `-1` (when the sentence is not found) with a `logging.warning` statement? I don't expect this to happen, but over millions of papers it may come up.

2. Can you check to see if `pysbd` ever changes the text in the `sentence.sent` from how the text is written in the input text (i.e. does it remove invalid characters from `sent`?) If the sentences will always be a substring of the plaintext, this looks just fine! If the sentences are sometimes transformed, then `find` will definitely return `-1` sometimes. Then we should add another rule that does something useful if `find` returns `-1`), for example, sets `real_start` at an offset of one character after the last sentence's `real_end`.

zkirby: Gotcha :). 
1. I know for a fact after looking through the actual pysbd that it will indeed remove some very very obscure special characters (these are used in the actual pysbd implementation). Maybe I can combine your point number 1 and 2 and just have a logging.warning statement that says "Warning failed to parse sentence, skipping sentence"?
2. I also wanted to reiterate that psybd fails for sentences with more than one punctuation that are not periods (say !! or !?). I'm honestly not really sure how to handle this, throwing the psybd parser in a try catch sounds like it could be very expensive over thousands of iterations, but I'm not sure if there is a better way? Maybe just write a regex expr and just don't search the string if it's true?

andrewhead: > throwing the psybd parser in a try catch sounds like it could be very expensive over thousands of iterations, but I'm not sure if there is a better way?

This would probably be fine, as I imagine the exceptions would be pretty rare, and string searching is one of the fastest parts of the pipeline currently.

> Maybe just write a regex expr and just don't search the string if it's true?

I think this could work! What type of regex do you have in mind?

zkirby: Gotcha, I'll just the try catch then bc it's going to be more robust then a regex

zkirby: Actually I guess there really isn't a point to a try catch now that I think about it? We would probably want it to fail as otherwise the rest of the pipeline wont work and it will be mislead

andrewhead: @zkirby I think the hope is we can keep extracting as many other sentences as we can even if there's a failure for one sentence. So I think the ideal behavior is:

* flag if the positions couldn't be recovered for this sentence
* keep going, recovering the correct position of the sentences after this one if we can

andrewhead: There are tiny exceptions like this everywhere in the pipeline, so the typical approach has been to try to catch unexpected errors, log them, and keep processing as much of the paper as possible. This tends to work pretty well in papers where there are hundreds of things that need to be processed.

zkirby: Gotcha! I think the double punctuation is actually another indexing error unfortunately which would mean using a try/catch for every sentence. Not sure if this is okay?

zkirby: Probably better to open another PR and try to fix it?

andrewhead: @zkirby does the double punctuation cause an exception, or just split sentences in a way that weren't expecting?

andrewhead: I'm game with rolling it into this PR, I'm just realizing I didn't catch your meaning about the double punctuation initially.

zkirby: It throws an index out of bounds in plaintext_to_tex_offset_map. Any ideas?

andrewhead: Just getting a bit more context (and also sorry if you've already answered this XD), why does it cause an index out of bounds exception? Can you give me a brief code example?

zkirby: Ah, as far as why I'm not sure I can spend sometime investigating right now if we are going to include it in the PR, but for example a sentence like 'hello. world!!!' returns: 
```
File "/data-processing/entities/sentences/extractor.py", line 54, in parse
    end = plaintext_to_tex_offset_map[sentence.end]
KeyError: 21
```

data-processing/entities/sentences/extractor.py
```diff
@@ -1,11 +1,15 @@
...
+from typing import Iterator, List
 
 import pysbd
 
 from common.parse_tex import DEFAULT_CONTEXT_SIZE, EntityExtractor, PlaintextExtractor
 from common.types import SerializableEntity
-
+# These are 'reserved characters' by the pysbd module and can potentially
+# cause issues if they are present in a string. This list was compiled from the
+# psybd source code as of 3/23/20
```
andrewhead: Are all of these characters from different places in the code? I wonder if we can provide permalinks back to the locations where you found them (which I realize might be half a dozen permalinks) so someone can go back and check on whether they have changed later.

data-processing/entities/sentences/extractor.py
```diff
@@ -1,11 +1,15 @@
...
+PSYDB_RESERVED_CHARACTERS: List[str] = ["∯", "ȸ", "♨", "☝", "✂", "⎋", "ᓰ", "ᓱ", "ᓳ", "ᓴ", "ᓷ", "ᓸ"]
```
andrewhead:
```suggestion
PYSBD_RESERVED_CHARACTERS: List[str] = ["∯", "ȸ", "♨", "☝", "✂", "⎋", "ᓰ", "ᓱ", "ᓳ", "ᓴ", "ᓷ", "ᓸ"]
```

data-processing/entities/sentences/extractor.py
```diff
@@ -20,6 +24,10 @@ class SentenceExtractor(EntityExtractor):
     """
 
     def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
+        for reserved_char in PSYDB_RESERVED_CHARACTERS:
+            if reserved_char in tex:
+                logging.warning('reserved character "%s" found in tex string, this might break the sentence extractor.', reserved_char)
```
andrewhead:
```suggestion
                logging.warning('Reserved character from pysbd "%s" found in tex string, this might break the sentence extractor.', reserved_char)
```

data-processing/entities/sentences/extractor.py
```diff
@@ -50,14 +58,29 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+            #    ex: ". hello. world. hi."
+            #    sol: check for sentences being longer than 1000 characters.
+            # 2. pysbd indexes are sometimes incorrectly set
+            #    ex: "hello. world. 1) item one. 2) item two. 3) item three" or "hello!!! world."
+            #    sol: set indexes manually using string search + sentence length
+            # 3. pysbd uses reserved characters for splitting sentences
+            #    ex: see PSYDB_RESERVED_CHARACTERS list.
+            #    sol: throw a warning if the sentence contains any of these characters.
+            if len(sentence.sent) > 1000:
+                logging.warning('exceptionally long sentence, this might indicate the sentence extractor is broken')
```
andrewhead:
```suggestion
                logging.warning('Exceptionally long sentence (length %d), this might indicate the sentence extractor failed to properly split text into sentences.', len(sentence.sent))
```

andrewhead: Just making some small updates to the logging---capitalizing it, and trying to clarify the log message.

data-processing/entities/sentences/extractor.py
```diff
@@ -50,14 +58,29 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+
             real_start = plaintext.find(sentence.sent, length_so_far_in_plain_text)
```
andrewhead: Now that we're no longer using `sentence.start` and `sentence.end`, these `real_` prefixes don't make much sense. Maybe just make them `start` and `end`?

data-processing/entities/sentences/extractor.py
```diff
@@ -50,14 +58,29 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
         # Record the current length of the plain text so account for the extractor bug
```
andrewhead: We no longer need the `char_span=True` argument in `Segmenter` since we are calculating offsets ourselves.

data-processing/entities/sentences/extractor.py
```diff
@@ -50,14 +58,29 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+            # 3. pysbd uses reserved characters for splitting sentences
+            #    ex: see PSYDB_RESERVED_CHARACTERS list.
+            #    sol: throw a warning if the sentence contains any of these characters.
+            if len(sentence.sent) > 1000:
+                logging.warning('exceptionally long sentence, this might indicate the sentence extractor is broken')
+
             real_start = plaintext.find(sentence.sent, length_so_far_in_plain_text)
+            real_end = real_start + len(sentence.sent)
+            if real_start not in plaintext_to_tex_offset_map or real_end not in plaintext_to_tex_offset_map:
+                logging.warning('a sentence boundary was incorrect, this is probably an issue with pysbd. Skipping sentence in extractor')
```
andrewhead:
```suggestion
                logging.warning('A sentence boundary was incorrect for sentence %s. This is probably an issue with pysbd. Skipping sentence in extractor.', sentences.sent)
```

data-processing/entities/sentences/extractor.py
andrewhead: @zkirby Can you add one more check here, to make sure that the `start` located is no more than, say, 500 characters or so after the `end` of the last sentence? This way if there's a super common sentence like say "QED." then this won't skip over huge parts of the file if it fails to find the first instance it was looking for.

data-processing/entities/sentences/extractor.py
zkirby: A yes I can add those

data-processing/entities/sentences/extractor.py
zkirby: I'll just call them plain_text_start/end

data-processing/entities/sentences/extractor.py
```diff
@@ -67,24 +79,26 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
-            real_end = real_start + len(sentence.sent)
-            if real_start not in plaintext_to_tex_offset_map or real_end not in plaintext_to_tex_offset_map:
-                logging.warning('A sentence boundary was incorrect for sentence %s. This is probably an issue with pysbd. Skipping sentence in extractor.', sentences.sent)
+            plaintext_start = plaintext.find(sentence, length_so_far_in_plain_text)
+            plaintext_end = plaintext_start + len(sentence)
+            if plaintext_start not in plaintext_to_tex_offset_map or plaintext_end not in plaintext_to_tex_offset_map:
+                logging.warning('A sentence boundary was incorrect for sentence %s. This is probably an issue with pysbd. Skipping sentence in extractor.', sentence)
                 continue
+            if plaintext_start - 500 > length_so_far_in_plain_text:
+                logging.warning('Sentence boundary start for sentence %s was %d characters a head of the previous sentence, this might indicate the sentence extractor failed to properly split text.', sentence, plaintext_start - length_so_far_in_plain_text)
```
andrewhead:
```suggestion
                logging.warning('Sentence boundary start for sentence %s was %d characters ahead of the previous sentence, this might indicate the sentence extractor failed to properly split text.', sentence, plaintext_start - length_so_far_in_plain_text)
```

data-processing/entities/sentences/extractor.py
```diff
@@ -67,24 +79,26 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
             #    ex: "hello. world. 1) item one. 2) item two. 3) item three" or "hello!!! world."
             #    sol: set indexes manually using string search + sentence length
             # 3. pysbd uses reserved characters for splitting sentences
-            #    ex: see PSYDB_RESERVED_CHARACTERS list.
+            #    ex: see PYSBD_RESERVED_CHARACTERS list.
```
andrewhead: good catch!

data-processing/entities/sentences/extractor.py
```diff
@@ -8,7 +8,19 @@
 from common.types import SerializableEntity
 # These are 'reserved characters' by the pysbd module and can potentially
 # cause issues if they are present in a string. This list was compiled from the
-# psybd source code as of 3/23/20
+# psybd source code as of 3/23/20. locations:
+# ∯: https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/abbreviation_replacer.py, https://github.com/nipunsadvilkar/pySBD/blob/npn-carriage-return-fix/pysbd/lists_item_replacer.py
```
andrewhead: @zkirby awesome!! can you make these permalinks, anchored to the specific version and line, e.g., https://github.com/nipunsadvilkar/pySBD/blob/995af9e3c3d324aa053cb06f79c0038c882f633d/pysbd/lists_item_replacer.py#L21 ? To do this, you find the line on GitHub, click on the line number, click on the ellipses that appear, and then click "Copy permalink":

image

zkirby: Oh dang hahaha, how important is it to do this? I'm not sure it will be super helpful bc the files themselves are quite short and the symbols are scattered many times throughout the files.

andrewhead: not that important! let's skip it and merge this work of art into the code base.

PR #68, opened 2020-03-20
Tooltip for citations without any linked paper

jocelynzsun:
This catches the fatal error threw when the api fails to fetch the paper linked from the server. With this implementation, a tooltip appears even when there isn't any linked information, and the user is encouraged to provide feedback on the tooltip and (hopefully) the missing paper.

(The screenshot is made by simulating the bug... it's not an actual bug now):

### DIFF
--- a/ui/src/Citation.tsx
+++ b/ui/src/Citation.tsx
@@ -16,9 +16,9 @@ export class Citation extends React.PureComponent<CitationProperties> {
           if (paper === undefined) {
             return (
               <div>
-                Sorry, we currently don't have any linked information for this paper yet.<br></br>
-                Let us know what went wrong by clicking on the icon below, so we could provide 
-                you with better reading experience in the future!
+                <p>Sorry, we currently don't have any linked information for this paper yet.</p>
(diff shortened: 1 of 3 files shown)


ui/src/Citation.tsx
```diff
@@ -13,31 +13,41 @@ export class Citation extends React.PureComponent<CitationProperties> {
...
-                </div>
-              ) : null}
-            </div>
-          );
+          if (paper === undefined) {
+            return (
+              <div>
+                Sorry, we currently don't have any linked information for this paper yet.<br></br>
+                Let us know what went wrong by clicking on the icon below, so we could provide 
+                you with better reading experience in the future!
```
andrewhead: Small style fix: let's use `<p>` tags, so that we preserve the inter-paragraph spacing determined in the CSS.

```suggestion
                <p>Sorry, we currently don't have any linked information for this paper yet.</p>
                <p>Let us know what went wrong by clicking on the icon below, so we could provide 
                you with better reading experience in the future!</p>
```

PR #71, opened 2020-03-21
Batch processing of API requests for paper details

jocelynzsun:
Batching API request for paper details through slicing list of paper id's into smaller chunks, thus making the request size smaller. 

Screenshots:

Using paper 2002.11102 as example, this paper's `s2ids` has 149 items (149 paper id's requested). Using a simple array slicing algorithm, the array is sliced into 3 arrays of 50, 50, 49 items respectively. These three arrays is transformed into 3 promises to use with `Promise.all()`, and the result is flattened into one long array.

### DIFF
--- a/ui/src/Citation.tsx
+++ b/ui/src/Citation.tsx
@@ -16,9 +16,9 @@ export class Citation extends React.PureComponent<CitationProperties> {
           if (paper === undefined) {
             return (
               <div>
-                Sorry, we currently don't have any linked information for this paper yet.<br></br>
-                Let us know what went wrong by clicking on the icon below, so we could provide 
-                you with better reading experience in the future!
+                <p>Sorry, we currently don't have any linked information for this paper yet.</p>
+                <p>Let us know what went wrong by clicking on the icon below, so we could provide 
+                you with better reading experience in the future!</p>
               </div>
             )
           } else {
--- a/ui/src/Citation.tsx
+++ b/ui/src/Citation.tsx
@@ -13,31 +13,41 @@ export class Citation extends React.PureComponent<CitationProperties> {
       <ScholarReaderContext.Consumer>
         {({ papers, setJumpPaperId, setSelectedCitation }) => {
           const paper = papers[this.props.citation.paper];
-          return (
-            <div className="citation">
-              <p className="title">{paper.title}</p>
-              {paper.authors.length > 0 && (
-                <>
-                  by {paper.authors[0].name}{" "}
-                  {paper.authors.length > 1 && " et al."}
-                </>
-              )}
-              {paper.abstract !== null ? (
-                <div className="citation__abstract">
-                  {truncateText(paper.abstract, 400)}
-                  <span
-                    className="citation__abstract__sidebar-link"
-                    onClick={() => {
-                      setSelectedCitation(this.props.citation);
-                      setJumpPaperId(this.props.citation.paper);
-                    }}
-                  >
-                    (see more)
-                  </span>
-                </div>
-              ) : null}
-            </div>
-          );
+          if (paper === undefined) {
+            return (
+              <div>
+                Sorry, we currently don't have any linked information for this paper yet.<br></br>
+                Let us know what went wrong by clicking on the icon below, so we could provide 
+                you with better reading experience in the future!
+              </div>
+            )
+          } else {
+            return (
+              <div className="citation">
+                <p className="title">{paper.title}</p>
+                {paper.authors.length > 0 && (
+                  <>
+                    by {paper.authors[0].name}{" "}
+                    {paper.authors.length > 1 && " et al."}
+                  </>
+                )}
+                {paper.abstract !== null ? (
+                  <div className="citation__abstract">
+                    {truncateText(paper.abstract, 400)}
+                    <span
+                      className="citation__abstract__sidebar-link"
+                      onClick={() => {
+                        setSelectedCitation(this.props.citation);
+                        setJumpPaperId(this.props.citation.paper);
+                      }}
+                    >
+                      (see more)
+                    </span>
+                  </div>
+                ) : null}
+              </div>
+            );
+          }
         }}
       </ScholarReaderContext.Consumer>
     );
--- a/ui/src/api.ts
+++ b/ui/src/api.ts
@@ -25,14 +25,8 @@ export async function getAllPapers() {
 }
 
 export async function papers(s2Ids: string[]) {
-  const data = await doGet(
-    axios.get("/api/v0/papers", {
-      params: {
-        id: s2Ids.join(",")
-      }
-    })
-  );
-  return (data || []) as Paper[];
+  var promises = collectPromise(s2Ids, 50);
+  return Promise.all(promises).then(responses => (responses as Paper[][]).reduce((a, b) => a.concat(b), []));
 }
 
 export async function symbolsForArxivId(arxivId: string) {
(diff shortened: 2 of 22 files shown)


andrewhead:
Great documentation!

ui/src/api.ts
```diff
@@ -25,14 +25,8 @@ export async function getAllPapers() {
...
 export async function papers(s2Ids: string[]) {
-  const data = await doGet(
-    axios.get("/api/v0/papers", {
-      params: {
-        id: s2Ids.join(",")
-      }
-    })
-  );
-  return (data || []) as Paper[];
+  var promises = collectPromise(s2Ids, 50);
```
andrewhead: @jocelynzsun can you pull this out into a named constant, with a detailed commend describing why we needed to split up the list of papers for the API call in the first place (i.e. describe the bug we were seeing)?

ui/src/api.ts
```diff
@@ -25,14 +25,8 @@ export async function getAllPapers() {
...
+  return Promise.all(promises).then(responses => (responses as Paper[][]).reduce((a, b) => a.concat(b), []));
```
andrewhead: Would this work?

```suggestion
  return Promise.all(promises).then(responses => (responses.flat() as Paper[]));
```

ui/src/api.ts
```diff
@@ -123,3 +117,36 @@ async function doGet<T>(get: Promise<AxiosResponse<T>>) {
   }
   return null;
 }
+
+/** 
+ * helper function for slicing arrays.
+ */
+function sliceArray(array: string[], size: number) {
```
andrewhead: @jocelynzsun can we name this something like "splitArray" instead? Slice is an operation that already has a well-defined, different set of behavior than this (see `Array.slice` documentation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice)).

Because we're not calling it elsewhere, actually, maybe we should just get rid of this function and include the code itself in the `collectPromise` method, to keep from unnecessary code fragmentation?

ui/src/api.ts
```diff
@@ -123,3 +117,36 @@ async function doGet<T>(get: Promise<AxiosResponse<T>>) {
...
+}
+
+/** 
+ * Collect array of promises generated based on array of paper ids.
+ */
+function collectPromise(papers:string[], size:number) {
+  var paperChuck = sliceArray(papers, size);
+  var promises = [];
+
+  for (var i = 0; i < paperChuck.length; i++) {
```
andrewhead: is this a typo, should this be `paperChunk`?

ui/src/api.ts
```diff
@@ -123,3 +117,36 @@ async function doGet<T>(get: Promise<AxiosResponse<T>>) {
...
+    result.push(array.slice(i, i+size));
+  }
+
+  return result
+}
+
+/** 
+ * Collect array of promises generated based on array of paper ids.
+ */
+function collectPromise(papers:string[], size:number) {
```
andrewhead: Similarly to the comment for `sliceArray`, can you try just putting this into the body of the `papers` method? I think the functionality is small enough and relevant enough to fetch papers that it might make sense to combine it all in one place.

ui/src/api.ts
```diff
@@ -123,3 +117,36 @@ async function doGet<T>(get: Promise<AxiosResponse<T>>) {
...
 }
+
+/** 
+ * helper function for slicing arrays.
+ */
+function sliceArray(array: string[], size: number) {
+  var result = [];
+
+  for (var i = 0; i < array.length; i += size) {
+    result.push(array.slice(i, i+size));
```
andrewhead: TODO for @andrewhead: get Jocelyn the style settings for the project

ui/src/api.ts
andrewhead: You can see some examples of named constants here in this code https://github.com/allenai/scholar-reader/blob/2412d2364be3783166456b9e806a23d4596752a9/ui/src/ScholarReader.tsx#L420-L421

ui/src/api.ts
jocelynzsun: Ooooooops yes!

ui/src/api.ts
jocelynzsun: How does this work after I combine all three functions into one?

ui/src/Citation.tsx
```diff
@@ -13,31 +13,41 @@ export class Citation extends React.PureComponent<CitationProperties> {
...
-                      setJumpPaperId(this.props.citation.paper);
-                    }}
-                  >
-                    (see more)
-                  </span>
-                </div>
-              ) : null}
-            </div>
-          );
+          if (paper === undefined) {
```
zkirby: @andrewhead wondering your thoughts on either sticking to var == null or var === undefined || var === null. I think we should maintain consistency across the UI as far as null checks :o.

ui/src/api.ts
andrewhead: Pretty much the same way! You would defined a constant (`e.g., API_PAPER_REQUEST_BATCH_SIZE`) to 50 at the top of the `papers/` function, and refer to that variable when you're splitting up the list of s2Ids.

ui/src/api.ts
andrewhead: @jocelynzsun okay, I have added style files for the project :-D Can you download this file, and then place it at `.vscode/settings.json` in the `ui/` folder? https://github.com/allenai/scholar-reader/blob/master/ui/.vscode/settings.json

andrewhead: Then I would recommend checking the box for "Format on Save" in the VSCode editor settings, and then saving this file so that it gets formatted to iron out small differences in code formatting :-)

ui/src/Citation.tsx
andrewhead: Good thinking, I'm always all for consistency! I think my (albeit weak) preference is to have explicit, strict-equality checks (`===`) for any types that the type definitions say are possible. In this case, the only value we would have to check against according to type definitions would be `undefined` (i.e. what you get when you look in a JavaScript object for a nonexistent property). What do you think of that policy? Is there a good argument in favor of the `var == null` check?

zkirby: No, i don't think so hahah (not unless anyone explicitly passes in null). Just wanted to be on the same page, since I've been leaning the var == null route haha

ui/src/api.ts
jocelynzsun: already done that! that's pretty much the only constant in the function?

andrewhead: ah I see! Did you push the updates? I don't think I'm seeing that version of the code in the PR, though I might just be looking at the wrong version of it XD

jocelynzsun: I'll push it now!

ui/src/api.ts
```diff
@@ -25,14 +25,25 @@ export async function getAllPapers() {
...
 export async function papers(s2Ids: string[]) {
-  const data = await doGet(
-    axios.get("/api/v0/papers", {
-      params: {
-        id: s2Ids.join(",")
-      }
-    })
-  );
-  return (data || []) as Paper[];
+  const PAPER_REQUEST_BATCH_SIZE = 50;
```
andrewhead: @jocelynzsun Can you add a comment here to describe why we need to batch the requests to the API (specifically, that it seems that in production, when more than about 50 papers are requested from the API, sometimes the API returns a 400?)

ui/src/api.ts
```diff
@@ -25,14 +25,25 @@ export async function getAllPapers() {
...
+  var paperChunk = [];
```
andrewhead: nit pick: how about this name, which conveys that the type of data is a group of S2 IDs, and uses the consistent terminology "batch"?

```suggestion
  var s2IdsBatch = [];
```

ui/src/api.ts
```diff
@@ -25,14 +25,25 @@ export async function getAllPapers() {
...
+  var promises = [];
+
+  for (var i = 0; i < s2Ids.length; i += PAPER_REQUEST_BATCH_SIZE) {
```
andrewhead: Can you change the variable declarations for the loop variable (here and on Line 36) to:

```suggestion
  for (let i = 0; i < s2Ids.length; i += PAPER_REQUEST_BATCH_SIZE) {
```

to make it clear that the scope of `i` is limited to the current block? (instead of the default global scope of `var` declarations? Check out https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let)

Issue #41, opened 2020-02-12
Include numeric subscripts and superscripts in symbols


andrewhead:
The bounding boxes for symbols currently don't include numeric subscripts and superscripts, but they probably should. This is to provide users with the right feel (to make it feel like the full symbol is getting highlighted). It's also important to include the superscript or subscript in the symbol in case it lends semantic meaning to the symbol.

As a concrete example, for arXiv paper `1811.12889`, only "s_k" gets highlighted when a user hovers over it, though we want the entirety of "s_k^3" to get highlighted.

![image](https://user-images.githubusercontent.com/2358524/74366437-de5ed480-4d84-11ea-8aca-bf7e443c003d.png)

Completion of this task requires:
- [ ] A change to the `get_symbols` method that extracts a symbol for `<mn>` numeric elements
- [ ] These numeric symbols get tagged with a property that they are numeric
- [ ] Make sure that numeric symbols are being added as children of the symbols that they are superscripts and subscripts to (this might already be implemented by default)
- [ ] Update the UI to make sure not to show bounding boxes for purely numeric symbols that are not children of other symbols (i.e. don't make a `1` or `0` interactive)

zkirby:
Haha before I dive into this too much, can you quickly remind me the difference between a character tag and a symbol tag?

zkirby:
Also slightly silly and nuanced but would we want to consider "1 + 2" a symbol (basically is a conjunction of multiple mn also a symbol)?

zkirby:
The reason I ask the above is I kind of see two approaches to this: 
1) Look for any mn and given some heuristic, decide if this particular mn should be a symbol.
2) Look through the descendent tree of a tag we've decided should be a symbol and 'symbolize' all of the descendents that should also now be symbols.

andrewhead:
> Haha before I dive into this too much, can you quickly remind me the difference between a character tag and a symbol tag?

Yeah, sure!

*Characters* are individual letters in the TeX for which we will detect the positions by coloring (e.g., in this case `x` or `2` in `x^2`. Character tags are any MathML tag that contains a single character that we will detect the position of.

*Symbols* are meaningful collections of *characters* that should be clickable in a paper. For instance `x^2`. Though `x` is also a symbol. Right now, we look for standalone letters (e.g., `<mi>x</mi>`), subscripts (e.g., `<msub>...</msub>`), and superscripts (`<msup>`) to indicate that all characters found within are part of the same "symbol".

[...]

PR #73, opened 2020-03-22
Include numeric subscripts and superscripts in symbols

zkirby:
Please see #41 for further discussion. Numbers are now considered symbols if and only if they are a direct descendent of a super script or sub script


andrewhead:
Thanks for linking to the issue!

data-processing/common/parse_equation.py
```diff
@@ -53,9 +57,12 @@ def get_symbols(mathml: str) -> List[Symbol]:
     node_symbols = []
     consecutive_char_sequence: List[BeautifulSoup] = [] # The symbol_nodes of consecutive characters
     consecutive_char_indices: List[int] = [] # Indices of the characters of a consecutive character sequence
-    # It's 'safe' to call soup.find_all as the underlying bs4 implementation 
+    # It's 'safe' to call soup.find_all as the underlying bs4 implementation
```
andrewhead: Hahaha I'm just noticing the foreboding quotes around 'safe' here XD

data-processing/common/parse_equation.py
```diff
@@ -53,9 +57,12 @@ def get_symbols(mathml: str) -> List[Symbol]:
...
     # appears to preserve order https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/bs4/element.py
     for symbol_node in soup.find_all(SYMBOL_TAGS):
+        # Skip partial symbols that aren't themselves descendants of another symbol
+        if symbol_node.name in PARTIAL_SYMBOLS and not symbol_node.parent.name in SYMBOL_TAGS:
+            continue
```
andrewhead: I like the brevity, kudos! Isn't another way of implementing this just adding `mn` to the list of `CHARACTER_TAGS`, or am I missing something? One advantage of that approach is that characters are already another form of "partial symbols" (i.e. they're the pieces symbols are made up of), so then we only one data structure to hold 'em all.

zkirby: Ah I see hahah, I think I misunderstood what characters were and what exactly symbols were but it appears that yes that is all that is needed LOL

data-processing/common/parse_equation.py
zkirby: XD

data-processing/common/parse_equation.py
andrewhead: Cool! My bad for my lackluster documentation.

data-processing/tests/mathml/x_sub_four_sub_three.xml
```diff
@@ -0,0 +1,17 @@
...
+            <mrow>
+                <msub s2:index="5">
+                    <mi s2:start="0" s2:end="1" s2:index="0">x</mi>
+                    <msub s2:index="4" s2:start="2" s2:end="7">
+                        <mn s2:start="3" s2:end="4" s2:index="1">4</mn>
+                        <mn s2:start="5" s2:end="6" s2:index="2">3</mn>
+                    </msub>
+                </msub>
+            </mrow>
+            <annotation encoding="application/x-tex">x_{t_i}</annotation>
```
andrewhead: nit pick: can you update this element to have the value `x_43`? It's not important for the sake of the test running, though might avoid future confusion about this test data.

data-processing/tests/test_parse_equation.py
```diff
@@ -86,6 +86,22 @@ def test_get_symbol_children():
...
+        symbols = get_symbols(mathml)
+
+    assert len(symbols) == 3
+    x_sub_four_sub_three = list(
+        filter(lambda s: "msub" in s.mathml and "x" in s.mathml, symbols)
+    )[0]
+    x = list(filter(lambda s: s.mathml == "<mi>x</mi>", symbols))[0]
+
+    assert len(x_sub_four_sub_three.children) == 2
+    assert x in x_sub_four_sub_three.children
```
andrewhead: Can you add an assertion to check here that the nodes `4` and `3` are among in `x_sub_four_sub_three.characters`?

data-processing/tests/test_parse_equation.py
```diff
@@ -86,6 +86,22 @@ def test_get_symbol_children():
...
     assert i in t_sub_i.children
 
+def test_get_symbol_sub_super_with_numeric_child():
+    with open(
+        get_test_path(os.path.join("mathml", "x_sub_four_sub_three.xml"))
+    ) as mathml_file:
+        mathml = mathml_file.read()
+        symbols = get_symbols(mathml)
+
+    assert len(symbols) == 3
```
andrewhead: This is a great test. It makes me realize, do we want `4_3` to be detected as a symbol?

I feel like I we probably don't want `4^3` to be detected as a symbol---that's just a number. `4_3` has no special meaning, and probably won't come up in papers. Though I'm curious what your thoughts are. I'm thinking we may want to add a rule to skip `msub` and `msup` elements that have an `mn` as their base.

zkirby: I think I mostly agree. However, I am interested to know the situations in which 4_3 (or for that matter any calculation involving just numeric symbols) would not have been reduced? My gut says that if there is an expression that has intentionally not been reduced, it holds a special significance in the paper and thus might be something someone wants to actually look for. I'm unsure, what do you think?

andrewhead: That's fair! Okay, then let's keep numbers that have a superscript or subscript. We can decide when testing out the tool what feels right.

zkirby:
Doing that right now, wanted to wait to see what we wanted to do for the fully numeric sub/sup :)

PR #88, opened 2020-03-28
Code review for updates to symbol detection

andrewhead:
Hey @zkirby! I updated how we extract symbols from MathML to try to make it more robust and extensible. The main challenge I was trying to address was that in our previous version with merging o consecutive symbols together, we didn't have the ability to assign the newly- merged symbol as a child to other symbols.

So, I rearchitected the algorithm a bit. Now it visit MathML nodes in the MathML. After visiting a child node, the visitor has a list of "child symbols". It will assign these child nodes the first ancestor that is considered to be a symbol (e.g., not `mrow`, but `msup` and `msub` both qualify as symbols).

Would you be willing to take a quick look at the updated `parse_equation.py` and let me know if this all looks about right? It passes the updated versions of the tests, including the ones you created for merging symbols. I'd love to get another pair of eyes on it as a sanity check.

(To clarify, there are a bunch of changes in this commit, though the only one I'd like your feedback on right now is `parse_equation.py`)


data-processing/common/parse_equation.py
```diff
@@ -1,151 +1,290 @@
...
+    tokens: List[Token]
+    """
+    List of all tokens found in the subtree starting at the parsed element.
+    """
+
+
+def parse_element(element: Tag) -> ParseResult:
+    """
+    Extract symbol nodes from an element in a BeautifulSoup MathML parse tree. This function
+    recursively visits
```
zkirby: nit: "This function recursively visits nodes"

data-processing/common/parse_equation.py
```diff
@@ -1,151 +1,290 @@
...
+    # Detect if this tag represents a KaTeX parse error. If so, return nothing.
+    if _is_error_element(element):
+        return ParseResult([], None, [])
+
+    # If this is an 'mrow' produced by KaTeX, its children are probably needlessly fragmented. For
+    # instance, the word 'true' will contain four '<mi>' elements, one for 't', 'r', 'u', and 'e'
+    # each. Merge such elements into single elements.
+    soup_children = element.children
+    if element.name == "mrow":
+        merged_children = MathMlElementMerger().merge(soup_children)
```
zkirby: curious why not make this a static method? Instead of spinning up a new instance on every merge. Or alternatively have the `soup_children` be passed in the constructor since the name implies that it will be merging some MathML elements

data-processing/common/parse_equation.py
```diff
@@ -1,151 +1,290 @@
...
+
+        # If the 'mrow' only contains one element after its children are merged, simplify the
+        # MathML tree replacing this node with its merged child.
+        if len(merged_children) == 1:
+            element = merged_children[0]
+            soup_children = element.children
```
zkirby: nit: I believe this line is redundant with line 93 since MathMlElementMerger does not modify it's input

data-processing/common/parse_equation.py
```diff
@@ -1,151 +1,290 @@
...
+
+def _extract_tokens(element: Tag) -> List[Token]:
+    """
+    Get the tokens defined in this element. Tokens are only found in low-level elements like
+    "<mi>" and "<mn>". This function will find no tokens in higher-level nodes that solely
+    group other low-level elements (like "<mrow>" and "<msub>").
+    """
+
+    tokens = []
+    if element.name in ["mi", "mn"] and _has_s2_token_annotations(element):
```
zkirby: curious why `["mi", "mn"]` is not a defined constant like SYMBOL_TAGS?

data-processing/common/parse_equation.py
```diff
@@ -1,151 +1,290 @@
...
+        tokens.append(
+            Token(
+                text=element.string,
+                token_index=int(element["s2:index"]),
```
zkirby: I believe we discussed this, but I think not every element is guaranteed to have the s2 annotations correct? Might be worth doing a check or something here.

zkirby: Ah never mind I see the _has_s2_token_annotations hahaha

data-processing/common/parse_equation.py
andrewhead: Good catch! I must have run out of steam on that comment XD Will fix it up.

data-processing/common/parse_equation.py
andrewhead: I agree, a method feels like the right call here! I'll refactor this.

data-processing/common/parse_equation.py
andrewhead: Currently because it's used only in one place, though do you think readability would be improved with a defined constant?

data-processing/common/parse_equation.py
andrewhead: Great catch this is certaintly correct.

data-processing/common/parse_equation.py
zkirby: Ah that makes sense,  is there a connection between this ["mi", "mn"] and the one in `_is_mergeable_type` or is it pure coincidence? If not, it might make sense to define then

data-processing/common/parse_equation.py
andrewhead: Ah actually, I realized why it's different. While lines 93 and 101 have the same code, `element` is different on line 101 than on line 93 (specifically, it's the child of the former element).

data-processing/common/parse_equation.py
andrewhead: Good point, they're slightly semantically different, and I think in the future they might diverge. The first list is the set of all tokens; the second is the list of all tokens that should be considered as mergeable with each other (which, in the future, may not contain all tokens, e.g., operators). I gave them names! 👍

PR #91, opened 2020-04-01
pushing terms folder, contains extractor.py, init.py and the tests fo…

jusheen:
Terms entity folder along with test file


data-processing/entities/terms_v1/__init__.py
```diff
@@ -0,0 +1,33 @@
+# This is omitting imports
```
andrewhead, line 1: What is meant by "this is omitting imports"?

andrewhead, line 1: Can the directory be renamed from `terms_v1` to `terms`?

data-processing/entities/terms_v1/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+from .extractor import Sentence, SentenceExtractor
+from .upload import Sentence as SentenceModel
+from .upload import upload_sentences
+
+commands = create_entity_localization_command_sequence(
+    "terms",
+    TermExtractor
+    # Don't need these two arguments
+    # DetectedEntityType=Sentence,
+    # upload_func=upload_sentences,
```
andrewhead, line 20: This code can be deleted

data-processing/entities/terms_v1/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+)
+
+# Don't need this
+# Register additional directories to be used by the upload function
+# directories.register("sentences-model-ids")
```
andrewhead, line 25: This code can be deleted as well

data-processing/entities/terms_v1/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+terms_pipeline = EntityPipeline(
+    "terms", commands
+    # We'll get to database uploads later; you don't need this for now
+    # , database_models=[SentenceModel]
```
andrewhead, line 30: This code can be deleted too

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+import pysbd
+
+from common.parse_tex import DEFAULT_CONTEXT_SIZE, EntityExtractor, PlaintextExtractor
+from common.types import SerializableEntity
+
+
+###First convert the aggregate glossary json into a csv, this was done on a seperate file
+
+###Extracting all the terms from glossary csv
+glossary = pd.read_csv("gloss.csv")
```
andrewhead, line 17: Can this be put in the `resources/` directory? https://github.com/allenai/scholar-reader/tree/master/data-processing/resources. That's where external data files are kept.

andrewhead, line 17: To minimize dependencies, can this be read in with the Python builtin library `csv.reader` instead of `pandas`? Check out this example: https://docs.python.org/3/library/csv.html#csv.reader

andrewhead: Also, wasn't the list of terms in JSON? Can you remind me why we decided to make it into CSV instead?

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+terms = glossary["terms"].tolist()
+
+###Converting terms to lower case
+terms = [item.lower() for item in terms]
+
+
+# Edit location #1: Define a Term entity; the below code is probably okay.
```
andrewhead, line 24: These edit tip comments can be removed

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+from common.types import SerializableEntity
+
+
+###First convert the aggregate glossary json into a csv, this was done on a seperate file
+
+###Extracting all the terms from glossary csv
+glossary = pd.read_csv("gloss.csv")
+terms = glossary["terms"].tolist()
+
+###Converting terms to lower case
```
andrewhead, line 20: Style nit pick for consistency: for comments, do a single `#` followed by a space, then the comment.

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+    text: str
+
+
+# Edit location #2: Create a term extractor;
+# It should include your test code from last week; it just has an output
+# in a slightly different shape.
+class TermExtractor(EntityExtractor):
+    """
+    <descriptive comment>
+    """
```
andrewhead, line 36: TODO(jusheen)

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+terms = glossary["terms"].tolist()
+
+###Converting terms to lower case
+terms = [item.lower() for item in terms]
+
+
+# Edit location #1: Define a Term entity; the below code is probably okay.
+@dataclass(frozen=True)
+class Term(SerializableEntity):
+    text: str
```
andrewhead, line 27: Might a more representative name for this field be `name`?

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+class TermExtractor(EntityExtractor):
+    """
+    <descriptive comment>
+    """
+
+    def parse(self, tex_path: str, tex: str) -> Iterator[Term]:
+        # Extract plaintext segments from TeX
+        plaintext_extractor = PlaintextExtractor()
+        plaintext_segments = plaintext_extractor.parse(tex_path, tex)
+        ###Creating 3 columns for the dataframe, this dataframe will later be turned into final json
```
andrewhead, line 42: See note above on comments.
Also, when a comment splits two chunks of code that are doing something different, separate them with an extra newline. For example:

```python
        plaintext_extractor = PlaintextExtractor()
        plaintext_segments = plaintext_extractor.parse(tex_path, tex)

        ###Creating 3 columns for the dataframe, this dataframe will later be turned into final json
        matched_terms_start = []
```

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+        matched_terms_end = []
+        matched_terms = []
+        matched_terms_context_str = []
+
+        for text_segment in plaintext_segments:
+            text_segment_string = str(text_segment.text.lower())
+
+            ###Making two lists from the text_segment_string; one is to capture words without dashes
+			###The other is to capture words such as "Gradient-Activated" (but this word happens to not be
+			###in the glossary
```
andrewhead, line 53: Indentation between successive lines in a multi-line comment should be consistent

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+
+			text_segment_string_split_on_dashes = re.split("[-\s+]", text_segment_string)
+            text_segment_string_not_split_on_dashes = re.split("[\s+]", text_segment_string)
```
andrewhead, line 56: Are these two lines meant to not have the same indentation?

andrewhead:
By the way, this is looking good so far! Good work commenting some of the rationale and higher-level purpose of the code. Good work as well on submitting tests with the code ⭐️

data-processing/entities/terms_v1/test_terms.py
```diff
@@ -0,0 +1,26 @@
...
+    terms = list(
+        extractor.parse(
+            "main.tex",
+            "/title (towards interpretable semantic segmentation via gradient-weighted class activation mapping)",
+        )
+    )
+    assert len(terms) == 3
+    term1 = terms[0]
+    assert term1.term == "segmentation"
+    assert term1.start == 1666
```
andrewhead, line 12: Shouldn't `term1.start` be 40?

data-processing/entities/terms_v1/test_terms.py
```diff
@@ -0,0 +1,26 @@
...
+    assert term1.end == 1677
+    assert term1.context_str == "/title (towards interpretable semantic segmentation via gradient-weighted class activation mapping)"
```
andrewhead, line 14: This should be `context_tex`

andrewhead: Also, there need not be a test for `context_str`, as it's for debugging purposes rather than serving an important functional role within the pipeline

data-processing/entities/terms_v1/extractor.py
```diff
@@ -0,0 +1,123 @@
...
+		    )
+            index += 1
+
+
+
+
+
+
+
+
```
andrewhead, line 123: nit pick: remove unnecessary whitespace

andrewhead: @jusheen are you using Visual Studio Code? If not, go ahead and download it, enable formatting on saving, and then save the code. That should take care of these dangling newlines, and fix up indents automatically 🎆

data-processing/entities/terms_v1/test_terms.py
```diff
@@ -0,0 +1,26 @@
+def test_extract_sentences():
```
andrewhead, line 1: There should also be a test for multi-word terms.

data-processing/entities/terms_v1/__init__.py
jusheen: It was just to let me know I had to add the import statements when you handed off the starter code to me

data-processing/entities/terms_v1/extractor.py
jusheen: They are meant to have the same indentation, strange. These two lines have the same indentation on my editor.

andrewhead: Hm, that's interesting 🤔 ! Ah, from selecting text in the two lines, it looks like Line 55 uses tabs and Line 56 uses spaces. Do you have a setting in your editor that can turn tabs into spaces?

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+						matched_terms_start.pop(index)
+						matched_terms_end.pop(index)
+						matched_terms_context_str.pop(index)
+						matched_definition_texts.pop(index)
+						matched_definition_htmls.pop(index)
+
+		delete_duplicate_start(matched_terms, matched_terms_start, matched_terms_end, matched_terms_context_str)
+
+		index = 0
+		for term_object_number in range(len(matched_terms)):
```
andrewhead, line 160: Nit pick: cleaner syntax for this pattern:

```suggestion
		for i, term in enumerate(matched_terms):
```

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+			yield Term(
+				name=matched_terms[term_object_number],
+				defintion_text=matched_definition_texts[term_object_number],
+				definition_html=matched_definition_htmls[term_object_number],
+				# start position of the term in the TeX
+				start=matched_terms_start[term_object_number],
+				end=matched_terms_end[term_object_number],  # end position
```
andrewhead, line 167: nit pick: Axe these comments, unless they convey something new that isn't inferable from the code or lending additional documentation beyond what's already there for the SerializableEntity class.

```suggestion
				end=matched_terms_end[term_object_number],
```

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+
+		def delete_duplicate_start(term, start, end, context_str):
+			for start_index in start:
+
+				# 1)
+				duplicate_indices = duplicates(start, start_index)
+
+				# 2) Keep the object with the largest end index.
+				if len(duplicate_indices) > 1:
+					print("OH NO")
```
andrewhead, line 136: If you want to keep this message, can this use `logging.warning` instead (see these other examples in the code https://github.com/allenai/scholar-reader/search?q=logging.warning&unscoped_q=logging.warning), and change the message to clearly indicate what's going wrong?

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+					# remove the objects that are located at the remaining indices from all lists.
+					for index in duplicate_indices:
+						matched_terms.pop(index)
+						matched_terms_start.pop(index)
+						matched_terms_end.pop(index)
+						matched_terms_context_str.pop(index)
+						matched_definition_texts.pop(index)
+						matched_definition_htmls.pop(index)
+
+		delete_duplicate_start(matched_terms, matched_terms_start, matched_terms_end, matched_terms_context_str)
```
andrewhead, line 157: why would there be duplicate terms detected? This suggests to me that it might be worth refactoring the approach above to not have duplicates, to yield a simpler and more maintainable solution with fewer steps

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+			for object in terms_and_def_text_html:
+				term_list = re.split("\s+", object[0])
+				if len(term_list) > 1:
+					if object[0] in text_segment_string:
+						matches_tuples.append([object[0], text_segment_string, object[1], object[2]])
+
+			for i in matches_tuples:
+				list_to_add = i[:-1]
+				if isinstance(i[-1], list):
+					last_elem = tuple(i[-1])
```
andrewhead, line 102: Why does an element in the match tuples need to be cast from a list to a tuple here?

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+		new_triple.append(obj["definition_text"])
+		new_triple.append(obj["definition_html"])
+		terms_and_def_text_html.append(new_triple)
+
+	matched_terms = []
+	matched_definition_texts = []
+	matched_definition_htmls = []
+	matched_terms_start = []
+	matched_terms_end = []
+	matched_terms_context_str = []
```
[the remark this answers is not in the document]
jusheen: They are inside the parse method now; i realized I need to make a new .__init__ function because we have to add a new parameter for the TermExtractor (namely, a parameter for the input glossary).

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+	defintion_text: str
+
+	#list if there is more than one string for the definition html
+	definition_html: str or list
 
 
 # Edit location #2: Create a term extractor;
 # It should include your test code from last week; it just has an output
 # in a slightly different shape.
 class TermExtractor(EntityExtractor, glossary):
```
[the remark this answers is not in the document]
jusheen: Revised, that was a mistake. Made a new init function and added a glossary parameter

data-processing/entities/terms/extractor.py
jusheen: There technically aren't any "duplicate" terms; it's an issue with the glossary. The glossary has terms that are really similar: i.e., "convolutional neural networks", "convolutional neural network". The only difference is that one is plural and the other is singular (this is due to the glossary being compiled from different websites). I decided that it would be redundant to capture both of these (albeit they would have different definitions because they are listed in the glossary as different terms).

jusheen: How do you want me to handle this situation? I can capture both the singular and plural form, or just capture one of them. Currently, I am capturing the singular form

data-processing/entities/terms/extractor.py
jusheen: I wanted to add it to the matches set. Python won't allow lists to be added to set objects, but it allows tuples to be added.

data-processing/entities/terms/extractor.py
jusheen: Removed this debugging print statement

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+	# Extracting all the terms, definition_text and definition_html from glossary
+	terms_and_def_text_html = []
+
+	for obj in glossary:
+		new_triple = []
+
+		#Making terms lower case in one step
+		new_triple.append(obj["terms"].lower())
+		new_triple.append(obj["definition_text"])
+		new_triple.append(obj["definition_html"])
```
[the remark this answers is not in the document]
jusheen: Done, made a new class called GlossaryTerm

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+			for word in text_segment_string_split_on_dashes:
+				for object in terms_and_def_text_html:
+
+					# Object[0] is the term
+					if object[0] == word:
+
+						# Adding term name, text_segment from which term was found, def text and def html
+						matches_tuples.append([object[0], text_segment_string, object[1], object[2]])
+
+			for word2 in text_segment_string_not_split_on_dashes:
```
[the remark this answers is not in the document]
jusheen: Good point, word is the same as word2. Made the change.

data-processing/entities/terms/extractor.py
andrewhead: Good catch! Can we just capture the singular form for each? Would there be any downsides to doing that?

jusheen: The code as is does that already; I'll leave it as is for now then. We might have to come back to this later, I thought of some different approaches that might be cleaner for the future (for example, checking if a word is in the glossary by checking once if the word (in it original form from the tex) is in the glossary, and then checking one more time if the word (after adding an "s" character to the end, making it the plural form) is in the glossary.

data-processing/entities/terms/extractor.py
```diff
@@ -10,110 +9,169 @@
...
+		global matched_terms_context_str
+
+		for text_segment in plaintext_segments:
+			text_segment_string = str(text_segment.text.lower())
+
+			# Making two lists from the text_segment_string; one is to capture words without dashes
+			# The other is to capture words such as "Gradient-Activated" (but this word happens to not be
+			# in the glossary
+			text_segment_string_split_on_dashes = re.split(
+				"[-\s+]", text_segment_string)
```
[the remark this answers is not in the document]
jusheen: This was a mistake. The "+" sign should be outside the brackets. Thanks for the catch!

data-processing/entities/terms/extractor.py
jusheen: I thought of some different approaches that might be cleaner (for example, checking if a word is in the glossary by checking once if the word (in it original form from the tex) is in the glossary, and then checking one more time if the word (after adding an "s" character to the end, making it the plural form) is in the glossary. This is something I am gonna come back to

andrewhead:
Closing for now, as I think Raymond is following up on this PR with some improvements.

PR #92, opened 2020-04-06
PDF Pipeline

rodneykinney:
Populate the scholar-reader database with citations and symbols from the Grobid-based PDF processing pipeline. Calls the Grobid API as a service and converts the Grobid annotations and token bounding boxes to reader entities.

Depends on [these changes](https://github.com/allenai/grobid/pull/10) to Grobid.

Had to do some mild refactoring of the existing pipeline code to enable this alternate entry point. Mostly this involves moving existing methods into different modules, like some data classes and the `upload` methods, which have been put into the `util` modules.

### DIFF
--- a/data-processing/common/models.py
+++ b/data-processing/common/models.py
@@ -152,7 +152,8 @@ def init_database_connections(
     Initialize database connections.
     """
 
-    config = configparser.ConfigParser()
+    import os
+    config = configparser.ConfigParser(os.environ)
     config.read(DATABASE_CONFIG)
 
     # By default, data will be placed in a schema with the timestamp of the time that this
--- a/data-processing/common/types.py
+++ b/data-processing/common/types.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, List, NamedTuple, Optional
+from typing import Any, Callable, Dict, Set, List, NamedTuple, Optional
 
 """
 FILE PROCESSING
@@ -484,6 +484,14 @@ CharacterLocations = Dict[CharacterId, List[BoundingBox]]
 class CitationLocation(BoundingBox):
     key: str
     cluster_index: int
+
+@dataclass(frozen=True)
+class CitationData:
+    arxiv_id: ArxivId
+    s2_id: S2Id
+    citation_locations: Dict[str, Dict[int, Set[CitationLocation]]]
+    key_s2_ids: Dict[str, str]
+    s2_data: Dict[S2Id, SerializableReference]
 
 
 @dataclass(frozen=True)
--- a/data-processing/config.ini
+++ b/data-processing/config.ini
@@ -49,7 +49,7 @@ port = <input-port>
 [output-db]
 db_name = scholar-reader
 user = data-pipeline
-password = <substitute-with-password>
+password = %(READER_DB_PASSWORD)s
 host = scholar-reader.c5tvjmptvzlz.us-west-2.rds.amazonaws.com
 port = 5432
 
--- a/data-processing/entities/citations/commands/resolve_bibitems.py
+++ b/data-processing/entities/citations/commands/resolve_bibitems.py
@@ -7,6 +7,7 @@ from typing import Iterator, List, Set
 from common import directories, file_utils
 from common.commands.base import ArxivBatchCommand
 from common.types import ArxivId, Bibitem, BibitemMatch, SerializableReference
+from ..utils import extract_ngrams, ngram_sim
 
 
 @dataclass(frozen=True)
@@ -21,23 +22,6 @@ The title and authors must have at least this much overlap with the bibitem's te
 the similarity metric below to be considered a match.
 """
 SIMILARITY_THRESHOLD = 0.5
-
-
-def extract_ngrams(s: str, n: int = 3) -> Set[str]:
-    """
-    This and the 'ngram_sim' method below were provided by Kyle Lo.
-    """
-    s = re.sub(r"\W", "", s.lower())
-    ngrams = zip(*[s[i:] for i in range(n)])
-    return {"".join(ngram) for ngram in ngrams}
-
-
-def ngram_sim(s1: str, s2: str) -> float:
-    s1_grams = extract_ngrams(s1)
-    s2_grams = extract_ngrams(s2)
-    if len(s1_grams) == 0 or len(s2_grams) == 0:
-        return 0
-    return len(s1_grams.intersection(s2_grams)) / min(len(s1_grams), len(s2_grams))
 
 
 class ResolveBibitems(ArxivBatchCommand[MatchTask, BibitemMatch]):
--- a/data-processing/entities/citations/commands/upload_citations.py
+++ b/data-processing/entities/citations/commands/upload_citations.py
@@ -12,24 +12,14 @@ from common.commands.database import DatabaseUploadCommand
 from common.models import BoundingBox as BoundingBoxModel
 from common.models import (Citation, CitationPaper, Entity, EntityBoundingBox,
                            Paper, Summary, output_database)
-from common.types import (ArxivId, BibitemMatch, CitationLocation,
+from common.types import (ArxivId, BibitemMatch, CitationLocation, CitationData,
                           SerializableReference)
 
-from ..utils import load_located_citations
+from ..utils import load_located_citations, upload_citations
 
 CitationKey = str
 LocationIndex = int
 S2Id = str
-
-
-@dataclass(frozen=True)
-class CitationData:
-    arxiv_id: ArxivId
-    s2_id: S2Id
-    citation_locations: Dict[CitationKey, Dict[LocationIndex, Set[CitationLocation]]]
-    key_s2_ids: Dict[CitationKey, S2Id]
-    s2_data: Dict[S2Id, SerializableReference]
-
 
 class UploadCitations(DatabaseUploadCommand[CitationData, None]):
     @staticmethod
@@ -105,71 +95,4 @@ class UploadCitations(DatabaseUploadCommand[CitationData, None]):
         yield None
 
     def save(self, item: CitationData, _: None) -> None:
-
-        arxiv_id = item.arxiv_id
-        s2_id = item.s2_id
-        citation_locations = item.citation_locations
-        key_s2_ids = item.key_s2_ids
-        s2_data = item.s2_data
-
-        try:
-            paper = Paper.get(Paper.s2_id == s2_id)
-        except Paper.DoesNotExist:
-            paper = Paper.create(s2_id=s2_id, arxiv_id=arxiv_id)
-
-        for citation_key, locations in citation_locations.items():
-
-            if citation_key in key_s2_ids:
-                s2_id = key_s2_ids[citation_key]
-                if s2_id in s2_data:
-                    paper_data = s2_data[s2_id]
-
-                    try:
-                        cited_paper = Paper.get(Paper.s2_id == s2_id)
-                    except Paper.DoesNotExist:
-                        cited_paper = Paper.create(
-                            s2_id=s2_id, arxiv_id=paper_data.arxivId or None
-                        )
-
-                    try:
-                        Summary.get(Summary.paper == cited_paper)
-                    except Summary.DoesNotExist:
-                        Summary.create(
-                            paper=cited_paper,
-                            title=paper_data.title,
-                            authors=paper_data.authors,
-                            doi=paper_data.doi or None,
-                            venue=paper_data.venue or None,
-                            year=paper_data.year,
-                        )
-
-            for location_set in locations.values():
-                citation = Citation.create(paper=paper)
-                try:
-                    with output_database.atomic():
-                        CitationPaper.create(paper=cited_paper, citation=citation)
-                except IntegrityError:
-                    logging.warning(  # pylint: disable=logging-not-lazy
-                        (
-                            "Cited paper %s and citation %s are already linked. This suggests a bug in "
-                            + "the citation resolution code; perhaps multiple citation keys were "
-                            + "matched to the same paper?"
-                        ),
-                        cited_paper.arxiv_id,
-                        citation.id,
-                    )
-
-                entity = Entity.create(
-                    type="citation", source="tex-pipeline", entity_id=citation.id
-                )
-
-                for location in location_set:
-                    bounding_box = BoundingBoxModel.create(
-                        page=location.page,
-                        left=location.left,
-                        top=location.top,
-                        width=location.width,
-                        height=location.height,
-                    )
-
-                    EntityBoundingBox.create(bounding_box=bounding_box, entity=entity)
+        upload_citations(item)
--- a/data-processing/entities/symbols/commands/upload_symbols.py
+++ b/data-processing/entities/symbols/commands/upload_symbols.py
@@ -1,59 +1,26 @@
 import logging
 import os.path
-from dataclasses import dataclass
-from typing import Dict, Iterator, List
+from typing import Dict, Iterator
 
 from common import directories, file_utils
 from common.commands.database import DatabaseUploadCommand
-from common.models import BoundingBox as BoundingBoxModel
-from common.models import Entity, EntityBoundingBox, output_database
 from common.s2_data import get_s2_id
 from common.types import (
-    ArxivId,
     BoundingBox,
     Match,
     Matches,
-    MathML,
     SymbolId,
     SymbolLocation,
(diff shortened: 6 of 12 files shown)


data-processing/common/types.py
```diff
@@ -485,6 +485,14 @@ class CitationLocation(BoundingBox):
     key: str
     cluster_index: int
 
+@dataclass(frozen=True)
+class CitationData:
```
rodneykinney, line 489: Move this class.

data-processing/common/models.py
```diff
@@ -152,7 +152,9 @@ def init_database_connections(
     Initialize database connections.
     """
 
-    config = configparser.ConfigParser()
+    import os
+    #config = configparser.ConfigParser()
+    config = configparser.ConfigParser(os.environ)
```
rodneykinney, line 156: To be able to specify the database password in an env variable.

data-processing/config.ini
```diff
@@ -38,18 +38,23 @@ binary = /root/perl5/perlbrew/perls/perl-5.22.4/bin/perl
...
 [input-db]
-db_name = <input-db-name>
-user = <input-user-name>
-password = <substitute-with-password>
-host = <input-host>
-port = <input-port>
+db_name = scholar-reader
+user = data-pipeline
+password = %(READER_DB_PASSWORD)s
+host = scholar-reader.c5tvjmptvzlz.us-west-2.rds.amazonaws.com
```
rodneykinney, line 44: Is this all right? Do you prefer to just edit this file locally?

data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -7,6 +7,7 @@
 from common import directories, file_utils
 from common.commands.base import ArxivBatchCommand
 from common.types import ArxivId, Bibitem, BibitemMatch, SerializableReference
+from ..utils import extract_ngrams, ngram_sim
```
rodneykinney, line 10: Put these into `utils`

data-processing/entities/citations/commands/upload_citations.py
```diff
@@ -105,71 +95,4 @@ def process(self, _: CitationData) -> Iterator[None]:
...
-                    bounding_box = BoundingBoxModel.create(
-                        page=location.page,
-                        left=location.left,
-                        top=location.top,
-                        width=location.width,
-                        height=location.height,
-                    )
-
-                    EntityBoundingBox.create(bounding_box=bounding_box, entity=entity)
+        upload_citations(item)
```
rodneykinney, line 98: Factor out this method into `utils`

data-processing/entities/symbols/commands/upload_symbols.py
```diff
@@ -1,60 +1,27 @@
...
 )
 from entities.sentences.commands.find_entity_sentences import EntitySentencePairIds
 from entities.sentences.upload import SentenceIdAndModelId
 
-from ..models import MathMl as MathMlModel
-from ..models import MathMlMatch, Paper
-from ..models import Symbol as SymbolModel
-from ..models import SymbolChild, SymbolSentence
+from ..types import SymbolData, SymbolKey, SentenceKey
+from ..utils import upload_symbols
```
rodneykinney, line 19: Move these.

data-processing/entities/symbols/utils.py
```diff
@@ -0,0 +1,132 @@
...
+from .types import SymbolData, SymbolKey, SentenceKey
+from .models import MathMl as MathMlModel
+from .models import MathMlMatch, SymbolChild, SymbolSentence
+from .models import Symbol as SymbolModel
+from common.types import MathML, SymbolId
+from common.models import BoundingBox as BoundingBoxModel
+from common.models import output_database, Paper, Entity, EntityBoundingBox
+
+
+def upload_symbols(item: SymbolData, source: str = "tex-pipeline") -> None:
```
rodneykinney, line 12: Moved but otherwise unchanged.

data-processing/pdf/grobid_client.py
```diff
@@ -0,0 +1,24 @@
...
+    try:
+        files = {'input':(pdf_file, open(pdf_file,'rb'), 'application/pdf',{'Expires':'0'})}
+        resp = requests.post('http://localhost:8070/api/processPdfStructure',files=files)
+
+        if resp.status_code == 200:
+            return json.loads(resp.text)
+        else:
+            raise Exception("Grobid returned status code {}".format(resp.status_code))
+    except requests.exceptions.ConnectionError:
+        msg="""Could not connect to local grobid service.
```
rodneykinney, line 19: Running the PDF pipeline only requires pulling and running this Grobid Docker container.

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+        citations = self.get_citations()
+        upload_citations(citations, 'pdf-pipeline')
+        symbols = self.get_symbols()
+        upload_symbols(symbols,'pdf-pipeline')
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.INFO)
+
+    pdf_hashes = ['8c53cd64e46cf382bf81ce2c4e0087ef31351919']
+    if sys.argv[1:]:
```
rodneykinney, line 214: To run the pipeline, point this script to a file of PDF hashes.

data-processing/pdf/grobid_client.py
andrewhead: Sweet! Loving the embedded docs 👍

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+
+import pdf.grobid_client
+from common.models import BoundingBox as BoundingBoxModel
+from common.models import init_database_connections
+from common.types import CitationData, CitationLocation, SerializableReference, BoundingBox, Matches, Match, Symbol
+from entities.symbols.types import SymbolData, SymbolWithId, SymbolId
+from entities.citations.utils import upload_citations, extract_ngrams, ngram_sim
+from entities.symbols.utils import upload_symbols
+
+class PdfStructureParser(object):
```
andrewhead, line 17: nit pick: afaik, `(object)` object base shouldn't be necessary for a Python 3 runtime?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+            if direct:
+                text += direct
+            else:
+                for i in range(s['left'], s['right']):
+                    page, token = self.find_token(i)
+                    text.append(token['text'])
+        return ' '.join(text)
+
+    def union(self, bbox1: BoundingBoxModel, bbox2: BoundingBoxModel):
+        if not bbox1:
```
andrewhead, line 49: Does this need an analogous check to see if `bbox2` is None?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+            return bbox2
+        x1 = min(bbox1.left,bbox2.left)
```
andrewhead, line 51: Just double-checking as I know that has bitten in the butt more than once with PDF: `top` = the lower number (i.e. measured from the top of the document) and `bottom` = the higher number?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+        y1 = min(bbox1.top,bbox2.top)
+        x2 = max(bbox1.left+bbox1.width, bbox2.left+bbox2.width)
+        y2 = max(bbox1.top+bbox1.height, bbox2.top+bbox2.height)
+        return BoundingBoxModel(page = bbox1.page,
```
andrewhead, line 55: Are pages indexed starting at 0?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+                                left = x1, top = y1, width = x2-x1, height=y2-y1)
+
+    @staticmethod
+    def should_combine(bbox1: BoundingBoxModel, bbox2: BoundingBoxModel):
+        return bbox1.page == bbox2.page and (
+            abs(bbox1.top - bbox2.top) < 4 # Same y-coordinate
+        ) and (
+            abs(bbox2.left - bbox1.left - bbox1.width) < 15 # To the right
```
andrewhead, line 63: Should this be comparing the box2 left to box1 right? Doesn't this equation mean that this would be considered overlapping, because bbox1 right would be less than 15 pixels away from bbox2 left?

```
------      ------
|box1|<-14->|box2|
------      ------
```

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+                    height=token['height']
+                )
+                if not bbox:
+                    bbox = token_bbox
+                elif self.should_combine(bbox, token_bbox):
+                     bbox = self.union(bbox, token_bbox)
+                else:
+                    bboxes.append(bbox)
+                    bbox = token_bbox
+            bboxes.append(bbox)
```
andrewhead, line 87: What's this last append for? Will it repeat the work in Line 85?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+        return bboxes
+
+
+    def find_cited_paper(self, bib_item_title):
+        if self.references is None:
+            resp = requests.get(f"https://api.semanticscholar.org/v1/paper/{self.pdf_hash}")
```
andrewhead, line 93: Neat!

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+                self.references = []
+
+        max_similarity = 0.0
+        best_matching_paper = None
+        for reference_data in self.references:
+            reference_title = reference_data['title']
+            similarity = ngram_sim(reference_title, bib_item_title)
+            if similarity > 0.5 and similarity > max_similarity:
+                max_similarity = similarity
+                best_matching_paper = reference_data
```
andrewhead, line 107: This paper looks pretty similar to what we have in `resolve_bibitems.py`. Do you think it makes sense to pull it out into a new method?

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+        symbols_with_ids: List[SymbolWithId] = []
+        boxes: Dict[SymbolId, BoundingBox] = {}
+        matches: Matches = {}
+        symbol_index = 0
+        for sym in self.elements['<symbol>']:
+            id = sym['tags'].get('id')
+            if id:
+                spans = sym['spans']
+                bboxes = self.get_bounding_boxes(spans)
+                mock_math_ml = '<pdf_symbol>{}</pdf_symbol>'.format(id)
```
andrewhead, line 121: 👍 this all looks fine

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+    def upload(self):
+        citations = self.get_citations()
+        upload_citations(citations, 'pdf-pipeline')
+        symbols = self.get_symbols()
+        upload_symbols(symbols,'pdf-pipeline')
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.INFO)
+
+    pdf_hashes = ['8c53cd64e46cf382bf81ce2c4e0087ef31351919']
```
andrewhead, line 213: Hahaha I see you were testing this out on one of the world's shortest papers as well 👍

data-processing/pdf/process_pdf.py
```diff
@@ -0,0 +1,236 @@
...
+    init_database_connections('public')
+    with TemporaryDirectory() as tempdir:
+        for pdf_hash in pdf_hashes:
+            start_time = time.time()
+            logging.info("Processing PDF %s", pdf_hash)
+            try:
+                pdf_file = "{}/{}.pdf".format(tempdir, pdf_hash)
+                try:
+                    subprocess.check_call([
+                        "aws", "s3", "cp", "s3://ai2-s2-pdfs/{}/{}.pdf".format(pdf_hash[:4], pdf_hash[4:]), pdf_file
```
andrewhead, line 226: Does this mean this script should be run only within the AI2-internal AWS account?

data-processing/pdf/process_pdf.py
rodneykinney: I admit I'm a little confused here. When I looked at some examples of bounding boxes from the tex pipeline, they seemed to be starting from 1, but I see many bounding boxes with `page=0` in the database. What's the right convention?

data-processing/pdf/process_pdf.py
rodneykinney: There's no `.right` field on `BoundingBox`, so it's comparing to `bbox1.left+bbox1.width`.

data-processing/pdf/process_pdf.py
rodneykinney: No, because `bbox` is re-assigned after line 85. This line at the end of the loop is needed to add the last bounding box. I should add a check for `None`, though, to handle the empty case.

data-processing/pdf/process_pdf.py
rodneykinney: I think that's probably not worth the effort, given that the logic is simple, and the two loops are iterating over different types of objects.

data-processing/pdf/process_pdf.py
rodneykinney: 😁

data-processing/pdf/process_pdf.py
rodneykinney: Well, if you run it using another account, the first S3 call will succeed, since that's a public bucket, but the fallback to `ai2-s2-pdfs-private` will fail.

data-processing/pdf/grobid_client.py
```diff
@@ -0,0 +1,24 @@
...
+import json
+
+def get_pdf_structure(pdf_file):
+    """
+    Get parsed PDF data from a local instance of the Grobid service
+    :param pdf_file:
+    :return:
+    """
+    try:
+        files = {'input':(pdf_file, open(pdf_file,'rb'), 'application/pdf',{'Expires':'0'})}
```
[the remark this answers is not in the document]
andrewhead: Roger, that's a good idea. Let's skip formatting for this PR.

data-processing/pdf/process_pdf.py
andrewhead: Got it! I misunderstood the intent of this code initially (I thought it was trying to merge overlapping bounding boxes, not adjacent ones). This looks good.

data-processing/pdf/process_pdf.py
andrewhead: After refreshing my memory on some of the papers I've run recently, it looks like the right convention is to start at `page=0`. Out of curiosity, which examples are you looking at where they start from 1? I'd like to check on that to make sure the pipeline's being consistent.

rodneykinney: I used this query to find TeX-pipeline bounding boxes and compare them with the PDF pipeline's:

```
select
       e.source,
       cp.paper_id,
       s.title,
       array_agg(bb.page || '-[' || bb."left" || ',' || bb.top || ',' || bb.width || ',' || bb.height || ']' order by bb.page) as bounding_boxes
from citation c
inner join entity e on e.entity_id = c.id
inner join entityboundingbox ebb on e.id = ebb.entity_id
inner join boundingbox bb on bb.id = ebb.bounding_box_id
inner join citationpaper cp on c.id = cp.citation_id
left outer join summary s on s.paper_id = cp.paper_id
where e.type = 'citation'
and c.paper_id = '3febb2bed8865945e7fddc99efd791887bb7e14f'
group by cp.paper_id, e.source, s.title
order by cp.paper_id, e.source
```

Looking at this again, I think what I was interpreting as 1-index pages were probably false positives on the wrong page.

BTW, I did notice that the schema doesn't have a way to distinguish between entities inserted in different runs, so it's pretty easy to create duplicates.

andrewhead: > BTW, I did notice that the schema doesn't have a way to distinguish between entities inserted in different runs, so it's pretty easy to create duplicates.

This is a good point.  The way I've been handling this so far has been to insert data from different runs into different schemas in the same Postgres database. In your scripts, this would mean calling [`init_database_connections`](https://github.com/allenai/scholar-reader/blob/master/data-processing/common/models.py#L148) with a different schema name when you want to insert data into a fresh copy of the tables.

That said, maybe this is a hassle (it has certainly required a specific workflow for me), so let me know if changing the tables to include run version information would be a big value add and it's something I'll add to the docket.

PR #103, opened 2020-04-20
Add symbol search

zkirby:
Semi-working implementation of symbol search.
**TODO:**
- Fix typescript errors
- Remove hack to clear search query on close

**Search for a symbol:**
ezgif com-video-to-gif (2)
**Interrupt a symbol search with text and vice versa:**
ezgif com-video-to-gif (1)

### DIFF
--- a/ui/src/FindBar.tsx
+++ b/ui/src/FindBar.tsx
@@ -129,7 +129,7 @@ export class FindBar extends React.PureComponent<FindBarProps, FindBarState> {
       const newEntityId = order[
         this.wrapIndex(order.indexOf(this.context.selectedEntityId || '') + 1, order.length)
       ];
-      jumpToBoundingBox(matches.get(newEntityId));
+      // jumpToBoundingBox(matches.get(newEntityId));
       this.context.setSelectedEntity(newEntityId, "symbol");
     } else {
       this.dispatchEventToPdfJs("findagain");
@@ -144,7 +144,7 @@ export class FindBar extends React.PureComponent<FindBarProps, FindBarState> {
       const newEntityId = order[
         this.wrapIndex(order.indexOf(this.context.selectedEntityId || '') - 1, order.length)
       ];
-      jumpToBoundingBox(matches.get(newEntityId));
+      // jumpToBoundingBox(matches.get(newEntityId));
       this.context.setSelectedEntity(newEntityId, "symbol");
     } else {
       this.dispatchEventToPdfJs("findagain", true);
--- a/ui/src/FindBar.tsx
+++ b/ui/src/FindBar.tsx
@@ -1,9 +1,16 @@
 import React from "react";
 import { EventBus } from "./types/pdfjs-viewer";
+import { ScholarReaderContext } from "./state";
+
+interface FindBarProps {
+  matches: Map<String, Object>;
+  jumpToBoundingBox: Function;
+}
 
 interface FindBarState {
   currentMatch: number | null;
   matchCount: number | null;
+  mode: string,
   /**
    * Event bus for the pdf.js application. The logic for performing text search within the PDF is
    * provided by pdf.js. This event bus is needed for communicating with the pdf.js code.
@@ -11,13 +18,32 @@ interface FindBarState {
   pdfJsEventBus?: EventBus;
 }
 
-export class FindBar extends React.PureComponent<{}, FindBarState> {
-  constructor(props: {}) {
-    super(props);
-    this.state = {
-      currentMatch: null,
-      matchCount: null,
-    };
+export class FindBar extends React.PureComponent<FindBarProps, FindBarState> {
+  state = {
+    currentMatch: null,
+    matchCount: null,
+    mode: 'text-search',
+  };
+
+  static contextType = ScholarReaderContext;
+  context!: React.ContextType<typeof ScholarReaderContext>;
+
+  /**
+   * This is considered an anti pattern in React (https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html),
+   * unfortunately since we are using this component for both string search (which does update match count without props) and symbol search
+   * we have to derive the match count and current match from the props. In the future we should probably divide these
+   * into separate components.
+   */
+  componentDidUpdate() {
+    if (this.context.selectedEntityId !== null) {
+      const order = [...this.props.matches.keys()];
+      console.log(this.props.matches, order, this.context.selectedEntityId)
+      this.setState({
+        currentMatch: order.indexOf(this.context.selectedEntityId) + 1,
+        matchCount: this.props.matches.size,
+        mode: 'symbol-search',
+      })
+     } 
   }
 
   componentDidMount() {
@@ -38,10 +64,10 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
     pdfViewerApplicationAny.externalServices = {
       ...pdfViewerApplicationAny.externalServices,
       updateFindControlState: (data: any) => {
-        console.log("Updating find control state", data);
+        // console.log("Updating find control state", data);
       },
       updateFindMatchesCount: (data: any) => {
-        console.log("Update find matches count", data);
+        // console.log('updating find matches count', data)
       },
       supportsIntegratedFind: true,
     };
@@ -92,12 +118,38 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
     this.dispatchEventToPdfJs("find");
   }
 
-  handleNextButtonClick() {
-    this.dispatchEventToPdfJs("findagain");
+  wrapIndex(index: number, len: number) {
+    return (index%len + len)%len;
   }
 
-  handlePreviousButtonClick() {
-    this.dispatchEventToPdfJs("findagain", true);
+  handleNextButtonClick(e) {
+    if (this.state.mode === 'symbol-search' && this.context.symbols) {
+      const { matches, jumpToBoundingBox } = this.props;
+      const order = [...matches.keys()];
+      const newEntityId = order[
+        this.wrapIndex(order.indexOf(this.context.selectedEntityId || '') + 1, order.length)
+      ];
+      jumpToBoundingBox(matches.get(newEntityId));
+      this.context.setSelectedEntity(newEntityId, "symbol");
+    } else {
+      this.dispatchEventToPdfJs("findagain");
+    }
+    e.preventDefault();
+  }
+
+  handlePreviousButtonClick(e) {
+    if (this.state.mode === 'symbol-search' && this.context.symbols) {
+      const { matches, jumpToBoundingBox } = this.props;
+      const order = [...matches.keys()];
+      const newEntityId = order[
+        this.wrapIndex(order.indexOf(this.context.selectedEntityId || '') - 1, order.length)
+      ];
+      jumpToBoundingBox(matches.get(newEntityId));
+      this.context.setSelectedEntity(newEntityId, "symbol");
+    } else {
+      this.dispatchEventToPdfJs("findagain", true);
+    }
+    e.preventDefault();
   }
 
   dispatchEventToPdfJs(eventType: string, findPrevious?: boolean) {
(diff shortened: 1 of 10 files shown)


andrewhead:
Wow!!!

ui/src/FindBar.tsx
```diff
@@ -1,23 +1,53 @@
...
+interface FindBarProps {
+  mappingToBounds: Map<String, BoundingBox>;
+  matches: string[];
+  selectedSymbol: string;
+}
 
 interface FindBarState {
   currentMatch: number | null;
   matchCount: number | null;
+  mode: string,
```
andrewhead, line 16: Thought: this might help make the available types for the mode explicit:

```suggestion
  mode: "hidden" | "text-search" | "symbol-search",
```

ui/src/FindBar.tsx
```diff
@@ -1,23 +1,53 @@
...
-      matchCount: null,
-    };
+const initialState = {
+  currentMatch: null,
+  matchCount: null,
+  mode: 'hidden',
+}
+
+export class FindBar extends React.PureComponent<FindBarProps, FindBarState> {
+  state = initialState;
```
andrewhead, line 31: Is it a better idiom in React to set state in this way vs. in a constructor?

andrewhead: Also, can we be sure that the `initalState` object isn't changed by later calls to `setState` within the component?

ui/src/FindBar.tsx
```diff
@@ -1,23 +1,53 @@
 import React from "react";
+import { convertToAnnotationId } from './selectors/annotation'
 import { EventBus } from "./types/pdfjs-viewer";
+import { ScholarReaderContext } from "./state";
+import { BoundingBox } from "./types/api";
+
+interface FindBarProps {
+  mappingToBounds: Map<String, BoundingBox>;
+  matches: string[];
+  selectedSymbol: string;
```
andrewhead, line 10: Might it make more sense to access these three properties through `ScholarReaderContext.Consumer`? I think that's the pattern that tends to be used across the other components to get access to information about global state of what's selected.

ui/src/FindBar.tsx
```diff
@@ -1,23 +1,53 @@
...
+export class FindBar extends React.PureComponent<FindBarProps, FindBarState> {
+  state = initialState;
+  static contextType = ScholarReaderContext;
+  context!: React.ContextType<typeof ScholarReaderContext>;
+
+  /**
+   * This is considered an anti pattern in React (https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html),
+   * unfortunately since we are using this component for both string search (which does update match count without props) and symbol search
+   * we have to derive the match count and current match from the props. In the future we should probably divide these
+   * into separate components.
```
andrewhead, line 39: I'm interested in your commend "divide these into separate components". I wonder if we can have a quick call about what a better way to organize this code would be. I'm planning on cleaning up the connection to the Mozilla code, and I might be able to implement a cleaner version of this as part of those efforts.

ui/src/FindBar.tsx
```diff
@@ -92,12 +133,59 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
...
 
-  handlePreviousButtonClick() {
-    this.dispatchEventToPdfJs("findagain", true);
+  selectSymbol(id: string, boxId: string) {
+    this.context.setSelectedEntity(id, "symbol");
+    this.context.setSelectedAnnotationId(convertToAnnotationId(id));
+    this.context.setSelectedAnnotationSpanId(boxId);
+  }
+
+  unselectSymbol() {
```
andrewhead, line 146: nit pick for consistency with terminology elsewhere in the code base

```suggestion
  deselectSymbol() {
```

ui/src/FindBar.tsx
```diff
@@ -92,12 +133,59 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
...
+    this.context.setSelectedEntity(null, null);
+    this.context.setSelectedAnnotationId(null);
+    this.context.setSelectedAnnotationSpanId(null);
+  }
+
+  moveToNextSymbol(movement: number) {
+    const { matches, mappingToBounds, selectedSymbol } = this.props;
+    const newEntityId = matches[
+      this.wrapIndex(matches.indexOf(selectedSymbol || '') + movement, matches.length)
```
andrewhead, line 155: Is there a way to structure the `matches` data so that we don't need to look up the index of the selectedSymbol every time we change which symbol is focused? Also, isn't the index of the current symbol stored in the `currentMatch` state property?

ui/src/FindBar.tsx
```diff
@@ -92,12 +133,59 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
...
+    } else {
+      this.dispatchEventToPdfJs("findagain", true);
+    }
+    e.preventDefault();
+  }
+
+  closeFinder = (e) => {
+    this.setState(initialState);
+    this.unselectSymbol();
+    // TODO: fix this super hacky way to remove the green highlight on close.
```
andrewhead, line 183: Hahaha got anything in mind for a better implementation for this one?

andrewhead: If not, you can still leave the TODO in here and assign it to me (`TODO(andrewhead)`) as I might get a chance to figure out pdf.js's preferred way of terminating searches when I clean up our integration with pdf.js's search capabilities.

ui/src/FindBar.tsx
```diff
@@ -131,22 +219,26 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
       return null;
     }
 
-    if (matchCount > MATCH_COUNT_LIMIT) {
+    if ((matchCount || 0) > MATCH_COUNT_LIMIT) {
```
andrewhead, line 222: Is this fix just to satisfy type checking?

ui/src/FindBar.tsx
```diff
@@ -131,22 +219,26 @@ export class FindBar extends React.PureComponent<{}, FindBarState> {
...
       return `More than ${MATCH_COUNT_LIMIT} matches`;
     }
 
     return `${currentMatch} of ${matchCount} matches`;
   }
 
   render() {
+    if (this.state.mode === 'hidden') {
+      return <div></div>
```
andrewhead, line 231: Why not return null or undefined?

ui/src/PageOverlay.tsx
```diff
@@ -12,6 +12,7 @@ import { UserAnnotationLayer } from "./UserAnnotationLayer";
 interface PageProps {
   pageNumber: number;
   view: PDFPageView;
+  highlightedSymbols: string[];
```
andrewhead, line 15: Same goes here as for the comment on the FindBarProps---wouldn't this make more sense to store in the ScholarReader's state, and then access it here through context?

ui/src/ScholarReader.tsx
```diff
@@ -430,14 +417,18 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
     const SCROLL_OFFSET_X = -400;
     const SCROLL_OFFSET_Y = +100;
 
-    if (this.state.pdfViewer !== null) {
+    if (this.state.pdfViewer !== null && this.state.pages !== null && this.state.pages[box.page + 1] !== undefined) {
```
andrewhead, line 420: I think the checks `this.state.pages !== null && this.state.pages[box.page + 1] !== undefined` might be overly restrictive. Pages only get added to `this.state` when pdf.js tells us that the page has been rendered, which will only happen when the page scrolls into focus.

I think it will be frequent that `jumpToBoundingBox` will be called for symbol on pages that hasn't yet been rendered by pdf.js. So if we can, this condition should be removed.

ui/src/ScholarReader.tsx
```diff
@@ -430,14 +417,18 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+      const { left, top } = selectors.divDimensionStyles(
+        this.state.pages[box.page + 1].view,
+        box 
+      )
```
andrewhead, line 424: I'm not sure this is quite right. If I recall correctly (though I might be wrong, perhaps you can double check?), `scrollPageIntoView` expects the scroll coordinates in PDF coordinates. `divDimensionStyles` returns pixel coordinates, which will change based on the size of the page div (which will change with zooming, etc.).

I think this might need to be updated to convert `left` and `top` to PDF coordinates. I wonder if the PDF page width and height is available as metadata on the `view` object.

ui/src/ScholarReader.tsx
```diff
@@ -451,6 +442,33 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
     );
+    const highlightedSymbols: Map<String, BoundingBox> = new Map();
+    if (
+      this.state.symbols !== null &&
+      this.state.mathMls !== null &&
+      this.state.selectedEntityType === "symbol" &&
+      this.state.selectedEntityId !== null
+    ) {
+      // we need these highlights to be in sorted
+      // order by highlight y value so we must use a Map.
```
andrewhead, line 453: Why must we use a map? I think I know but I'm not sure.

ui/src/ScholarReader.tsx
```diff
@@ -451,6 +442,33 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+        this.state.symbols,
+        this.state.mathMls
+      )
+      .map(id => {
+        const symbol = this.state.symbols.byId[id];
+        const symbolBox = symbol.bounding_boxes[0];
+        return [id, symbolBox]
+      })
+      .sort((s1, s2) => {
+        const pg = s1[1].page - s2[1].page;
```
andrewhead, line 465: nit pick: maybe `pg` -> `pgNumberDiff`?

ui/src/ScholarReader.tsx
```diff
@@ -451,6 +442,33 @@ class ScholarReader extends React.PureComponent<ScholarReaderProps, State> {
...
+        return pg === 0 ? s1[1].top - s2[1].top : pg;
+      })
+      .map(([id, bounding]) => {
```
andrewhead, line 468:
```suggestion
      .map(([id, boundingBox]) => {
```

ui/src/FindBar.tsx
andrewhead: (I also see your comment below in ScholarReader about something needing to be a prop. I'm not sure I understand for now---let's catch up over a video call soon.)

ui/src/FindBar.tsx
zkirby: Yes, if the only thing in your constructor is state then it's best to not have the constructor (here is a [blog post](https://overreacted.io/why-do-we-write-super-props/) about it from one of the core maintainers of react if you are curious!).

zkirby: Since state is immutable I believe that this should be safe

ui/src/FindBar.tsx
zkirby: Ah I'm not sure I quite understand, what would that look like accessing it through context? I don't believe it would be possible to access it via the render method since we need to update the state.

ui/src/FindBar.tsx
zkirby: Yes good point! I'll change that :)

ui/src/FindBar.tsx
zkirby: Hahah no didn't have anything in particular, I figured if we keep this implementation then I would go rummaging through pdf.js for a better API :)

ui/src/FindBar.tsx
zkirby: yes :(

ui/src/FindBar.tsx
zkirby: I can change this to null if you want :), I believe returning undefined from render is not allowed

ui/src/PageOverlay.tsx
zkirby: I believe in this case this would actually be okay, I can change it if you want but it will still need to be a prop on the findbar

ui/src/ScholarReader.tsx
zkirby: Yes agreed but unfortunately ts yells at me without this here :(

ui/src/ScholarReader.tsx
zkirby: Need the keys to maintain their order

ui/src/ScholarReader.tsx
zkirby: Aaahhh I was wondering why it wasn't working hahaha, thanks! I'll look into switching this over after we talk if we decide to use it :). If not I'll return it to its original state

PR #104, opened 2020-04-21
Basic Abbreviation Detection

rishamsidhu:
The most basic abbreviation detection model added in to make sure that setup is correct and further edits can just be added on. Does not locate the expansions, only the short form of the abbreviations.


data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,109 @@
...
+        soup_text = []
+        self.get_text(soup, soup_text)
+        contents = ""
+        for p, l in soup_text:
+            contents += p
+
+        contents = re.sub("\n", " ", contents)
+        abb_short_forms = {}
+        doc = nlp(contents)
+        for abrv in doc._.abbreviations:
```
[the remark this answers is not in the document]
rishamsidhu: The scispacy github page specified this as the way to access abbreviations, so I would assume it is safe to access.

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,109 @@
...
+            count = 0
+            for s in symbols:
+                count += str(abrv).count(s)
```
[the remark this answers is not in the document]
rishamsidhu: Symbols is a global variable, a list of common non-alphanumeric characters that often indicate that we have selected something that is not an abbreviation. For example, a { or ^ in an abbreviation means we probably have selected a variable in an equation instead. This is meant as a rough filter to not identify them.

data-processing/entities/abbreviations/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+commands = create_entity_localization_command_sequence(
+    "abbreviations",
+    AbbreviationExtractor,
+    DetectedEntityType=Abbreviation
+)
+
+# Register additional directories to be used by the upload function
+directories.register("abbreviations-model-ids")
```
andrewhead, line 17: This line can be removed. I've refactored the code a bit to make this unnecessary.

data-processing/requirements.txt
```diff
@@ -9,6 +9,8 @@ pysbd
 reportlab
 requests
 TexSoup==0.2.0
+spacy==2.2.3
+scispacy
```
andrewhead, line 13: Can the version number for scispacy be specified?

data-processing/entities/abbreviations/types.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+
+from common.types import SerializableEntity
+
+
+@dataclass(frozen=True)
+class Abbreviation(SerializableEntity):
+    text: str
+    expansion : str
+    #the str_id tells us whether it is a full form or short form (f or s),
```
andrewhead, line 10: Nit pick for comments style:
* one space between `#` and first character
* capitalize the first word
* end each one with a `.`

data-processing/tests/test_extract_abbreviations.py
```diff
@@ -0,0 +1,63 @@
...
+def test_abbreviation(abb, text, start, end, expansion, tex, first, last):
+    #test the equivalence of an abbreviation instance abb and the values it should have
+    assert abb.text == text, \
+        "Abbreviation Short Form Incorrect: got {bad}, but expected {good}".format(bad = abb.text, good = text)
+    assert abb.start == start, \
+        "Abbreviation Start Position Incorrect: got {bad}, but expected {good}".format(bad = abb.start, good = start)
+    assert abb.end == end, \
+        "Abbreviation End Position Incorrect: got {bad}, but expected {good}".format(bad = abb.end, good = end)
+    assert abb.expansion == expansion, \
+        "Abbreviation Expansion Incorrect: got {bad}, but expected {good}".format(bad = abb.expansion, good = expansion)
```
andrewhead, line 59: Wow, look at these detailed error messages! Great!

data-processing/tests/test_extract_abbreviations.py
```diff
@@ -0,0 +1,63 @@
...
+    test_abbreviation(abbs[2], 'AI', 61, 84, "artificial intelligence", "artificial intelligence", 'f', '0')
+    test_abbreviation(abbs[3], 'AI', 86, 88, "artificial intelligence", "AI", 's', '1')
+    assert abbs[2].str_id[1] == abbs[3].str_id[1] == abbs[4].str_id[1], "Different Instances of Abbreviation Do Not Match str_id second char"
+    test_abbreviation(abbs[5], 'CV', 182, 207, "computer vision", "c}omputer \\textbf{v}ision", 'f', '0')
+    test_abbreviation(abbs[6], 'CV', 209, 211, "computer vision", "CV", 's', '1')
+    assert abbs[5].str_id[1] == abbs[6].str_id[1], "Different Instances of Abbreviation Do Not Match str_id second char"
+
+def test_abbreviation(abb, text, start, end, expansion, tex, first, last):
+    #test the equivalence of an abbreviation instance abb and the values it should have
+    assert abb.text == text, \
```
andrewhead, line 52: Cleanup idea: the `got {bad}, but expected {good}` messages can be removed without consequence, as the test runner will already show the expected and actual value when they mismatch if I recall correctly.

data-processing/tests/test_extract_abbreviations.py
```diff
@@ -0,0 +1,63 @@
...
+    abbs.sort(key = lambda x: x.start)
+
+    assert len(abbs) == 6
+    assert abbs[0].str_id[1] == abbs[1].str_id[1] == abbs[5].str_id[1], "Different Instances of Abbreviation Do Not Match str_id second char"
+    assert abbs[2].str_id[1] == abbs[3].str_id[1] == abbs[4].str_id[1], "Different Instances of Abbreviation Do Not Match str_id second char"
+    test_abbreviation(abbs[4], 'AI', 86, 88, "artificial intelligence", "AI", 's', '2')
+    test_abbreviation(abbs[5], 'NLP', 137, 140, "Natural Language Processing", "NLP", 's', '2')
+
+def test_latex():
+    #more realistic example, latex formatting included
```
andrewhead, line 33: great!!

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,163 @@
...
+            scount = 0
+
+            #yields the abbreviated forms as Abbreviation types
+            for location in abb_short_forms[abb]:
+                scount += 1
+                start, end = location
+                tex_sub = tex[start:end]
+                context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
+                yield Abbreviation(
+                    text= abb,
```
andrewhead, line 151: nit pick: a couple of cleanup steps are needed for this code:
* format it with the `black` code formatter (to normalize spacing, etc.)
* run type-checking on it with mypy

Do you use VSCode for writing Python code? The settings in this project are configured so that it should be able to automatically format and type check the code when you save it.

data-processing/entities/abbreviations/extractor.py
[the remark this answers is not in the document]
andrewhead: Ah okay. Could symbols be given a more meaningful name from the context you just described? Maybe something like `NON_ACRONYM_CHARACTERS` (captilalized as it's a constant)?

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,163 @@
...
+                count += str(abrv).count(s)
+            #count makes sure that we don't accidentally include symbols or variables
+            if count == 0:
+                abb_short_forms[str(abrv)] = [[plaintext_to_tex_offset_map[m.start()], plaintext_to_tex_offset_map[m.start() + len(str(abrv))]] for m in re.finditer(str(abrv), plaintext)]
+                abb_expansions[str(abrv)] = str(abrv._.long_form)
+                x = plaintext.find(str(abrv._.long_form))
+                if x != -1:
+                    expanded_locations[str(abrv)] = [plaintext_to_tex_offset_map[x], plaintext_to_tex_offset_map[x + len(str(abrv._.long_form))]]
+                else:
+                    expanded_locations[str(abrv)] = [0, 0]
```
andrewhead, line 115: Does this seem to be working correctly? I remember there were some open questions about that before.

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,163 @@
...
+                    else segment.tex_start
+                )
+                plaintext_to_tex_offset_map[len(plaintext) + i] = tex_offset
+
+            # While building the map, also create a contiguous plaintext string
+            plaintext += segment.text
+            last_segment = segment
+
+        if last_segment is not None:
+            plaintext_to_tex_offset_map[len(plaintext)] = last_segment.tex_end
```
andrewhead, line 87: If I recall correctly, this code also appears in the sentence extractor. Would it make sense to factor it out into another file (say, `common/parse_tex.py`) so that the code can be maintained in just one place?

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,163 @@
...
+                yield Abbreviation(
+                    text= abb,
+                    start=start,
+                    end=end,
+                    expansion = expanded,
+                    id_= count,
+                    tex_path=tex_path,
+                    tex=tex_sub,
+                    context_tex=context_tex,
+                    str_id = "s" + str(fcount) + "-" + str(scount)
```
andrewhead, line 159: nit pick: can the names for `fcount` and `scount` be expanded? It's not clear to me what they stand for currently

data-processing/entities/abbreviations/types.py
andrewhead, line 10: Ah I see! 'f' is full and 's' is short. Can this just be 'full' and 'short'? The meaning of this property may also be clearer with a name like `str_type` or `abbr_pair_type`

data-processing/entities/abbreviations/extractor.py
rishamsidhu: Yes, it is. I guess it might be better to factor it out.

data-processing/entities/abbreviations/extractor.py
rishamsidhu: It definitely works with the current setup, I can get rid of the else case for now.

data-processing/entities/abbreviations/types.py
rishamsidhu: I thought we wanted it all in one string, but it can broken up if preferred.

data-processing/entities/abbreviations/extractor.py
rishamsidhu: I do not use VSCode

data-processing/common/parse_tex.py
```diff
@@ -617,6 +648,44 @@ def parse_soup(tex: str) -> TexSoup:
     except (TypeError, EOFError) as e:
         raise TexSoupParseError(str(e))
 
+def reserved_char_check(tex):
```
andrewhead, line 651: Looks great! One small nit pick: can these function be renamed to be an action rather than a noun (e.g., `check_for_reserved_characters`)

data-processing/common/parse_tex.py
```diff
@@ -32,6 +32,37 @@
...
+    "✂",
+    "⎋",
+    "ᓰ",
+    "ᓱ",
+    "ᓳ",
+    "ᓴ",
+    "ᓷ",
+    "ᓸ",
+]
+
```
andrewhead, line 64: Can this be grouped down with the other code for plaintext / sentence extraction?

data-processing/entities/abbreviations/extractor.py
```diff
@@ -0,0 +1,99 @@
...
+import sys
+
+NON_ACRONYM_CHARACTERS = ["%", "^", "{", "}", "[", "]", "\\", "=", "#", "&", "~", "$", "|", "_", ":", ";"]
+
+class AbbreviationExtractor(EntityExtractor):
+    """
+    Extract plaintext sentences from TeX, with offsets of the characters they correspond to in
+    the input TeX strings. The extracted sentences might include some junk TeX, having the same
+    limitations as the plaintext produced by PlaintextExtractor.
+    """
```
andrewhead, line 22: Update this comment.

data-processing/requirements.txt
```diff
@@ -9,6 +9,9 @@ pysbd
 reportlab
 requests
 TexSoup==0.2.0
+nltk==3.4.5
```
andrewhead, line 12: Is this still necessary? I couldn't find references to it elsewhere in the code.

PR #106, opened 2020-04-22
"View PDF" button on citation tooltip

jocelynzsun:
This is a working version of the feature for issue #90 and it has two parts:

1. API
On the API side, we switched to a new API, because the previous API we were using doesn't contain a link to the PDF of the cited paper (or a link to the website where the paper is published in the case that the paper isn't available to view for free), this link is stored in the property `primaryPaperLink`.
Edits in `s2-api.ts` were made to match with the JSON structure provided by the new API, it's more nested (hence all the interfaces created). The purpose is to make the data structure consistent for the UI, so minimum changes were needed on the UI side for this new API to work for our needs.

2. UI
As detailed in #90 ,now users could access the PDF of the cited paper (or a website where they could access the paper) directly from the citation tooltip, under the paper abstract section.
image
When user click on `View PDF`, the link opens on a new tab. The `<a>` tag uses `rel="noopener noreferrer"`, this is suggested by node.js when compiling for security purposes, more information about these tags could be find [here](https://developers.google.com/web/tools/lighthouse/audits/noopener).

3. More work
I could imagine a better version of this might include the styling of the button, e.g. padding (especially top padding), whether we want to make it a "button" button so it's more consistent with the paper detail page on semantic scholar ([example](https://www.semanticscholar.org/paper/Multimodal-Deep-Learning-Ngiam-Khosla/80e9e3fc3670482c1fee16b2542061b779f47c4f)).

### DIFF
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -1,5 +1,5 @@
 import axios, { AxiosResponse } from "axios";
-import { isS2ApiResponseSuccess, S2ApiPaper } from "./types/s2-api";
+import { isS2ApiResponseSuccess, S2Api } from "./types/s2-api";
 
 /**
  * Base URL for requests to Semantic Scholar API.
@@ -9,7 +9,6 @@ const SEMANTIC_SCHOLAR_API_URL = "https://www.semanticscholar.org/api/1";
 interface Author {
   id: string;
   name: string;
-  url: string;
 }
 
 interface Paper {
@@ -17,11 +16,11 @@ interface Paper {
   title: string;
   authors: Author[];
   abstract: string | null;
-  url: string;
   venue: string | null;
   year: number | null;
   citationVelocity: number;
   influentialCitationCount: number;
+  primaryPaperLink: string;
 }
 
 /**
@@ -49,22 +48,26 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
     return;
   }
   if (isS2ApiResponseSuccess(response)) {
-    const data = response.data as S2ApiPaper;
-    const year = parseInt(data.year) || null;
+    const data = response.data as S2Api;
+    var year;
+    if (data.paper.year.text === undefined) {
+      year = null;
+    } else {
+      year = parseInt(data.paper.year.text)
+    }
     return {
       s2Id,
-      title: data.title,
-      authors: data.authors.map(a => ({
-        id: a.authorId,
+      title: data.paper.title.text,
+      authors: data.paper.authors.map(a => ({
+        id: a.ids,
         name: a.name,
-        url: a.url
       })),
-      abstract: data.abstract,
-      url: data.url,
+      abstract: data.paper.paperAbstract.text,
       year,
-      venue: data.venue,
-      citationVelocity: data.citationVelocity || 0,
-      influentialCitationCount: data.influentialCitationCount || 0
+      venue: data.paper.venue.text,
+      citationVelocity: data.paper.citationStats.citationVelocity || 0,
+      influentialCitationCount: data.paper.citationStats.numKeyCitations || 0,
+      primaryPaperLink: data.paper.primaryPaperLink.url,
     };
   }
 }
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -47,8 +47,12 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
     console.error("Failed to fetch data from Semantic Scholar API", err);
     return;
   }
+
   if (isS2ApiResponseSuccess(response)) {
     const data = response.data as S2Api;
+    if (data.paper === undefined) {
+      return;
+    }
     var year;
     if (data.paper.year.text === undefined) {
       year = null;
@@ -58,7 +62,7 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
     return {
       s2Id,
       title: data.paper.title.text,
-      authors: data.paper.authors.map(a => ({
+      authors: data.paper.authors[0].map(a => ({
         id: a.ids,
         name: a.name,
       })),
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -4,7 +4,7 @@ import { isS2ApiResponseSuccess, S2ApiPaper } from "./types/s2-api";
 /**
  * Base URL for requests to Semantic Scholar API.
  */
-const SEMANTIC_SCHOLAR_API_URL = "http://api.semanticscholar.org/v1";
+const SEMANTIC_SCHOLAR_API_URL = "https://www.semanticscholar.org/api/1";
 
 interface Author {
   id: string;
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -1,5 +1,5 @@
 import axios, { AxiosResponse } from "axios";
-import { isS2ApiResponseSuccess, S2Api } from "./types/s2-api";
+import { isS2ApiResponseSuccess, S2ApiPaperResponse } from "./types/s2-api";
 
 /**
  * Base URL for requests to Semantic Scholar API.
@@ -21,6 +21,7 @@ interface Paper {
   citationVelocity: number;
   influentialCitationCount: number;
   primaryPaperLink: string;
+  paperLinkType: string;
 }
 
 /**
@@ -49,7 +50,7 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
   }
 
   if (isS2ApiResponseSuccess(response)) {
-    const data = response.data as S2Api;
+    const data = response.data as S2ApiPaperResponse;
     if (data.paper === undefined) {
       return;
     }
@@ -72,6 +73,7 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
       citationVelocity: data.paper.citationStats.citationVelocity || 0,
       influentialCitationCount: data.paper.citationStats.numKeyCitations || 0,
       primaryPaperLink: data.paper.primaryPaperLink.url,
+      paperLinkType: data.paper.primaryPaperLink.linkType,
     };
   }
 }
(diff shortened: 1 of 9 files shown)


api/src/types/s2-api.ts
```diff
@@ -13,21 +13,52 @@ interface S2ApiSuccess {
...
-  title: string;
+export interface S2Api {
+  paper: S2ApiPaper;
+}
+
+interface S2ApiPaper {
+  id: string;
+  title: S2ApiTitle;
+  paperAbstract: S2ApiAbstract;
+  authors: S2ApiAuthor[][];
```
andrewhead, line 24: Is `authors` really a list of lists of `S2ApiAuthor`? Or is it just a list of type `S2ApiAuthor`?

api/src/types/s2-api.ts
```diff
@@ -13,21 +13,52 @@ interface S2ApiSuccess {
   status: 200;
 }
 
-export interface S2ApiPaper {
-  abstract: string;
-  arxivId: string | null;
-  authors: S2ApiAuthor[];
-  doi: string;
-  title: string;
+export interface S2Api {
```
andrewhead, line 16: nit pick: `S2Api` feels a bit too vague for this interface name, as it would also describe other calls that we make to the S2 API for data that are not papers. Can we give another name, say `S2ApiPaperResponse`?

api/src/types/s2-api.ts
```diff
@@ -13,21 +13,52 @@ interface S2ApiSuccess {
...
+}
+
+interface S2ApiCitationStats {
+  citationVelocity: number;
+  numKeyCitations: number;
 }
 
 interface S2ApiAuthor {
-  authorId: string;
+  ids: string;
```
andrewhead, line 62: Should this just be `id` if it's a single ID?

ui/src/PaperSummary.tsx
```diff
@@ -101,6 +101,7 @@ export class PaperSummary extends React.PureComponent<
           const hasMetrics = paper.citationVelocity !== 0 || paper.influentialCitationCount !== 0;
           const truncatedAbstract = paper.abstract ? truncateText(paper.abstract, 300) : null;
           const inLibrary = userLibrary ? userLibrary.paperIds.includes(this.props.paperId) : false;
+          const summaryUrl = `https://www.semanticscholar.org/paper/{paper.Id}`
```
andrewhead, line 104: Should this be `id` with a lower case `i`?
```suggestion
          const summaryUrl = `https://www.semanticscholar.org/paper/{paper.id}`
```

api/src/types/s2-api.ts
jocelynzsun: image

this is what the interface has :/

api/src/types/s2-api.ts
jocelynzsun: Annotation 2020-05-19 155413

all authors are contained in a list with a single item, that's just the way this api is :/

api/src/s2-api.ts
```diff
@@ -48,23 +48,32 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
...
+      year = null;
+    } else {
+      year = parseInt(data.paper.year.text)
+    }
     return {
       s2Id,
-      title: data.title,
-      authors: data.authors.map(a => ({
-        id: a.authorId,
+      title: data.paper.title.text,
```
andrewhead, line 65: Jocelyn, do we always know that the properties will be defined and non-null? If you haven't already, can you check to see if these properties will be defined for a set of different papers?

api/src/s2-api.ts
```diff
@@ -48,23 +48,32 @@ async function getPaper(s2Id: string): Promise<Paper | undefined> {
...
+      abstract: data.paper.paperAbstract.text,
       year,
-      venue: data.venue,
-      citationVelocity: data.citationVelocity || 0,
-      influentialCitationCount: data.influentialCitationCount || 0
+      venue: data.paper.venue.text,
+      citationVelocity: data.paper.citationStats.citationVelocity || 0,
+      influentialCitationCount: data.paper.citationStats.numKeyCitations || 0,
+      primaryPaperLink: data.paper.primaryPaperLink.url,
+      paperLinkType: data.paper.primaryPaperLink.linkType,
```
andrewhead, line 76: nit pick: can this be `primaryPaperLinkType` to make it clear that it is related to `primaryPaperLink`?

ui/src/Citation.tsx
```diff
@@ -13,11 +13,12 @@ export class Citation extends React.PureComponent<CitationProperties> {
       <ScholarReaderContext.Consumer>
         {({ papers, setJumpPaperId, setSelectedCitation }) => {
           const paper = papers[this.props.citation.paper];
+          console.log(paper.paperLinkType);
```
andrewhead, line 16: nit pick: remove this console.log before we merge

ui/src/Citation.tsx
```diff
@@ -45,6 +46,11 @@ export class Citation extends React.PureComponent<CitationProperties> {
                     </span>
                   </div>
                 ) : null}
+                {paper.primaryPaperLink !== null ? (
```
andrewhead, line 49: Will this ever be null? If so, can you update the `Paper` interface declaration for the `primaryPaperLink` property in `api/src/s2-api.ts` to indicate that it's sometimes null?

ui/src/PaperSummary.tsx
```diff
@@ -101,6 +101,7 @@ export class PaperSummary extends React.PureComponent<
           const hasMetrics = paper.citationVelocity !== 0 || paper.influentialCitationCount !== 0;
           const truncatedAbstract = paper.abstract ? truncateText(paper.abstract, 300) : null;
           const inLibrary = userLibrary ? userLibrary.paperIds.includes(this.props.paperId) : false;
+          const summaryUrl = `https://www.semanticscholar.org/paper/{paper.s2Id}`
```
andrewhead, line 104: nit pick: maybe this name? 

```suggestion
          const paperUrl = `https://www.semanticscholar.org/paper/{paper.s2Id}`
```

Alternatively `pdpUrl` (short for 'paper details page URL') or `paperDetailsUrl` might make sense too! This is the naming convention used at Semantic Scholar to refer to the page at this URL.

ui/src/style/variables.less
```diff
@@ -17,6 +17,9 @@
 @equation-user-annotation-color: #e82;
 @symbol-user-annotation-color: #f5f;
 
+@button-background-color: #1857b6;
```
andrewhead, line 20: nit pick: perhaps this naming to make it clearer which buttons this style will apply to?

```suggestion
@s2-primary-action-background-color: #1857b6;
```

ui/src/style/citation.less
```diff
@@ -34,3 +34,32 @@
...
+ * Citation utilities.
+ */
+.citation__utils {
+  padding-top: @padding;
+}
+
+/**
+ * button for viewing paper in PDF
+ */
+.citation__view__pdf {
```
andrewhead, line 48: This is some good reverse engineering! 👍

ui/src/style/variables.less
```diff
@@ -56,6 +61,8 @@
  */
 @large: 1.6rem;
 @medium: 1.4rem;
+@button-font-size: 14px;
```
andrewhead, line 64: See note above about naming for the button

ui/src/types/api.ts
```diff
@@ -45,11 +45,12 @@ export interface Paper {
   title: string;
   authors: Author[];
   abstract: string | null;
-  url: string;
   venue: string | null;
   year: number | null;
   citationVelocity: number;
   influentialCitationCount: number;
+  primaryPaperLink: string;
+  paperLinkType: string;
```
andrewhead, line 53: Should these be `string | null`? I think this depends on what the S2 API returns.

ui/src/ui-utils.ts
```diff
@@ -64,3 +64,62 @@ export function truncateText(text: string, limit: number, withEllipsis: boolean
     return text;
   }
 }
+
+/**
+ * Get the correct display name for certain publishers.
+ *
```
andrewhead, line 70:
```suggestion
```

ui/src/ui-utils.ts
```diff
@@ -64,3 +64,62 @@ export function truncateText(text: string, limit: number, withEllipsis: boolean
     return text;
   }
 }
+
+/**
+ * Get the correct display name for certain publishers.
+ * 
+ * @param {string} the type of primaryPaperLInk
+ */
+export function getLinkText(input: string): string {
```
andrewhead, line 73: nit pick: can we use a more descriptive name than `input` here? Perhaps `linkType`?

ui/src/ui-utils.ts
```diff
@@ -64,3 +64,62 @@ export function truncateText(text: string, limit: number, withEllipsis: boolean
...
+
+  var linkText: string = "";
+
+  if (input.length > 17) {
```
andrewhead, line 77: can you give 17 a descriptive constant name, describing what the value means?

```suggestion

  const <DESCRIPTIVE_CONSTANT_NAME> = 17;
  if (input.length > <DESCRIPTIVE_CONSTANT_NAME>) {
```

ui/src/ui-utils.ts
```diff
@@ -64,3 +64,62 @@ export function truncateText(text: string, limit: number, withEllipsis: boolean
...
+    linkText = "View Via Publisher";
+    return linkText;
+  }
+
+  switch (input) {
+    case "acm":
+      linkText = "View On ACM";
+      break;
```
andrewhead, line 85: Why not just: 
```suggestion
      return "View On ACM";
```

Could that let you get rid of the `break;`?

ui/src/types/api.ts
jocelynzsun: upon inspecting a sample of papers, I found that some of the older papers (published in the 20th century) didn't get published on the web, therefore the api doesn't even have the `primaryPaperLink` field for it. (example: https://www.semanticscholar.org/api/1/paper/ac5e724f4223720c1b4f5bcc4c7b8c11ccea83c1)

will this cause the code to break, because the property doesn't exist?

PR #120, opened 2020-06-26
initial version of improvement for SentenceExtractor

dykang:
- additional fields: cleaned_text, is_sentence,
current_{section,figure,table}, label, ref, cite, symbol, others
 - detecting richer tex context from context_tex
 - regex for detecting various tex entities (e.g., section, label,
abstract, table, figure, reference, citation, symbol, etc)
 - cleaned_text has all reference entities with placeholders like
CITATION, SYMBOL, TABLE, FIGURE, SECTION, and EQUATION.


data-processing/entities/sentences/types.py
```diff
@@ -6,3 +6,14 @@
 @dataclass(frozen=True)
 class Sentence(SerializableEntity):
     text: str
+    cleaned_text: str
```
andrewhead, line 9: Why not just use the value of 'cleaned_text' for 'text'? I don't think anyone will plan to use the old 'text' field once we have a more cleaned text field.

data-processing/entities/sentences/types.py
```diff
@@ -6,3 +6,14 @@
...
+    is_sentence: bool
```
andrewhead, line 10: Why would this be false? Can the SentenceExtractor just not yield a sentence if it's not detected to be a sentence?

data-processing/entities/sentences/extractor.py
```diff
@@ -1,5 +1,9 @@
 import logging
+import re, regex
 from typing import Iterator, List
+import numpy as np
+from nltk import word_tokenize
```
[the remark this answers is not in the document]
andrewhead: Oh never mind you already added it to the requirements.txt file!

data-processing/requirements.txt
```diff
@@ -9,6 +9,7 @@ pysbd
 reportlab
 requests
 TexSoup==0.2.0
+nltk
```
andrewhead, line 12: Clarification question: will things go wrong if another more recent version of nltk is installed? If so, then perhaps a version number should be specified here

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+    for c in tex:
+        if c in list(NESTING_CHARACTERS_MAPPING.keys()) + list(NESTING_CHARACTERS_MAPPING.values()):
+            nesting_characters_in_tex.append(c)
+    if not nesting_characters_in_tex:
+        return True
+    return check_nesting_structure(nesting_characters_in_tex)
+
+
+
+def extract_richer_tex(context_tex, tex):
```
andrewhead, line 106: Can you add type annotations to the functions (argument types and return types)?

andrewhead, line 106: not entirely sure I understand what's going on  in this function. Is it just splitting context_tex into lines, and removing all but the line before and after? What's richer about the returned context?

data-processing/entities/sentences/extractor.py
```diff
@@ -123,6 +210,102 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+                entity_dict['figure_begin'] = re.findall(PATTERN_FIGURE_BEGIN, extended_tex_sub)
+                entity_dict['figure_end'] = re.findall(PATTERN_FIGURE_END, extended_tex_sub)
+                entity_dict['ref'] = re.findall(PATTERN_REF, extended_tex_sub)
+                entity_dict['cite'] = re.findall(PATTERN_CITE, extended_tex_sub)
+                entity_dict['symbol'] = regex.findall(PATTERN_SYMBOL, extended_tex_sub, overlapped=False)
+                entity_dict['any'] = re.findall(PATTERN_ANY, extended_tex_sub)
+
+                # store 'others' field by [any] - [section,label,...]
+                if len(entity_dict['any']) > 0:
+                    entity_values_flattend = [fone for f in entity_dict.keys() if f != 'any' for fone in entity_dict[f]]
```
andrewhead, line 235:
```suggestion
                    entity_values_flattened = [fone for f in entity_dict.keys() if f != 'any' for fone in entity_dict[f]]
```

data-processing/entities/sentences/extractor.py
```diff
@@ -82,6 +166,9 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
         segmenter = pysbd.Segmenter(language="en", clean=False)
         # Record the current length of the plain text so account for the extractor bug
         length_so_far_in_plain_text = 0
+
+
+        current_section, current_table, current_figure = False, False, False
```
andrewhead, line 171: nit pick: if current_section is a string, should its default be None? And if the other two variables track whether the sentence is in a table or figure but not the identity of the current table or figure, are these better names?

```suggestion
        current_section, in_table, in_figure = None, False, False
```

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+        return None
+    surrounding_tex = context_tex.split(tex)
+    if len(surrounding_tex) > 2:
+        surrounding_tex = [tex.join(surrounding_tex[:-1]), surrounding_tex[-1]]
+    before_tex, after_tex = surrounding_tex
+    richer_tex = before_tex.split('\n')[-1] + tex + after_tex.split('\n')[0]
+    return richer_tex
+
+
+def extract_text_from_tex_entity(entity):
```
andrewhead, line 120: Is this a more accurate name? Is this just extracting text from a TeX group or an argument to a command?

```suggestion
def extract_text_from_tex_group(entity):
```

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+    return entity[entity.find("{")+1:entity.find("}")]
```
andrewhead, line 121: nitpick: spacing around operators. Did you get a chance to run the "black" Python code formatter on this code, as well as the mypy type checker? If not, can you?

data-processing/entities/sentences/extractor.py
```diff
@@ -123,6 +210,102 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
             length_so_far_in_plain_text = plaintext_end
             tex_sub = tex[start:end]
             context_tex = tex[start - DEFAULT_CONTEXT_SIZE : end + DEFAULT_CONTEXT_SIZE]
+
+            # extract richer context of tex using '\n'
+            extended_tex_sub = extract_richer_tex(context_tex, tex_sub)
+
+            # detect entity fields using pre-defined regex
+            entity_dict = {}
```
andrewhead, line 218: nitpick: can another word besides `entity` be used here? Elsewhere in this codebase, entity is used to refer to a clickable object in the paper---like a sentence, a citation, a symbol, etc.

data-processing/entities/sentences/extractor.py
```diff
@@ -123,6 +210,102 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+            if current_figure:
+                entity_dict['current_figure'] = current_figure
+            if current_table:
+                entity_dict['current_table'] = current_table
+
+            # detect whether current line is sentence or not
+            is_sentence = check_sentence_or_not(tex_sub, entity_dict)
+
+            # clean sentence
+            cleaned_sentence = "{}".format(sentence) if is_sentence else ""
```
andrewhead, line 277: nitpick: why not just `sentence if is_sentence else ""`?

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+        return False
+    # check whether text is caption in figure/table or not. currently, we remove all text including captions in figure/table
+    # TODO @dykang later, distinguish lines in table and lines in captions
+    if 'current_figure' in entity_dict and entity_dict['current_figure']:
+        return False
+    if 'current_table' in entity_dict and entity_dict['current_table']:
+        return False
+
+    # check nesting structures from tex_sub
+    nesting_characters_in_tex = []
```
andrewhead, line 96: What is getting checked about nesting structures here? What are the criteria under which a sentence with nesting would be considered invalid? Can that be made a little more clear?

data-processing/entities/sentences/extractor.py
```diff
@@ -1,5 +1,9 @@
 import logging
+import re, regex
```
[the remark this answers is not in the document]
dykang: That's a good point. I originally used `re` only but there was some feature only supported by `regex`. I guess regex includes every feature of `re` so I am deleting `re` now.

data-processing/entities/sentences/extractor.py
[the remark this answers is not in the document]
dykang: Good points! I originally added my featurization functions in extractor.py but then I moved them to definition extractor now. So, both numpy and nltk should be deleted here.

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+PATTERN_SECTION = r"\\(?:sub)*section[*]*\{[A-Za-z0-9 \{\}\\_.,:-]*\}"
+PATTERN_LABEL = r"\\label\{[A-Za-z0-9 \\_.,:-]*\}"
+PATTERN_ABSTRACT_BEGIN = r"\\begin\{abstract\}"
+PATTERN_ABSTRACT_END = r"\\end\{abstract\}"
+PATTERN_TABLE_BEGIN = r"\\begin\{table[*]*\}"
+PATTERN_TABLE_END = r"\\end\{table[*]*\}"
+PATTERN_FIGURE_BEGIN = r"\\begin\{figure[*]*\}"
+PATTERN_FIGURE_END = r"\\end\{figure[*]*\}"
+PATTERN_REF = r"\\ref\{[A-Za-z0-9 \\_.,:-]*\}"
+#TODO @dykang Fix to capture citations patterns like \cite[GRUs;][]{cho2004}
```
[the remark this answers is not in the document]
dykang: Exactly!

data-processing/entities/sentences/extractor.py
```diff
@@ -37,6 +41,86 @@
...
+PATTERN_CITE = r"\\cite[A-Za-z0-9 \\_.,:-]*\{[A-Za-z0-9 \\_.,:-]*\}"
+PATTERN_SYMBOL = r"\$[A-Za-z0-9\\ \{\}\(\)\[\]^&*_.,\+:\;-\=#]*\$"
+PATTERN_ANY = r"\\[A-Za-z0-9\\_.,:-]*[\{[A-Za-z0-9 \\_.,:-]*\}]*"
+
+# objects for detecting nesting structures from the extended tex
+NESTING_CHARACTERS_MAPPING = {'{':'}', '(':')', '[':']'}
```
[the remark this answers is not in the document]
dykang: You are right! I deleted parentheses now.

data-processing/entities/sentences/extractor.py
dykang: There are some cases that the previous section/figure/table filters do not cover. For instance, some lines in section declaration have very noisy lines like `Introduction}` or `-9pt}`. Those cases can be simply filtered by checking their nesting structures. I will add more descriptions in the function.

data-processing/entities/sentences/extractor.py
dykang: Thanks for pointing this out! I should add them before I pushed! Done now.

data-processing/entities/sentences/extractor.py
dykang: The new name sounds better!

data-processing/entities/sentences/extractor.py
dykang: I should have done this before I commit! Thanks to the pointers. Done!

data-processing/entities/sentences/extractor.py
dykang: Yes, I changed the names from `in_{table,figure}` to `is_sentence_in_{table,figure}`. Also, type for `current_section` is changed to `str`.

data-processing/entities/sentences/extractor.py
dykang: That's a valid point! How about `tex_unit`? Let me know if you have a better idea!

data-processing/entities/sentences/types.py
dykang: Sure! I deleted `text`.

data-processing/entities/sentences/types.py
dykang: If `SentenceExtractor` is completely perfect, yes. However, for debugging purposes, I kept `is_sentence` alive to see our logic of checking a line is a true sentence or not. Later, I agree that `is_sentence` should be deleted. I will make a note that this field should be deleted in the near future once we have a stable version of extractor.

data-processing/requirements.txt
dykang: Good point! Added version info. Though, I only use `nltk` and `numpy` in definition extractor now.

dykang:
Ok, I've changed most of your comments, @andrewhead. Would you please do second pass on the latest codes?

dykang:
Some bugs caused by the current sentence extractor:
- some user-defined macros written in between sections are detected as sentences
- sentence/paragraph breakers are malfunctioning
- text in \label{} is regarded as sentences
- multi-tex merging causes issues in proper ordering of sentences
- some math notations are not properly substituted by [[math]] tags
- need algorithm regex

Note: feel free to add more issues

data-processing/entities/sentences/extractor.py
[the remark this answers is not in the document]
andrewhead: Sounds good! Maybe leave a note in the comments mentioning that the code can use the CitationExtractor here in the future in case higher accuracy is needed?

data-processing/entities/sentences/types.py
andrewhead: Okay sounds good. I think it's also okay to leave it in, the field should just be accompanied by a note that describes why it's there, to answer the question "wait, aren't all Sentence objects sentences?", or maybe given another name

data-processing/entities/sentences/extractor.py
andrewhead: Got it, it makes sense why you'd want this! And I'm excited to see the new comment. I'm still not entirely sure though what all is in the richer_tex. Is it the line before and after?

data-processing/entities/sentences/extractor.py
andrewhead: Sounds good! Incidentally do you use VSCode? It has a bunch of great built-in support for autocomplete, autoformatting, and automated type-checking. If that sounds like something you're interested in, I'd be happy to walk you through setting it up.

data-processing/entities/sentences/extractor.py
andrewhead: sure i love it!

data-processing/entities/sentences/types.py
```diff
@@ -1,8 +1,21 @@
 from dataclasses import dataclass
+from typing import List
 
 from common.types import SerializableEntity
 
 
 @dataclass(frozen=True)
 class Sentence(SerializableEntity):
     text: str
+    extended_tex: str
```
andrewhead, line 10: Would it make sense to just set the `context_tex` property to the more refined context extracted in the code, instead of defining another property? The only reason that `context_tex` currently has a fixed offset before and after the sentence is because it was easy to do at the time 😉

data-processing/entities/sentences/extractor.py
```diff
@@ -123,12 +233,165 @@ def parse(self, tex_path: str, tex: str) -> Iterator[Sentence]:
...
+                    current_section = "ABSTRACT"
+                if len(tex_unit_dict["abstract_end"]) > 0:
+                    current_section = None
+                if len(tex_unit_dict["section"]) > 0:
+                    section = tex_unit_dict["section"][0]
+                    current_section = extract_text_from_tex_group(section)
+
+                # store figure/table information
+                # TODO @dykang add \label{} in table/figure for better matching
+                if len(tex_unit_dict["figure_begin"]) > 0:
```
andrewhead, line 309: just noticing the repetitious code, it occurs to me that maybe a more succinct or clear rewriting would be something like:

```python
is_sentence_in_figure =tex_unit_dict["cite"] = (
  regex.findall(PATTERN_FIGURE_BEGIN, extended_tex_sub) and
  regex.findall(PATTERN_FIGURE_END, extended_tex_sub)
```

for each of the types of tex units. No biggie and not important, but just wanted to throw the idea out there.

andrewhead:
Hey @dykang! I took another look, and it's looking great. I'll take one more stylistic look at this once the edits to the sentence extractor are integrated for improving accuracy. Great work!!

andrewhead:
It took quite a long time (sorry 'bout that!!), though I've finished analyzing my first paper of sentences. Here are the major observations:

Count of Comments

The most common issues were:

1 . Sentences that were commented out using `\iffalse` should presumably have not counted as sentences ("not shown in paper")
2. Arguments to \begin{} and \end{} were extracted as sentences much of the time (like `definition}` from `\begin{definition}`  ("env argument")
3. Sometimes it looked like several sentences should be combined into a larger one. The most frequent example of this was when there was a display style equation in the middle of a sentence ("We define x to be... <display style equation with x>"... where the terms are..."). This happened a lot in this particular paper ("should be combined")
4. Less common cases involved sentences that should have been extracted from captions but aren't yet ("caption text"), I though that section headers should be extracted as sentences (curious what your thoughts are on this) ("section should be sentence"), and that the value of labels was added as a token to the text ("label literal")

Because of these issues, I'm seeing 43% accuracy for sentences and 52% for the sections. Though admittedly, a huge amount of that is because a huge number of the sentences that were extracted were commented out in the TeX

So I'd expect to see those accuracy numbers spike quite a bit once we the sentence extractor can check for `\if 0` and `\iffalse`

data-processing/entities/sentences/extractor.py
dykang: `regex.findall(PATTERN_FIGURE_BEGIN, extended_tex_sub)` only checks whether the current line includes \begin{figure} or not, since input tex only includes "single" line of tex, not a batch of entire tex lines for a figure. So, checking begin and end conditions together do not capture the block of lines in figure.

data-processing/entities/sentences/types.py
dykang: This is hard to decide because sometimes `context_tex` has more accurate context than `extended_tex`, so I kept both of them for now. Once we have a better tex normalized in the later version, I will keep one of them for sure.

PR #151, opened 2020-10-23
Strip ArXiv version when fetching S2 metadata

rreas:
I'd like to do something like this so that versions can run in the pipeline without any changes on the S2 side.

Just a starting point, I haven't checked thoroughly for S2 querying.

### DIFF
--- a/data-processing/common/commands/fetch_s2_data.py
+++ b/data-processing/common/commands/fetch_s2_data.py
@@ -33,7 +33,10 @@ class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
     def process(self, item: ArxivId) -> Iterator[S2Metadata]:
         # XXX(andrewhead): S2 API does not have versions of arXiv papers. I don't think this
         # will be an issue, but it's something to pay attention to.
-        resp = requests.get(f"https://api.semanticscholar.org/v1/paper/arXiv:{item}")
+        versionless_id = self._strip_arxiv_version(item)
+        resp = requests.get(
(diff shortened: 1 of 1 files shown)


data-processing/common/commands/fetch_s2_data.py
```diff
@@ -81,3 +83,43 @@ def save(self, item: ArxivId, result: S2Metadata) -> None:
         s2_id_path = os.path.join(s2_metadata_dir, "s2_id")
         with open(s2_id_path, "w") as s2_id_file:
             s2_id_file.write(result.s2_id)
+
+    def _strip_arxiv_version(self, item: ArxivId) -> ArxivId:
+        """Remove any version identifier from the passed ArXiv ID.
+        Looks for the last 'v' (case-insensitive) and strips beyond it.
```
andrewhead, line 89: Why search strip after the last v and not the first? Is the reasoning that there might be a 'v' in some old-style arXiv IDs?

andrewhead (approved):
Looks good to me! I just had one small question behind the ideal behavior of the version-stripping code. Approved, and feel free to merge away 👍

data-processing/common/commands/fetch_s2_data.py
rreas: Actually you just made me realize something I missed, will update the PR for identifiers like "math.VX/12133" where they are not versioned but the old ArXiv ID has a "v" in the sub-field piece (VX is made-up here).

Specifically to your question, I just think this case isn't going to happen in practice and so I just picked one way and added the example to illustrate.

andrewhead (approved):
LGTM!

PR #154, opened 2020-11-13
Increase request header limit for nginx

mjlangan:
Ref https://github.com/allenai/scholar/issues/24644

S2 (primarily for internal users) can use very large cookies, which often exceed default header size limits. This is a follow-up to https://github.com/allenai/scholarphi/pull/153 that increases the nginx limit to match the limit used by the API server (and, also, S2)

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -102,7 +102,6 @@ function(
             annotations: annotations + {
                 'cert-manager.io/cluster-issuer': 'letsencrypt-prod',
                 'kubernetes.io/ingress.class': 'nginx',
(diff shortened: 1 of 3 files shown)


codeviking (approved):
LGTM, though you'll need something [here](https://github.com/allenai/scholarphi/blob/master/.skiff/webapp.jsonnet#L105) too.  We run Ingress NGINX -- so there's a proxy that sits in front of the proxy that you're modifying here.

You'll add an annotation that looks something like this:

```
'nginx.ingress.kubernetes.io/configuration-snippet': 'client_header_buffer_size 64k;\n large_client_header_buffers 4 64k',
```

PR #155, opened 2020-11-23
Run ScholarPhi against S2 managed database

rreas:
Gathers changes to run ScholarPhi against S2 database:

1. Move all DB parameters to environment variables
2. Clean-up with try/finally when `--one-paper-at-a-time`
3. Change DB name to 'scholarphi' and use public schema as we have different DB instances for dev vs. prod

This cannot be merged yet b/c the production Skiff application probably cannot reach the DB instances in the private network (AI2 inf VPC).

### DIFF
--- a/data-processing/common/commands/database.py
+++ b/data-processing/common/commands/database.py
@@ -14,7 +14,7 @@ class DatabaseUploadCommand(ArxivBatchCommand[I, R], ABC):
 
     def __init__(self, args: Any) -> None:
         super().__init__(args)
-        init_database_connections(schema_name=args.schema, create_tables=True)
+        init_database_connections(schema_name=args.schema, create_tables=False)
 
     @staticmethod
     def init_parser(parser: ArgumentParser) -> None:
--- a/data-processing/common/commands/database.py
+++ b/data-processing/common/commands/database.py
@@ -14,7 +14,9 @@ class DatabaseUploadCommand(ArxivBatchCommand[I, R], ABC):
 
     def __init__(self, args: Any) -> None:
         super().__init__(args)
-        init_database_connections(schema_name=args.schema, create_tables=False)
+        init_database_connections(
+            schema_name=args.schema, create_tables=args.create_tables
+        )
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


rreas:
See also https://github.com/allenai/scholar/issues/25164

rreas:
I'm not too sure how secrets needed here would make it to Skiff on deploy. The new RDS instances are accessible from GCloud at this point.

data-processing/common/commands/database.py
```diff
@@ -14,7 +14,7 @@ class DatabaseUploadCommand(ArxivBatchCommand[I, R], ABC):
 
     def __init__(self, args: Any) -> None:
         super().__init__(args)
-        init_database_connections(schema_name=args.schema, create_tables=True)
+        init_database_connections(schema_name=args.schema, create_tables=False)
```
andrewhead, line 17: I do like the ability to create tables by default if someone is uploading data to a short-lived test schema. Is there some way to expose the ability to create tables through command line arguments if we're going to disable default creation of tables?

rreas: I added `--database-create-tables` flag to address this. By default it's False b/c without more changes providing True requires admin permissions on the DB database.

data-processing/scripts/run_pipeline.py
```diff
@@ -288,6 +289,14 @@ def run_commands_for_arxiv_ids(
             + "(see https://www.postgresql.org/docs/9.2/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS)."
         ),
     )
+    parser.add_argument(
+        "--database-create-tables",
+        type=bool,
+        default=False,
```
andrewhead: This behavior can also be expressed with the shorthand `action="store_true"`

andrewhead (approved):

data-processing/scripts/run_pipeline.py
rreas: Yep! I had a second commit yesterday to use that https://github.com/allenai/scholarphi/pull/155/commits/22be9915f6ef83eddce505d8101d5ee74d3b253b

PR #159, opened 2020-12-03
arXiv version check API & various UI code tweaks

mjlangan:
There's a couple things in this PR:
1. Adds an API route at `papers/arxiv:{arxivId}/version` to return the latest version of a paper we have in the DB
2. `yarn run dev` in the /ui/ directory will start up a local version of the UI
3. The remote logger is now the default export of `logging.ts`, so it can be directly imported into files where it is needed (and we don't need to instantiate it with function calls or in a constructor or something like that). This is the same way the S2 webapp makes its client-side logger available to UI code.
4. ScholarReader component uses fat arrow functions, meaning we don't need all the `.bind(this)` goo in the constructor
5. Some tweaks to PaperSummary too. I also replaced empty element tags with `React.Fragment` to be more explicit, but this is subjective.

This adds the API but does not yet call it from the UI to actually perform the version check at startup.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -154,5 +154,27 @@ export const plugin = {
         },
       },
     });
+
+    server.route({
+      method: "GET",
+      path: "papers/arxiv:{arxivId}/version",
+      handler: async (request, h) => {
+        const arxivId = request.params.arxivId;
+        const version = await dbConnection.getLatestProcessedArxivVersion({ arxiv_id: arxivId });
+        if (version < 0) {
+          // We don't have version info for this ID
+          return h.response().code(404);
+        }
+        // TODO (mjlangan): solidify the response structure (maybe 'maxVersion' or 'latestVersion'?)
+        return h.response({ version }).code(200);
+      },
+      options: {
+        validate: {
+          params: validation.arxivId.append({
+            arxivId: Joi.string().required(),
+          }),
+        },
+      },
+    });
   },
 };
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -161,18 +161,15 @@ export const plugin = {
       handler: async (request, h) => {
         const arxivId = request.params.arxivId;
         const version = await dbConnection.getLatestProcessedArxivVersion({ arxiv_id: arxivId });
-        if (version < 0) {
+        if (!version) {
           // We don't have version info for this ID
           return h.response().code(404);
         }
-        // TODO (mjlangan): solidify the response structure (maybe 'maxVersion' or 'latestVersion'?)
         return h.response({ version }).code(200);
       },
       options: {
         validate: {
-          params: validation.arxivId.append({
-            arxivId: Joi.string().required(),
-          }),
+          params: validation.arxivId,
         },
       },
     });
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -112,17 +112,17 @@ export class Connection {
     return isNaN(version) ? null : version;
   }
 
-  async getLatestProcessedArxivVersion(paperSelector: PaperSelector): Promise<number> {
+  async getLatestProcessedArxivVersion(paperSelector: PaperSelector): Promise<number | null> {
     if (isS2Selector(paperSelector)) {
-      // TODO (mjlangan): Error, or maybe null, think about this
-      return -1;
+      return null;
     }
     // Provided arXiv IDs might have a version suffix, but ignore that for this check.
     const versionDelimiterIndex = paperSelector.arxiv_id.indexOf('v');
     const arxivId = versionDelimiterIndex > -1 ? paperSelector.arxiv_id.substring(0, versionDelimiterIndex) : paperSelector.arxiv_id;
 
+    // TODO(mjlangan): This won't support arXiv IDs prior to 03/2007 as written
     const response = await this._knex.raw<{ rows: { arxiv_version: number }[] }>(`
-      SELECT CAST((REGEXP_MATCHES(arxiv_id,'^\\d{4}\\.\\d{5}v(\\d+)$'))[1] AS integer) AS arxiv_version
+      SELECT CAST((REGEXP_MATCHES(arxiv_id,'^\\d{4}\\.\\d{4,5}v(\\d+)$'))[1] AS integer) AS arxiv_version
         FROM paper
         WHERE arxiv_id ilike ?
         ORDER BY arxiv_version DESC
@@ -133,8 +133,7 @@ export class Connection {
       return response.rows[0].arxiv_version;
     }
 
-    // TODO (mjlangan): Error, or maybe null, think about this
-    return -1;
+    return null;
   }
 
   createBoundingBoxes(
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -103,13 +103,38 @@ export class Connection {
       }));
   }
 
-  async getLatestPaperDataVersion(paperSelector: PaperSelector) {
+  async getLatestPaperDataVersion(paperSelector: PaperSelector): Promise<number | null> {
     const rows = await this._knex("version")
       .max("index")
       .join("paper", { "paper.s2_id": "version.paper_id" })
       .where(paperSelector);
     const version = Number(rows[0].max);
     return isNaN(version) ? null : version;
+  }
+
+  async getLatestProcessedArxivVersion(paperSelector: PaperSelector): Promise<number> {
+    if (isS2Selector(paperSelector)) {
+      // TODO (mjlangan): Error, or maybe null, think about this
+      return -1;
+    }
+    // Provided arXiv IDs might have a version suffix, but ignore that for this check.
+    const versionDelimiterIndex = paperSelector.arxiv_id.indexOf('v');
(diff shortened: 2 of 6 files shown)


api/src/api.ts
```diff
@@ -154,5 +154,27 @@ export const plugin = {
...
+      path: "papers/arxiv:{arxivId}/version",
+      handler: async (request, h) => {
+        const arxivId = request.params.arxivId;
+        const version = await dbConnection.getLatestProcessedArxivVersion({ arxiv_id: arxivId });
+        if (version < 0) {
+          // We don't have version info for this ID
+          return h.response().code(404);
+        }
+        // TODO (mjlangan): solidify the response structure (maybe 'maxVersion' or 'latestVersion'?)
+        return h.response({ version }).code(200);
```
andrewhead, line 169: @mjlangan is this open question something we should discuss as part of this PR?

api/src/api.ts
```diff
@@ -154,5 +154,27 @@ export const plugin = {
...
+      },
+      options: {
+        validate: {
+          params: validation.arxivId.append({
+            arxivId: Joi.string().required(),
```
andrewhead, line 174: Is this extra validation (`.required()`) necessary? I thought hapi would require arXiv ID to be specified by default due to how it parses the path, though I might be wrong.

api/src/db-connection.ts
```diff
@@ -112,6 +112,31 @@ export class Connection {
     return isNaN(version) ? null : version;
   }
 
+  async getLatestProcessedArxivVersion(paperSelector: PaperSelector): Promise<number> {
+    if (isS2Selector(paperSelector)) {
+      // TODO (mjlangan): Error, or maybe null, think about this
```
andrewhead, line 117: @mjlangan would it be helpful to discuss this?

api/src/db-connection.ts
```diff
@@ -112,6 +112,31 @@ export class Connection {
...
+      return -1;
+    }
+    // Provided arXiv IDs might have a version suffix, but ignore that for this check.
+    const versionDelimiterIndex = paperSelector.arxiv_id.indexOf('v');
```
andrewhead, line 121: Perhaps also take a look at Russell's code for stripping the version ID on the backend. One corner case I noticed in his code is that old-style arXiv IDs can have 'v' in them (like `math.VT/######v#`). I'm not sure if that's worth handling here, though maybe it is.

https://github.com/allenai/scholarphi/blob/aeb4d577d43f31f559dfd627f0ec2272d204006f/data-processing/common/commands/fetch_s2_data.py#L87

api/src/db-connection.ts
```diff
@@ -112,6 +112,31 @@ export class Connection {
...
+    const arxivId = versionDelimiterIndex > -1 ? paperSelector.arxiv_id.substring(0, versionDelimiterIndex) : paperSelector.arxiv_id;
+
+    const response = await this._knex.raw<{ rows: { arxiv_version: number }[] }>(`
+      SELECT CAST((REGEXP_MATCHES(arxiv_id,'^\\d{4}\\.\\d{5}v(\\d+)$'))[1] AS integer) AS arxiv_version
```
andrewhead, line 125: It looks like older-style arXiv IDs before 2015 will have 4 digits after the dot instead of 5: https://arxiv.org/help/arxiv_identifier

api/src/db-connection.ts
```diff
@@ -112,6 +112,31 @@ export class Connection {
...
+        FROM paper
+        WHERE arxiv_id ilike ?
+        ORDER BY arxiv_version DESC
```
andrewhead, line 128: will this ordering handle double-digit version numbers well? For instance, will it order them like:

```v11, v10, v9, ... v2, v1``` (numerical)

or 

```v9, v8, ... v11, v10, v1``` (alphabetical) ?

File: ui/package.json [hunk 7]
```diff
@@ -22,6 +22,7 @@
   },
   "scripts": {
     "start": "node scripts/start.js",
+    "dev": "PROXY='http://localhost:3000' node scripts/start.js",
```
andrewhead, line 25: 🎆

ui/src/PaperSummary.tsx
```diff
@@ -2,19 +2,18 @@ import Button from "@material-ui/core/Button";
 import Tooltip from "@material-ui/core/Tooltip";
 import SaveIcon from "@material-ui/icons/Bookmark";
 import CiteIcon from "@material-ui/icons/FormatQuote";
-import React from "react";
 import AuthorList from "./AuthorList";
 import FeedbackButton from "./FeedbackButton";
 import ChartIcon from "./icon/ChartIcon";
 import InfluentialCitationIcon from "./icon/InfluentialCitationIcon";
-import { getRemoteLogger } from "./logging";
+import logger from "./logging";
```
andrewhead, line 9: 👍

andrewhead (commented):
This is looking good so far. I've taken a look up through Line 232 of PaperSummary, and will finish up my review hopefully later this afternoon.

ui/src/ScholarReader.tsx
```diff
@@ -673,7 +628,7 @@ class ScholarReader extends React.PureComponent<Props, State> {
     });
   }
 
-  setFindQuery(findQuery: FindQuery) {
+  setFindQuery = (findQuery: FindQuery): void => {
```
andrewhead, line 631: I'm fine with all these explicit `void` annotations, though also want to ask if there's a reason that you added them in for my personal edification.

api/src/api.ts
mjlangan: We can but until this is actually being used, it's not important. Maybe `version` is plenty fine for now.

andrewhead (commented):
Okay, I got a chance to finish taking a look. Overall, LGTM! Let's hash out the handful of questions about arXiv versions before merging.

api/src/db-connection.ts
mjlangan: Not sure I understand...are you saying that arxiv IDs are not guaranteed to be a series of digits followed by a `v#`?

mjlangan: Ah, I see. The IDs prior to March 2007 throw a wrench in this.

api/src/db-connection.ts
mjlangan: It should, since we're treating the extracted versions as integers.

ui/src/ScholarReader.tsx
mjlangan: It's a practice we follow in the S2 webapp. Makes it crystal clear the intent of this function is not to return anything. More of a code style thing than a functionality thing.

api/src/api.ts
mjlangan: I'm not entirely sure....not that familiar with hapi myself. I might have been cargo-culting here 😅

api/src/db-connection.ts
mjlangan: Sure. Promises can return either a success or a failure, so that's where the "maybe return an error" thought comes from (it also gives the opportunity to return some data about why it failed). Returning null makes it more consistent with the other functions in here. Since this is returning just a `number` type, and that number value has to be 0 or larger, `-1` is also a reasonable value to return when there's no data.

I think it seems fine to return null in this case for consistency, but we might want to look into failing the promise with some kind of error object so the caller has more context about why the request failed.

api/src/api.ts
andrewhead: Sounds good to me! Should we leave in the TODO note or take it out?

mjlangan: I'll remove it.

api/src/api.ts
andrewhead: Hahahah nice! Then I think it should be good to take out. As far as I know, the parameters are already required by being part of the path spec above, and then these additional requirements do additional validation of the data in the params (such as length constraints, etc.)

api/src/db-connection.ts
andrewhead: Agreed! All three of these ideas (-1, null, and error) sound suitable to me. Based on your discussion, I do have a slight preference for throwing an error to keep with conventions of promises (i.e., calling a `reject` callback with details).

You're also right that `getLatestPaperDataVersion` is another async function that returns null. It makes me think maybe we should migrate that method to throw an error at some point too. I can't remember my rationale behind having it return null.

api/src/db-connection.ts
andrewhead: oh yes! you're absolutely right.

api/src/db-connection.ts
mjlangan: I'll go with returning `null` to keep it consistent with the rest of the file, and we can revisit with thrown errors in a subsequent change.

api/src/db-connection.ts
mjlangan: Going to punt on those older ones for now. @rreas says he doesn't plan on driving any papers that old through the pipeline yet. I think a better approach is for the pipeline to store this version as a separate column in the DB so the queries don't need to understand how to parse arxiv ID formats and can simply query integer values & sort them.

andrewhead (approved):
LGTM!

PR #165, opened 2020-12-10
Upload symbol definitions so they can be seen in ScholarPhi interface

andrewhead:
Recently, Dongyeop and Vivek updated the definition extractor to have higher recall and precision. Since then, the data format expected by the UI got out-of-step with the data uploaded from the definition extractor.

This PR includes changes that upload definitions extracted for both tems and symbols to the database in a format that the ScholarPhi UI expects. As a result, you can see that for papers like `1601.00978`, symbol nicknames like `x` are clickable, their definitions show up, and declutter is supported for terms detected by the definition detector.

image

@rreas and @dykang I would appreciate if you two take a look! I realize the code is tangled, though it's probably a good thing to have others on the team familiar with this code as we approach alpha. Of particular note is the file `entities/definitions/upload.py`, where the biggest logical changes were made.

Smaller changes include:

* renaming `acronym*` variables to `abbreviation*` in definition detector
* moving the 'Context' type into the `common.types` instead of `entities.sentences.types`

### DIFF
--- a/data-processing/common/upload_entities.py
+++ b/data-processing/common/upload_entities.py
@@ -209,6 +209,13 @@ def make_data_models(
                 type_ = "float"
             elif isinstance(v, str):
                 type_ = "string"
+            else:
+                logging.debug(  # pylint: disable=logging-not-lazy
+                    "When create a row of entity data, a primitive type could not be "
+                    + "determined for a value with the key '%s'. Make sure that all "
+                    + "entity data is either a primitive type or a list of primitive types. "
+                    + "No row will be created for this data value."
+                )
 
             if type_ is not None:
                 if entity_model is not None:
--- a/data-processing/common/upload_entities.py
+++ b/data-processing/common/upload_entities.py
@@ -109,29 +109,31 @@ def upload_entities(
     # should happen after the entity models are created, because the relationships may include
     # references between entities (e.g., some symbols may be children of others), and we need to
     # know the row IDs of the uploaded entities to make links between them in the database.
-    uploaded_entities = fetch_existing_entities(s2_id, data_version)
+    row_ids = fetch_entity_row_ids(s2_id, data_version)
 
     entity_relationship_models: List[EntityDataModel] = []
     for entity in entities:
         if entity.relationships is not None:
             entity_relationship_models.extend(
                 make_relationship_models(
-                    (entity.type_, entity.id_), entity.relationships, uploaded_entities
+                    (entity.type_, entity.id_), entity.relationships, row_ids
                 )
             )
 
-    EntityDataModel.bulk_create(entity_relationship_models, 200)
+    with output_database.atomic():
+        EntityDataModel.bulk_create(entity_relationship_models, 200)
 
 
 EntityType = str
 WithinPaperId = str
 EntityIdentifier = Tuple[EntityType, WithinPaperId]
-EntityModels = Dict[EntityIdentifier, Entity]
+RowId = str
+EntityRowIds = Dict[EntityIdentifier, RowId]
 
 
-def fetch_existing_entities(
+def fetch_entity_row_ids(
     s2_id: str, data_version: Optional[int] = None
-) -> EntityModels:
+) -> EntityRowIds:
     """
     Build a map from the within-paper entity IDs to database row IDs. It is assumed that the
     number of entities already uploaded for this paper won't be so many that they can't all
@@ -141,14 +143,16 @@ def fetch_existing_entities(
     if data_version is None:
         data_version = get_or_create_data_version(s2_id)
 
-    entity_models: EntityModels = {}
-    uploaded_entities = Entity.select().where(
-        Entity.paper_id == s2_id, Entity.version == data_version
+    row_ids: EntityRowIds = {}
+    rows = (
+        Entity.select(Entity.type, Entity.within_paper_id, Entity.id)
+        .where(Entity.paper_id == s2_id, Entity.version == data_version)
+        .dicts()
     )
-    for entity_model in uploaded_entities:
-        entity_models[(entity_model.type, entity_model.within_paper_id)] = entity_model
+    for row in rows:
+        row_ids[(row["type"], row["within_paper_id"])] = row["id"]
 
-    return entity_models
+    return row_ids
 
 
 def make_data_models(
@@ -247,12 +251,12 @@ def make_data_models(
 def make_relationship_models(
     entity_identifier: EntityIdentifier,
     relationships: EntityRelationships,
-    entity_models: EntityModels,
+    row_ids: EntityRowIds,
 ) -> List[EntityDataModel]:
-    def resolve_model(entity_identifier: EntityIdentifier) -> Optional[Entity]:
+    def resolve_model(entity_identifier: EntityIdentifier) -> Optional[RowId]:
         " Helper for resolving an entity ID into a database row ID. "
         try:
-            return entity_models[entity_identifier]
+            return row_ids[entity_identifier]
         except KeyError:
             type_, within_paper_id = entity_identifier
             logging.warning(  # pylint: disable=logging-not-lazy
@@ -270,19 +274,19 @@ def make_relationship_models(
     models: List[EntityDataModel] = []
 
     # Get the model for the entity for which relationships will be saved.
-    entity_model = resolve_model(entity_identifier)
-    if entity_model is None:
+    row_id = resolve_model(entity_identifier)
+    if row_id is None:
         return []
 
     for k, v in relationships.items():
         if isinstance(v, EntityReference) and v.id_ is not None:
-            referenced_model = resolve_model((v.type_, v.id_))
-            if referenced_model is not None:
+            referenced_row_id = resolve_model((v.type_, v.id_))
+            if referenced_row_id is not None:
                 models.append(
                     EntityDataModel(
-                        entity_id=entity_model.id,
+                        entity_id=row_id,
                         key=k,
-                        value=referenced_model.id,
+                        value=referenced_row_id,
                         item_type="relation-id",
                         of_list=False,
                         relation_type=v.type_,
@@ -292,13 +296,13 @@ def make_relationship_models(
             for reference in v:
                 if reference.id_ is None:
                     continue
-                referenced_model = resolve_model((reference.type_, reference.id_))
-                if referenced_model is not None:
+                referenced_row_id = resolve_model((reference.type_, reference.id_))
+                if referenced_row_id is not None:
                     models.append(
                         EntityDataModel(
-                            entity_id=entity_model,
+                            entity_id=row_id,
                             key=k,
-                            value=referenced_model.id,
+                            value=referenced_row_id,
                             item_type="relation-id",
                             of_list=True,
                             relation_type=reference.type_,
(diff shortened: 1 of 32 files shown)


data-processing/common/upload_entities.py
```diff
@@ -155,73 +109,193 @@ def upload_entities(
...
+    entity_relationship_models: List[EntityDataModel] = []
+    for entity in entities:
+        if entity.relationships is not None:
+            entity_relationship_models.extend(
+                make_relationship_models(
+                    (entity.type_, entity.id_), entity.relationships, uploaded_entities
+                )
+            )
+
+    EntityDataModel.bulk_create(entity_relationship_models, 200)
```
rreas, line 123: I think that what's happening here is that we are adding the entitydata.entity_id field more or less and using these relations to create the necessary joins etc. I'm starting to think that we're missing a layer between the pipeline functional code (i.e., take this input data/files and produce that output data/files) and the model layer (the simple data classes). The in-between layer could look roughly like an ORM but persisted in memory until ready to ship to the database.

Something to think about longer-term... IMO, a "command" should really just be glue code that glues together bits of the underlying classes / functions. What is here now feels like a bit more than that :D Especially in terms of connecting relationships / fetching data for exiting items.

data-processing/common/upload_entities.py
```diff
@@ -155,73 +109,193 @@ def upload_entities(
...
+            # will pass the check 'isinstance(v, int)'.
+            if isinstance(v, bool):
+                type_ = "boolean"
+                v = 1 if v else 0
+            if isinstance(v, int):
+                type_ = "integer"
+            elif isinstance(v, float):
+                type_ = "float"
+            elif isinstance(v, str):
+                type_ = "string"
```
rreas, line 211: Missing some sort of else condition if the type is not one of these? I see that you check None-ness below, so perhaps a DEBUG log (or INFO-level) of some sort? DEBUG makes sense to me b/c the assumption here is that wherever we've generated these values/types can only provide these, however, in the long-term it's easy to forget that constraint!

data-processing/common/upload_entities.py
```diff
@@ -155,73 +109,193 @@ def upload_entities(
...
+
+        # Create a new row for each value, with information of the base data type
+        # and whether that row belongs to a list. If casting of values needs to occur
+        # for values based on type to make them appropriate for insertion in Postgres
+        # (e.g., casting booleans to 0 / 1), that should happen here.
+        for v in values:
+            type_ = None
+            # Check for boolean needs to come before check for int, because booleans
+            # will pass the check 'isinstance(v, int)'.
+            if isinstance(v, bool):
```
rreas, line 203: This all seems weird to me, but I don't understand enough about the end-to-end or API expected data format to know why it's like this! Why do we need item type?

rreas (approved):
I'm not sure I can provide much of value. I've tried to leave a few forward-looking notes! I don't expect that I'd be much help in catching any real bugs without helping run the end-to-end (which we can certainly do!). 

In general, once we start thinking about something like a V2 of this code we can think about decoupling the on-disk data formats and the "commands" from the data model and a storage layer and starting to pull apart the pieces to make it more testable.

andrewhead:
> I'm not sure I can provide much of value. I've tried to leave a few forward-looking notes! I don't expect that I'd be much help in catching any real bugs without helping run the end-to-end (which we can certainly do!).

Thanks Russell, this is immensely helpful!

data-processing/common/upload_entities.py
andrewhead: >The in-between layer could look roughly like an ORM but persisted in memory until ready to ship to the database.

This is a great idea. I'm fully behind it. As you noted, one of the clunky parts of the code as it stands is that many of these entities need to reference each other, though there will be thousands of rows to upload so it doesn't make sense to upload them all individually and fetch their row IDs whenever a foreign key relation needs to be established. I like this idea of a layer like the one you mentioned that takes care of establishing those relationships and performing the bulk upload once a paper is finished processing.

andrewhead: >Something to think about longer-term... IMO, a "command" should really just be glue code that glues together bits of the underlying classes / functions.

100% agreed. The current command structure has ballooned in complexity beyond my initial expectations 🎈 😬 . I'd love to discuss further what the ideal structure looks like, and think about whether there's small ways we can remove complexity in future PRs as we make other changes before we're ready for a large-scale refactor.

data-processing/common/upload_entities.py
andrewhead: Good thinking! I'll add this statement.

data-processing/common/upload_entities.py
rreas: I think related to a previous conversation with Kyle, one thing that could be a step forward (possibly a bit large) would be some sort of entity DAO that does the following. Suppose I am running command Z that needs the output of command X and Y. Due to the pipeline wiring I've run X locally by Y only exists in the DB. The entity DAO makes this transparent to the command.

1. Check local disk for entities if there then yield to Z
2. Check DB for entities if there then yield to Z and probably also keep locally for other commands
3. Else, fail... Some sort of wiring or ordering is not quite right
4. OR instead of failing, invoke the necessary command (this seems like beyond the scope of the DAO)

You've modeled commands as a DAG, and I think that was the right call. Now, it's a question of making each command live in isolation.

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -866,10 +872,17 @@ def process(
                     t.tex_path,
                     item.arxiv_id,
                 )
+                type_ = (
+                    "abbreviation"
+                    if t.text in abbreviations
+                    else "term"
+                    if t.text in term_phrases
+                    else "unknown"
+                )
```
dykang, line 881: Okay, now you combined terms and abbreviations but having a type in it. I wonder how you combine these with nicknames though.

data-processing/entities/definitions/types.py
```diff
@@ -103,7 +103,6 @@ class Definition(SerializableEntity):
     """
 
     text: str
-    sentence_id: str
```
dykang, line 106: Do you have a particular reason that you removed the `sentence_id` field? I thought that it might be useful in the future to extract the relative position of the sentence in a paper, etc.

data-processing/entities/definitions/upload.py
```diff
@@ -1,86 +1,265 @@
...
     processing_summary: PaperProcessingResult, data_version: Optional[int]
 ) -> None:
+    upload_term_definitions(processing_summary, data_version)
+    upload_symbol_definitions(processing_summary, data_version)
 
+
+TermName = str
+
+
+def is_textual_term(entity: SerializableEntity) -> bool:
```
dykang, line 28: Same question with the previous one: Do we treat terms and definiendums different from symbols?

data-processing/entities/definitions/upload.py
```diff
@@ -1,86 +1,265 @@
...
+                    else None,
+                ),
+                "definition_sentences": [
+                    EntityReference(type_="sentence", id_=id_)
+                    for id_ in definition_sentences
+                ],
+                "snippet_sentences": [
+                    EntityReference(type_="sentence", id_=id_)
+                    for id_ in snippet_sentences
+                ],
```
dykang, line 117: It would be great if you can add some description of each relationship type here. Also, how are they used differently in UI side?

dykang:
I particularly took a look at codes related to definitions and sentences. Overall, it looks good to me. I put some clarification questions when I don't understand the logic behind them by only looking at code.

data-processing/common/upload_entities.py
andrewhead: Good question! It's a bit convoluted:

At one point in time, I thought we might want to add arbitrary new data fields of any primitive type for entities, and then have them served up to the API without having to make changes to the API code.

Therefore, the API currently has code that scans through all entitydata rows for each entity, then uses the type listed in the `item_type` column to cast that item to the right type (e.g., a boolean, an integer, a string) before packaging that data up into a JSON response to send to the UI.

This design was to make it faster to prototype the UI as we discovered we needed to expose more information about the entities. I'm not sure it makes sense in the long term for the API, as I imagine we would the API to only expose certain data fields for each entity, rather than any type of data found in the entitydata table. Does that all make sense?

data-processing/entities/definitions/commands/detect_definitions.py
andrewhead: Hey @dykang ! It's a good question. Right now, we don't extract term references for symbols, because term references for symbols are already extracted elsewhere in the pipeline (namely, the `extract_symbols.py` command).

Then, the `upload.py` file for definitions loops through all `Definiendum`s (i.e., the instances of symbols that have nicknames), scans the extracted symbol data for all other appearances of those symbols, and then uploads definition information for each of those symbol appearances. Does that make sense? Might there be a clearer way to implement behavior in the code?

data-processing/entities/definitions/upload.py
andrewhead: yup, we sure do! As mentioned above, we don't have to detect appearnaces of symbols in this step of the pipeline, as they're already detected in an earlier stage of the pipeline. So this function here is used to filter out symbols, so that only appearances of terms and abbreviations are located.

Would it be helpful if I added a comment about this to the code?

data-processing/entities/definitions/upload.py
andrewhead: Sure thing! I will do this shortly.

`definition_sentences` are IDs of the sentences that contain each of the definitions for a term. They're use to populate the links that take you to the passage where a term is defined.

`snippets` is the list of all other appearances of the term everywhere else in the paper. A `snippet` contains the full text of a sentence where that term appears. `snippet_sentences` is to 'snippets' what 'definition_sentences' is to 'definitions'; specifically, it contains the IDs of each sentence where the term appears elsewhere in the paper, so that we can link from a list of usages of the term to the passage in the paper that contains it.

andrewhead:
Thank you, @dykang! I've added clarifications in my responses, and will integrate a few of them into comments where you asked for them. I'm curious to know if they answer your questions!

data-processing/entities/definitions/commands/detect_definitions.py
dykang: Thanks for the clarification. I think this makes sense to me!

data-processing/entities/definitions/upload.py
dykang: Yeah, it would be helpful to make a short note about what you described here!

data-processing/entities/definitions/upload.py
dykang: I see. This is clear now. Would it be possible to add a short description of each relationship type, as you described here?

dykang:
@andrewhead Yes, they all answer my questions clearly! Please let me know once you resolve the rest of the other issues!

data-processing/entities/definitions/types.py
andrewhead: Okay! I'll add it back in 👍  It wasn't currently being used in the pipeline, though I was too hasty in deleting it.

data-processing/entities/definitions/upload.py
andrewhead: Done! This clarification will be added as a comment into the body of the `upload_definitions` function right before the calls to both `upload_term_definitions` and `upload_symbol_definitions`

data-processing/entities/definitions/upload.py
andrewhead: Sure thing! Done!

PR #166, opened 2020-12-10
Fix the paper list.

codeviking:
A long time ago in a galaxy far, far away...I added a list of papers
that were interesting to look at in ScholarPhi.

This PR fixes that list, in hopes that it might help with things
like https://github.com/allenai/scholar/issues/25368.

The fix required revising a database query and removing a whitelist,
which is probably ok to remove since (I think) we're beyond whatever
demo window it was added for.

### DIFF
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -78,9 +78,6 @@ export class Connection {
   }
 
   async getAllPapers() {
-    // The doubly-nested subqueries below are probably kind of slow, and might be a lot faster
-    // via some joins and clever SQL acrobatics. I leave this for a future DB guru. For now
-    // the sub-queries are (at least to me) easier to read.
     const response = await this._knex.raw<{ rows: PaperWithEntityCounts[] }>(`
         WITH version AS (SELECT MAX(version) AS latest FROM entity)
       SELECT paper.*,
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -82,13 +82,15 @@ export class Connection {
       SELECT paper.*,
              (
                 SELECT COUNT(*)
-                  FROM citation
-                 WHERE citation.paper_id = paper.s2_id
+                  FROM entity
+                 WHERE entity.paper_id = paper.s2_id
+                   AND type = 'citation'
              ) AS citations,
              (
                 SELECT COUNT(*)
-                  FROM symbol
-                 WHERE symbol.paper_id = paper.s2_id
+                  FROM entity
+                 WHERE entity.paper_id = paper.s2_id
+                   AND type = 'symbol'
              ) AS symbols
         FROM paper
     ORDER BY symbols DESC, citations DESC
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -78,6 +78,9 @@ export class Connection {
   }
 
   async getAllPapers() {
+    // The doubly-nested subqueries below are probably kind of slow, and might be a lot faster
+    // via some joins and clever SQL acrobatics. I leave this for a future DB guru. For now
+    // the sub-queries are (at least to me) easier to read.
     const response = await this._knex.raw<{ rows: PaperWithEntityCounts[] }>(`
       SELECT paper.*,
              (
@@ -85,12 +88,24 @@ export class Connection {
                   FROM entity
                  WHERE entity.paper_id = paper.s2_id
                    AND type = 'citation'
+                   AND version = (
+                        SELECT MAX(version)
+                          FROM entity
+                         WHERE type = 'citation'
+                           AND entity.paper_id = paper.s2_id
+                   )
              ) AS citations,
              (
                 SELECT COUNT(*)
                   FROM entity
                  WHERE entity.paper_id = paper.s2_id
                    AND type = 'symbol'
+                   AND version = (
+                        SELECT MAX(version)
+                          FROM entity
+                         WHERE type = 'symbol'
+                           AND entity.paper_id = paper.s2_id
+                   )
              ) AS symbols
         FROM paper
     ORDER BY symbols DESC, citations DESC
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -82,30 +82,21 @@ export class Connection {
     // via some joins and clever SQL acrobatics. I leave this for a future DB guru. For now
     // the sub-queries are (at least to me) easier to read.
     const response = await this._knex.raw<{ rows: PaperWithEntityCounts[] }>(`
+        WITH version AS (SELECT MAX(version) AS latest FROM entity)
       SELECT paper.*,
              (
                 SELECT COUNT(*)
                   FROM entity
                  WHERE entity.paper_id = paper.s2_id
                    AND type = 'citation'
-                   AND version = (
-                        SELECT MAX(version)
-                          FROM entity
-                         WHERE type = 'citation'
-                           AND entity.paper_id = paper.s2_id
-                   )
+                   AND version = (SELECT latest FROM version)
              ) AS citations,
              (
                 SELECT COUNT(*)
                   FROM entity
                  WHERE entity.paper_id = paper.s2_id
                    AND type = 'symbol'
-                   AND version = (
-                        SELECT MAX(version)
-                          FROM entity
-                         WHERE type = 'symbol'
-                           AND entity.paper_id = paper.s2_id
-                   )
+                   AND version = (SELECT latest FROM version)
              ) AS symbols
         FROM paper
     ORDER BY symbols DESC, citations DESC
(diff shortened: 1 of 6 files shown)


ui/src/papers/index.tsx
```diff
@@ -162,17 +163,7 @@ const PaperList = () => {
...
-                    [
-                      "1508.07252",
-                      "1706.03762",
-                      "1811.12160",
-                      "1010.2823",
-                      "1309.7440",
-                      "1404.6635",
-                    ].indexOf(paper.arxivId) !== -1
-                  );
+                  return paper.arxivId !== undefined;
```
mjlangan, line 166:
```suggestion
                  return !!paper.arxivId;
```
If we just want to make sure that the arxivId is truthy, this would be better than explicitly checking that it isn't `undefined`.

codeviking: Gotcha. I go either way on this one.  I tend to avoid value coercion where I can, because it can be tricky and cause unexpected side-effects.  In this case we know that it'll be `undefined` or a non-empty string, so I like the explicitness of checking for the case that I know we don't want.

That said by the same argument I could say that there's no risk in the value coercion -- so let's go with `!!` because it's shorter (and I imagine more in line with S2's JavaScript styles!).

api/src/db-connection.ts
```diff
@@ -82,13 +82,15 @@ export class Connection {
...
+                 WHERE entity.paper_id = paper.s2_id
+                   AND type = 'citation'
              ) AS citations,
              (
                 SELECT COUNT(*)
-                  FROM symbol
-                 WHERE symbol.paper_id = paper.s2_id
+                  FROM entity
+                 WHERE entity.paper_id = paper.s2_id
+                   AND type = 'symbol'
```
rreas, line 93: I think these queries should take into account the 'version' table. The numbers may be inflated as a single paper may have many versions with 1000s or 10000s of entities. Specifically there is an `entity.version` column and the version table itself is indexed on paper_id. Easiest thing for now is probably aggregating to max(version.version) to find the latest. Happy to address this as a follow-up.

rreas:
Big help, thanks!

andrewhead (commented):
Looks good! I echo Russell's comment regarding versions. Do we want to incorporate that into this PR, or a later one?

api/src/db-connection.ts
codeviking: Sure, I can take care of that before merging.

codeviking:
> Looks good! I echo Russell's comment regarding versions. Do we want to incorporate that into this PR, or a later one?

I'll take care of it before merging.

codeviking:
@andrewhead @rreas Ok, I think I incorporated the necessary changes. Please TAL.

Here's before:

Screen Shot 2020-12-10 at 6 01 36 PM

...and after:

Screen Shot 2020-12-10 at 6 01 19 PM

The numbers look a lot more reasonable, though 3k symbols is still a lot!

api/src/db-connection.ts
codeviking: I took a stab at it -- but now that I'm reviewing it I wonder if it can even simpler.

I think y'all might be saying that there's only one `version` that I need to use (which is the most recent) for all entity types. So I can use one query to find that, and filter the counts based on that.

Let me know if that's the case I can tweak this to reflect that.

codeviking:
Ok, take 3.

I think I have the query correct now...and it looks like there's only one paper with symbols and citations of the latest version.

To prevent this paper from feeling less lonely I removed the bit that filters for papers who have more than 1 thing. Sure, we don't really want to show those, but at least we know what papers *have* been processed. We show the ones with data first -- so I don't think it's really too big of a deal to show the ones that don't have anything.

Here's how it looks:
Screen Shot 2020-12-10 at 8 21 04 PM

Please TAL and the query and lmk if this is right.

api/src/db-connection.ts
codeviking: Ok, this got another tweak, I think it's right now.

rreas: Yes, that looks right to me. My comment above wasn't quite right as the "version" column referenced in "entity" is actually "index" in the version table. What you've done here sidesteps the need to know that 👍 

I verified against the public schema that there are entities for papers I expect to have them (i.e., not all are 0).

rreas (approved):

codeviking:
Ok, I mucked around with the query a bit more, thanks to help from @rreas. I also:

- Added pagination
- Added counts for each entity type
- Sorting by the total # of all entities
- Added a cache-busting mechanism for the UI cache (I debated removing the cache entirely, but decided not to).

Here's how it looks:

Screen Shot 2020-12-11 at 1 47 04 PM

api/src/db-connection.ts
codeviking: I took _another_ stab at the query, with your help (it's more or less what you shared with me in Slack).

TAL.

api/src/db-connection.ts
```diff
@@ -77,30 +78,61 @@ export class Connection {
...
+    const response = await this._knex.raw<{ rows: Row[] }>(`
+      SELECT paper.arxiv_id,
+             paper.s2_id,
+             version.index AS version,
+             SUM(CASE WHEN entity.type = 'symbol' THEN 1 ELSE 0 END) AS symbol_count,
+             SUM(CASE WHEN entity.type = 'citation' THEN 1 ELSE 0 END) AS citation_count,
+             SUM(CASE WHEN entity.type = 'sentence' THEN 1 ELSE 0 END) AS sentence_count,
+             SUM(CASE WHEN entity.type = 'term' THEN 1 ELSE 0 END) AS term_count,
+             SUM(CASE WHEN entity.type = 'equation' THEN 1 ELSE 0 END) AS equation_count,
+             SUM(CASE WHEN entity.type = 'definition' THEN 1 ELSE 0 END) AS definition_count,
```
andrewhead, line 103: fyi: the way definitions are stored in the database is about to change with my next pipeline update. They will be sorted as rows in the `entitydata` table, where each row is with a 'term' or 'symbol' entity from the `entity` table.

Maybe it's worth removing this column for now? Or, if we want to include it anyway, we can put together a new query to get definition counts once I've incorporated the pipeline updates.

ui/src/papers/index.tsx
```diff
@@ -69,6 +73,12 @@ interface CachedPaperWithMeta {
   paper: PaperWithMeta;
 }
 
+function cacheKey(paperId: string) {
+    // We append a version prefix so that we can bust the cache in clients when we release
+    // updates.
+    return `v0/${paperId}`;
```
andrewhead, line 79: Just to check my understanding, does this mean when we release updates, we would bump the version number for the full API (say to `v0.1`) in order to bust the cache?

ui/src/papers/index.tsx
```diff
@@ -156,65 +216,92 @@ const PaperList = () => {
...
+                      </strong>
+                    </Typography>
+                    {paper.authors.map((author, idx) => (
+                      <React.Fragment key={author.url}>
+                        <a href={author.url}>{author.name}</a>
+                        {idx !== paper.authors.length - 1 ? ", " : null}
+                      </React.Fragment>
+                    ))}
+                    {paper.venue && ` • ${paper.venue}`}
+                    {" •"} {paper.year}
```
andrewhead, line 248: Is the paper.year guaranteed to not be null in the S2 API?

andrewhead (commented):
Thanks for this, Sam! This looks good. I added a few questions about particulars in the code.

api/src/db-connection.ts
codeviking: Sure. I have some work on it's way that'll fit nicely with this. I'll remove the column in that PR.

ui/src/papers/index.tsx
codeviking: Yup, exactly. It's only necessary where there's a backwards incompatible change -- say, if the format of the data changes (like it did in this PR).

ui/src/papers/index.tsx
codeviking: I'm not sure. I don't think I changed this code, rather I just moved it. If It's `undefined` we'll just end up with a stray bullet -- a little unideal, but I *think* this page is still for internal use, so it's probably ok?  Though to your point -- when this gets baked into the UI we'll need to handle those edge cases carefully!

rreas: For now, since we are focused on ArXiv papers, this value will always be available unless something has gone very, very wrong.

andrewhead: Good point! Okay, I'm convinced this is okay to keep as is.

api/src/db-connection.ts
andrewhead: sounds good, thank you!

PR #167, opened 2020-12-11
Run tests using a local database.

codeviking:
This does several things:

1. Adds the `./api/bin/start_local_db.sh` script for starting a local
   database. It uses `docker` to run things in a container, and uses
   the same Postgres version that we do now.

2. Modifies the tests to use that database.

3. Adds a CI configuration that uses GitHub actions to run the API
   tests on PRs and do some light validation of the UI.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -75,4 +75,3 @@ After you're done you can stop the database like so:
 docker stop scholarphi-db
 ```
 
-
--- a/api/README.md
+++ b/api/README.md
@@ -50,3 +50,29 @@ contents:
 
 Once you have, you can relaunch the server, and it should be 
 capable of querying for data from the database.
+
+## Running the Tests
+
+The tests use a local database that's ran in in [docker](https://www.docker.com/).
+If you don't have `docker` installed, you'll need to install it to
+run the tests.
+
+First, start the local database:
+
+```
+./bin/start_local_db.sh
+```
+
+Then run the tests:
+
+```
+npm test
+```
+
+After you're done you can stop the database like so:
+
+```
+docker stop scholarphi-db
+```
+
+
--- a/api/bin/start_local_db.sh
+++ b/api/bin/start_local_db.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+set -eu
+
+POSTGRES_USER="${POSTGRES_USER:-postgres}"
+POSTGRES_PASSWORD="${POSTGRES_PASSWORD:-pdfsarefun}"
+POSTGRES_DB="${POSTGRES_DB:-scholar-reader}"
+LOCAL_DB_PORT="${LOCAL_DB_PORT:-5555}"
+
+IMAGE="postgres:11.8"
+CONTAINER_NAME="scholarphi-db"
+
+is_running=$(docker ps -q --filter name="$CONTAINER_NAME")
+if [[ ! -z "$is_running" ]]; then
+    echo "The database is already running. For more information, run:"
+    echo "    docker ps --filter name=$CONTAINER_NAME"
+    exit 0
+fi
+
+docker run -d --rm --name "$CONTAINER_NAME" \
+           -p "$LOCAL_DB_PORT:5432"  \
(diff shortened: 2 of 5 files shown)


codeviking:
Sorry for the noise. I pinged y'all a bit early. There still might be some turbulence while I get CI working. I'll re-assign when it's ready for review.

codeviking:
Ok, CI is working now. Take a look!

mjlangan (approved):
LGTM! Surprised at how intuitive the workflows config is to someone who hasn't used it before 👍

api/bin/start_local_db.sh
```diff
@@ -0,0 +1,53 @@
...
+    docker exec "$CONTAINER_NAME" pg_isready \
+                                  -h localhost -p 5432 \
+                                  -U "$POSTGRES_USER" 2>&1 > /dev/null
+    is_db_ready=$?
+}
+while [[ $is_db_ready -ne 0 ]]; do
+    echo "Waiting for the database to start..."
+    sleep 1
+    check_if_db_is_ready
+done
```
rreas, line 42: Want to add some sort of max attempts here to avoid CI stuck in loop? I'm fine skipping that if a top-level timeout can be set for the CI step.

api/src/tests/api.test.ts
```diff
@@ -164,7 +164,8 @@ describe("API", () => {
       } as Entity);
     });
 
-    test("symbol", async () => {
+    // Skipped by @codeviking on 12/10/2020.
```
rreas, line 167: Would be nice to have a one-liner as to why :)

rreas (approved):
Thanks a ton for this!

api/bin/start_local_db.sh
codeviking: Good call. I think GitHub Actions have a timeout -- but let me confirm.

api/src/tests/api.test.ts
codeviking: Yea, I agree --  I myself don't know why.

@andrewhead Do you have details I could add here? I could also add:

```
// Skipped by @codeviking on 12/10/2002 -- appears to have been a long-standing regression, so I'm not sure why
// it stopped working.
```

api/bin/start_local_db.sh
codeviking: The default timeout (for the entire job) is 6 hours.  I'll set one that's more aggressive, as that's a *bit* more time than we need.

api/src/tests/api.test.ts
codeviking: I added a comment noting explaining why it was skipped (to the best of my ability).

api/README.md
```diff
@@ -50,3 +50,28 @@ contents:
 
 Once you have, you can relaunch the server, and it should be 
 capable of querying for data from the database.
+
+## Running the Tests
+
+The tests use a local database that's ran in in [docker](https://www.docker.com/).
```
andrewhead, line 56: nit pick: "ran in in" -> "run in"

api/README.md
```diff
@@ -50,3 +50,28 @@ contents:
...
+If you don't have `docker` installed, you'll need to install it to
+run the tests.
+
+First, start the local database:
+
+```
+./bin/start_local_db.sh
```
andrewhead, line 63: Is this necessary to also run the API locally? If so, maybe we should add a note about that up above in the API launch instructions on line 15 of the README

andrewhead (commented):
Thanks for this, Sam! This looks good. I added a couple of notes for small cleanups that could be done for the README.

api/README.md
codeviking: Nope. The tests start an instance of the API, which connects to this database.

api/README.md
codeviking: Oops, thanks!

api/README.md
andrewhead: D'oh! I'm not sure what I was thinking. Thanks for the clarification!

PR #169, opened 2020-12-14
Some dependency housekeeping

mjlangan:
Key points:
- `tslint` was deprecated at the beginning of last year and we aren't using it for linting, so I'm pulling it out of dependencies. ESLint has a TypeScript plugin that is now the official (or at least de facto) linter and I can set that up in a separate change.
- `lodash` also removed from the dependencies as we aren't using it and don't really need it.
- Updated axios, pg, nconf, and knex too.
- Just a little housekeeping thing, updated the `license` field in the api package.json to match what we have set on the repo. Not that we're publishing it to NPM, but might as well change it 🤷


andrewhead (approved):
Great, thank you!

PR #170, opened 2020-12-14
Query S2's API using an API key if we have one + some other cleanup.

codeviking:
This PR started as an effort to add search, but as discussed I ended
up pausing on that change in favor of eventually using S2's regular
search API. This PR instead captures changes that I made while
implementing search that seem worth preserving.

The largest functional difference is that the API now uses an API
key when querying S2's public API. This means we can safely query
for this information in the backend when preparing papers to return
for the `/papers/list` API endpoint.

To do this I did some cleanup of the way configuration parameters
are handled so that they can be accessed in a typesafe fashion.

Lastly I cleaned up the paper list UI a bit, as it was getting
a little unruly.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -1,9 +1,9 @@
 # Scholar Reader API
 
-The API server for the Scholar Reader application. For a 
-given paper, the application needs to know about the 
-citations, symbols, and other entities in that paper and 
-where they are located. This server provides that data upon 
+The API server for the Scholar Reader application. For a
+given paper, the application needs to know about the
+citations, symbols, and other entities in that paper and
+where they are located. This server provides that data upon
 request.
 
 To run the server, first install the dependencies:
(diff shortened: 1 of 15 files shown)


api/README.md
```diff
@@ -26,36 +26,38 @@ curl 0.0.0.0:3000/api/health
...
+  the entities we extract from them is stored.
+- Query [Semantic Scholar's Public API](https://api.semanticscholar.org/).
+
+The API queries work without an API key, but will get rate
+limited if you make more than 100 queries in 5 minutes.
+This can happen relatively quickly during normal use of the
+application.
+
+The database won't workout without the appropriate
+configuration.
```
mjlangan, line 47: Nit: What is the appropriate configuration? An example `secrets.json` file would be a useful reference so folks know what keys to use when plugging in the values without having to look at the TS interface.

api/src/conf.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * This module provides type information for the bits of configuration the API needs.
+ * We use classes instead of interfaces to enforce the immutability of the values.
```
mjlangan, line 3: 👍

api/src/conf.ts
```diff
@@ -0,0 +1,47 @@
...
+export class Config {
+    constructor(
+        readonly db: DBConfig,
+        readonly s2: S2APIConfig
+    ) {}
+
+    static fromConfig(conf: nconf.Provider) {
+        const db = conf.get("database");
+        const s2 = conf.get("s2"); 
+        const algolia = conf.get("algolia");
```
mjlangan, line 41: leftover bit of that search stuff

api/README.md
codeviking: Sure, can do.

api/src/conf.ts
codeviking: Oops, thanks!

api/README.md
codeviking: Done, TAL.

api/src/types/api.ts
```diff
@@ -48,6 +47,17 @@ export interface Paper {
   citationVelocity?: number;
 }
 
+export interface PaperWithEntityCounts extends PaperIdWithEntityCounts {
```
mjlangan, line 50: I feel like there's a better way to name this/these...but 🤷

api/src/api.ts
```diff
@@ -16,36 +37,48 @@ export const plugin = {
...
-            Array.isArray(request.query.offset)
-              ? request.query.offset[0]
-              : request.query.offset
-          );
-          if (!isNaN(o)) {
-            offset = o;
+        const offset = firstIntOrDefault(request, "offset", 0);
+        const size = firstIntOrDefault(request, "size", 25);
+
+        const papers = await dbConnection.getAllPapers(offset, size);
```
mjlangan, line 49: do we need any try/catch-ing around this?

api/src/api.ts
```diff
@@ -16,36 +37,48 @@ export const plugin = {
...
+        const paperIds = papers.rows.map(p => p.s2_id);
+
+        // We fetch metadata about each paper (it's title, author names, etc) from S2's public
+        // API and merged them into the result set.
+        const s2PaperInfoByPaperId: { [pid: string]: Paper } = {};
+        for (const s2Paper of await s2Api.getPapers(paperIds, config.s2.apiKey, true)) {
```
mjlangan, line 55: same here

ui/src/papers/index.tsx
```diff
@@ -23,178 +23,101 @@ import TableRow from "@material-ui/core/TableRow";
...
+  const s = parseInt(u.searchParams.get("size") || "");
+  if (isNaN(s)) {
+    return defaultSize;
+  }
+  return s;
 }
 
+/**
+ * SAM TO DO:
+ * - Display Algolia logo on search results (or update to paid account)
```
mjlangan, line 90: more search leftovers

mjlangan (approved):
LGTM

ui/src/papers/index.tsx
codeviking: Nice catch, removed.

api/src/api.ts
codeviking: Yup, we should. Good catch. I'll add that.

api/src/api.ts
codeviking: Yes and no.

In this case I don't think we need to, because I meant to set `failOnError` to `false`.  This causes the function we're calling to not fail *everything* if a single request fails.

I'll fix that, and add a `try / catch` above. Thanks!

api/src/api.ts
codeviking: So, in this case we're ok actually (I tested what happens).  An error downstream maps to a 500, but doesn't crash the server.

Elsewhere (in this codebase) I've ran into places where code that `throw`s causes the server to crash. Luckily this case isn't one, so we're ok (I think).

api/src/types/api.ts
codeviking: Yea, totally agree. I think some cleanup of the types is definitely warranted.

PR #171, opened 2020-12-15
Add loading indicator for citation data

mjlangan:
Ref https://github.com/allenai/scholar/issues/25245

I reworked the part of the PrimerPage that displayed the loading indicator for entity/definition data to also indicate whether or not we've loaded citation data.

Preview:

### DIFF
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -37,6 +37,7 @@ class ApiServer {
     const connectionParams = extractConnectionParams(this._config);
     const dbConnection = new Connection(connectionParams);
     this._dbConnection = dbConnection;
+    console.log(`Using database at ${connectionParams.host}`);
 
     /**
      * Register API endpoints.
--- a/ui/src/PrimerPage.tsx
+++ b/ui/src/PrimerPage.tsx
@@ -18,6 +18,7 @@ interface Props {
   annotationHintsEnabled: boolean;
   termGlossesEnabled: boolean;
   scrollToPageOnLoad?: boolean;
+  isCitationsLoading?: boolean;
   handleSetAnnotationHintsEnabled: (enabled: boolean) => void;
 }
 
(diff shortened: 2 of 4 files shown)


api/src/server.ts
```diff
@@ -37,6 +37,7 @@ class ApiServer {
     const connectionParams = extractConnectionParams(this._config);
     const dbConnection = new Connection(connectionParams);
     this._dbConnection = dbConnection;
+    console.log(`Using database at ${connectionParams.host}`);
```
andrewhead, line 40: Would you like this console.log statement to stay in the code?

ui/src/PrimerPage.tsx
```diff
@@ -18,6 +18,7 @@ interface Props {
   annotationHintsEnabled: boolean;
   termGlossesEnabled: boolean;
   scrollToPageOnLoad?: boolean;
+  isCitationsLoading?: boolean;
```
andrewhead, line 21: nit pick: maybe `areCitationsLoading` / `isLoadingCitations`?

andrewhead: Another thought is that below you've got a callback called 'isEntitiesLoaded' so maybe this could be  called 'isCitationsLoaded' (flagging completion rather than progress)

andrewhead (approved):
Looks good!

andrewhead:
By the way, that's a slick GIF. How do you produce your GIFs? I'm surprised to see one that's so high-res and long and still under 1MB.

api/src/server.ts
mjlangan: I figured it might be useful for the app to report which database its connecting to, given we've recently been changing that around.

mjlangan:
@andrewhead I use Recordit -- https://recordit.co/

ui/src/PrimerPage.tsx
mjlangan: the function there is a type guard, so it returns "this thing is of type Entities" (which is how we determine whether or not the entities are loaded). For the boolean, my preference is `true` for loading states and `false` in the (more normal) loaded state.

Issue #135, opened 2020-07-28
Resolve macros in LaTeX files before detecting entities


andrewhead:
cc @dykang @kyleclo 

To aid in definition and symbol detection, macros should be expanded in LaTeX files before they are processed.

I haven't had much luck so far finding a tool yet for expanding macros. Options seem to include:
* `de-macro`: a command line utility that will expand commands defined with \newcommand (though not \def)
* This script (https://tex.stackexchange.com/a/11236/198728) (also only works for \newcommand, I think)
* A home-brewed solution

(source: This TeXExchange question: https://tex.stackexchange.com/questions/3174/latex-macro-expander)

For the home-brewed solution, there are a couple of options.

1. if `\tracingmacros=1` is set at the beginning of a LaTeX file, then LaTeX will output expansions for each of the macros called. During compilation, it might possible to do a first compilation pass to get the macro expansions, and then replace the macros with their expansions before the next pass.

2. support a subset of macros (e.g., those with very simple expansion patterns), and write our own scanner that finds these macros and expands them in the files. I expect this might be the 80 / 20 solution (e.g., the one where we only have to spend 20% of the work to get 80% of the payoff).

PR #173, opened 2020-12-16
[WIP] Kylel/2020 11 10/demacro

kyleclo:
- Some hacky scripts for performing demacro

### DIFF
--- a/data-processing/demacro/README.md
+++ b/data-processing/demacro/README.md
@@ -14,3 +14,64 @@ The steps are:
 - Do we need to point to the "main" `.tex` file?
 - Do we need  
 
+
+### setup experiments
+
+We ssh into a local docker instance w/ all reader pipeline functionality.  This will handle the (1) downloading, (2) unpacking, and (3) compiling operations as needed.
+
+If the docker is already built:
+```python
+# this ssh into it
+sudo docker exec -it blissful_noether /bin/bash
+
+# cd into the right place
+cd /data-processing/
+
+# verify the arxiv IDs already there
+# if not, manually paste a bunch of them there
+ls arxiv_ids.txt
+
+# verify the arxiv IDs have already been downloaded
+# if not, run the download & unzip operations
+ls /data-processing/data/02-sources-archives/
+ls /data-processing/data/04-sources/
+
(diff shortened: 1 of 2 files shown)


andrewhead:
Related to Issue #135

andrewhead:
[WIP comment]

Below, I am characterizing the presence of macros in TeX papers from recent AI / ML proceedings (NeurIPS, ICML, CVPR, ACL) to motivate what approach we need to take to resolve macros in papers.

This table is based on a quick look at papers; they may not be 100% correct, but represent what can be sussed about a paper after about 5 minutes of skimming its sources.


Columns:
* _Has complex commands_: Are there conditionals or other custom macros (i.e., something defined in another newcommand or def in the paper) inside the resolution of a macro? 
* _In .sty file?_: If there are macros or commands, do they appear in a .sty file? (suggests that we will need to inline style files in order to resolve macros)

Paper | Venue | Has \newcommand? | Has \def? | Has \DeclareMathOperator | Has complex macros? | In .sty file?
-|-|-|-|-|-|-
1704.06851 | ACL |  - | - | - | - | -
1805.09016 | ACL | Y | - | Y | - | -
1905.05475 | ACL | - | - | Y | - | -
1705.06566 |
1803.00443 | 
1902.02671 |
1611.09852 |
1801.01615 |
1904.03582 |
1709.07902 |
1804.00792 |
1905.10887 |
1702.01287 |
1805.08660 |
1908.00300 |
1706.08482 |
1903.00621 |
1706.03850 |
1802.07740 |
1811.12359 |
1707.00683 |
1711.08028 |
1906.04604 |

Other tricky cases that might be worth checking for:
* Is a macro or command overwritten somewhere in the file? (I don't think this is likely to happen this often and we can probably ignore it as a rare case)

kyleclo:
**\renewcommand**?

there's also **\newenvironment** and **\renewenvironment** but im not sure that matters as much

andrewhead:
Great point! I added `\renewcommand` to th elist above. Yeah, for `\newenvironment` I'm also not sure. Have you ever seen newenvironmennt or renewenvironment used in a way that would make it impossible to detect terms or symbols using our typical mechanisms? If so, then I would want to find and replace those environments.

andrewhead:
Corner cases to handle:

* When a subscript follows right after the use of a macro
* ...

PR #175, opened 2020-12-16
Add folder structure to UI components

mjlangan:
I took a stab at organizing the various component files into folders. I tried to group them up by related purpose and those used together to make bigger chunks of the UI, and then went back fixing up the imports and finding the various weird dependencies that emerged. A further task is to split out some of the more general-purpose interfaces implemented among the components & organize them, but that's outside the scope of this PR.

I also deleted 6 unused files.

Head on over to the branch to explore the new structure: https://github.com/allenai/scholarphi/tree/mjlangan/folders/ui/src
Also please let me know if I misinterpreted anything or grouped things that shouldn't with how the files were arranged :)

```
/ui/src/
- /api
- /components
  - /common
  - /control
  - /drawer
  - /entity
    - /citation
    - /creation
    - /equation
  - /icon
  - /mask
  - /overlay
  - /pdfjs
  - /preview
  - /primer
```


andrewhead:
This looks lovely @mjlangan.

The current organization is crisp and thoughtful. I don't take issue with any of it. That said, I'll raise a few questions which might suggest other places that some of the components could fit conceptually:

1. Perhaps the `entity/creation` folder should be moved into the `control` folder. I do not anticipate us exposing entity creation tools to users any time soon; they're more of an admin capability when one of us needs to clean up data quality.

2. Same question for `EntityPropertyEditor` and `EntityPropertyField`

3. Would `search` be a more appropriate name for the folder currently named `pdfjs`? All the components in that directory are currently related to search functionality. Alternatively, maybe another name could be `pdfjs-integration` to suggest that the components are meant to plug into the PDF.js interface.

4. Should the `TextSelectionMenu` be moved into a new folder?

5. I could imagine moving `EntityLink` and `EntityPageLink` into `common/` as they feel like small utility components that will appear in many places (i.e., they appear both in the drawer and the glosses), though this is not a strong preference.

andrewhead (approved):

andrewhead:
Btw, great catch on the dead files!

PR #176, opened 2020-12-18
Pipeline: Fix up command filters; disable glossary terms by default

andrewhead:
This PR serves two purposes.

First, it disables the detection of glossary terms by the pipeline by default. This is because the glossary terms appear so often in papers, and provide such generic definitions, that I believe it will be better to disable them in our first alpha tests. Paper-specific terms and definitions (i.e., that appear within a paper, rather than in an external glossary like the Google Machine Learning Glossary) will still be extracted by default via the `definitions` entity type.

Second, it fixes buggy behavior in the pipeline arguments that filter which stages of the pipeline are run (e.g., `--entities`, `--commands`, `--start`, and `--end`. Namely, it supports:

* Default lists of entities to process that is not all entities but rather a subset (i.e., citations, symbols, and definitions)
* It allows specified in `--commands` to be run even after they appear after the `--end` filter

The following tests demonstrate that the filter arguments now achieving the desired behavior.

### Test 1: default behavior omits processing glossary terms:

```
bash
python scripts/run_pipeline.py -v 2>&1 | head -4`
```

Output:

```
...The following commands will be run, in this order: ['fetch-arxiv-sources', 'fetch-s2-metadata', 'unpack-sources', 'compile-tex', 'normalize-tex', 'compile-normalized-tex', 'raster-pages', 'extract-bibitems', 'resolve-bibitems', 'locate-bounding-boxes-for-citations', 'locate-citations', 'upload-citations', 'detect-sentences', 'locate-bounding-boxes-for-sentences', 'upload-sentences', 'detect-equations', 'locate-bounding-boxes-for-equations', 'upload-equations', 'extract-symbols', 'find-symbol-matches', 'extract-contexts-for-symbols', 'locate-bounding-boxes-for-equation-tokens', 'locate-bounding-boxes-for-symbols-with-affixes', 'locate-composite-symbols', 'collect-symbol-locations', 'upload-symbols', 'tokenize-sentences', 'create-annotation-files', 'detect-definitions', 'locate-bounding-boxes-for-definitions', 'upload-definitions']
```

### Test 2: Specifying commands without any other filters results in running just those commands

```bash
python scripts/run_pipeline.py -v --commands upload-citations 2>&1 | head -4
```

Output:
```
...The following commands will be run, in this order: ['upload-citations']
```

### Test 3: Specifying both `--entities` and `--commands` adds the command from 'commands' to the list of commands for the specified entity.

```bash
python scripts/run_pipeline.py -v --entities glossary-terms --commands upload-citations 2>&1 | head -4
```

Output (see 'upload-citations' about halfway through the list):

```
...The following commands will be run, in this order: ['fetch-arxiv-sources', 'fetch-s2-metadata', 'unpack-sources', 'compile-tex', 'normalize-tex', 'compile-normalized-tex', 'raster-pages', 'upload-citations', 'detect-sentences', 'locate-bounding-boxes-for-sentences', 'upload-sentences', 'detect-glossary-terms', 'locate-bounding-boxes-for-glossary-terms', 'extract-contexts-for-glossary-terms', 'upload-glossary-terms']
```

### Test 4: `--start` and `--end` filters apply to the list of commands for entities, but don't filter out the commands from `--commands`

```bash
python scripts/run_pipeline.py -v --entities glossary-terms --commands upload-citations --start detect-sentences --end upload-sentences  2>&1 | head -4
```

Output:

```
The following commands will be run, in this order: ['upload-citations', 'detect-sentences', 'locate-bounding-boxes-for-sentences', 'upload-sentences']
```

### DIFF
--- a/data-processing/scripts/run_pipeline.py
+++ b/data-processing/scripts/run_pipeline.py
@@ -36,11 +36,11 @@ from scripts.process import (
     run_command,
 )
 
+DEFAULT_ENTITIES = ["citations", "symbols", "definitions"]
+
 
 def run_commands_for_arxiv_ids(
-    CommandClasses: CommandList,
-    arxiv_id_list: List[str],
-    pipeline_args: Namespace,
+    CommandClasses: CommandList, arxiv_id_list: List[str], pipeline_args: Namespace,
 ) -> PipelineDigest:
     " Run a sequence of pipeline commands for a list of arXiv IDs. "
 
@@ -129,12 +129,24 @@ if __name__ == "__main__":
         "--entities",
         help=(
             "What type of entities to process. Commands that do not process the specified entities "
-            + "will be skipped. Defaults to all entities. You can specify multiple entity types. "
+            + f"will be skipped. Defaults to {DEFAULT_ENTITIES}. You can specify multiple entity types. "
             + "If specifying multiple entity types, use the format '--entities <type1> <type2> "
             + "without an equal sign after '--entities'."
         ),
         choices=ENTITY_NAMES,
         nargs="+",
+    )
+    parser.add_argument(
+        "--commands",
+        type=str,
+        nargs="+",
+        help=(
+            "Which commands to run. Typically when '--commands' is set, only those commands "
+            + "will be run. However, if other command filters are set (e.g., "
+            + "'--entities', '--start', or '--end', then all commands specified by the other "
+            + "filters will be run, _in addition to_ those in '--commands'."
+        ),
+        choices=command_names,
     )
     parser.add_argument(
         "--start",
@@ -143,18 +155,6 @@ if __name__ == "__main__":
     )
     parser.add_argument(
         "--end", help="Command to stop running the pipeline at.", choices=command_names
-    )
-    parser.add_argument(
-        "--commands",
-        type=str,
-        nargs="+",
-        help=(
-            "Which commands to run. Commands in this list will be run even if they don't belong "
-            + "to the pipelines in '--entities' or fit between '--start' and '--end'. If no other "
-            + "entity filters are set (--entities, --start, --end), then only the commands "
-            + "specified will be run."
-        ),
-        choices=command_names,
     )
     add_arxiv_id_filter_args(parser)
     parser.add_argument(
@@ -377,9 +377,19 @@ if __name__ == "__main__":
     if args.store_log:
         command_classes.append(StorePipelineLog)
 
+    # Determine what commands should be run.
     filtered_commands = []
 
+    # Determine the entities that will be processed. If no entities were specified,
+    # then set the entities to the default (or empty if a list of commands was provided).
+    entities: List[str] = []
+    if args.entities is not None:
+        entities = args.entities
+    elif args.commands is None or (args.start or args.end):
+        entities = DEFAULT_ENTITIES
+
     start_reached = True if args.start is None else False
+    end_reached = False
     for CommandClass in command_classes:
         command_name = CommandClass.get_name()
         # Run any command that the user explicitly requested to run.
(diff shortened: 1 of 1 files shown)


data-processing/scripts/run_pipeline.py
```diff
@@ -389,8 +399,10 @@ def run_commands_for_arxiv_ids(
                 continue
             # If no other command filters have been specified and the user didn't
             # request this command, skip it.
-            if args.start is None and args.end is None and args.entities is None:
+            if args.start is None and args.end is None and entities == []:
```
rreas, line 402: I'm pretty sure that what you want here is just

```python
if not args.start and not args.end and not entities:
  ...
```

Consider:

```text
python3
Python 3.8.6 (default, Oct  8 2020, 14:06:32)
[Clang 12.0.0 (clang-1200.0.32.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> if []:
...     print("HEY")
... else:
...     print("NO")
...
NO
>>> a = None
>>> b = None
>>> c = []
>>> d = None
>>> if not a and not b and not c:
...     print("HEY")
... else:
...     print("NO")
...
HEY
>>> if a is None and b is None and not c:
...     print("HEY")
... else:
...     print("NO")
...
HEY
>>> if not a and not b and not d:
...     print("HEY")
... else:
...     print("NO")
...
HEY
>>> if not a and b and not c:
...     print("HEY")
... else:
...     print("NO")
...
NO
>>>
```

data-processing/scripts/run_pipeline.py
```diff
@@ -377,9 +377,19 @@ def run_commands_for_arxiv_ids(
     if args.store_log:
         command_classes.append(StorePipelineLog)
 
+    # Determine what commands should be run.
     filtered_commands = []
 
+    # Determine the entities that will be processed. If no entities were specified,
+    # then set the entities to the default (or empty if a list of commands was provided).
+    entities: List[str] = []
+    if args.entities is not None:
```
rreas, line 386: Just nitting... `[]` is False in Python when used in an `if` statement (behaves like None). See my more extensive note below. Typically it's more Pythonic to say simply `if args.entities:` which will be True when `len(args.entities) > 0`

data-processing/scripts/run_pipeline.py
```diff
@@ -377,9 +377,19 @@ def run_commands_for_arxiv_ids(
...
+        entities = args.entities
+    elif args.commands is None or (args.start or args.end):
```
rreas, line 388: Parenthesis aren't needed here since logical ORs are commutative ;) ?

data-processing/scripts/run_pipeline.py
rreas: That said, if you prefer explicitness over what may not invoke 100% confidence in readers then I am totally A-OK with what you have for pretty much all my nits!

rreas (approved):
Excellent!

One thing I'm wondering about is in your Test 4 above... Shouldn't `glossary-terms` imply `fetch-arxiv-sources`, `fetch-s2-metadata`, etc. as well? If I ran that command on a new box, I think it would fail without that data?

rreas:
Actual comments on code are just nits, feel free to ignore. Main question is in conversation text above.

andrewhead:
Hey @rreas! 

> One thing I'm wondering about is in your Test 4 above... Shouldn't glossary-terms imply fetch-arxiv-sources, fetch-s2-metadata, etc. as well? If I ran that command on a new box, I think it would fail without that data?

Good thinking. In the case of Test 4, the addition `--start` and `--end` flags start the pipeline after the fetch-arxiv-sources, fetch-s2-metadata, etc. commands. They're kind of like the `--start-at-task` flags like in a tool like Ansible that tell the pipeline to skip a bunch of steps. These flags are mostly there for dev purposes to make it easier to test changes to a subset of commands in the middle of the pipeline. Does that make sense, or do you think there's an alternative behavior that would be more intuitive?

data-processing/scripts/run_pipeline.py
andrewhead: This works! I'm usually explicit to a fault in these scripts XD. I think your way looks great too; I'll switch it over!

andrewhead:
Thanks for taking a look @rreas! It's been updated with style fixes :-)

PR #177, opened 2020-12-18
[WIP] Improve quality of definitions extracted by rule-based nickname detector

andrewhead:
@nuwandavek is leading the creation of this PR to:

1. Filter out low-quality nicknames consisting mostly of other symbols
2. Detecting other types of nickanmes (e.g., ones of the form "<symbol>: <nickname>"

I will let @nuwandavek add additional clarifications here.

### DIFF
--- a/data-processing/entities/definitions/commands/detect_definitions.py
+++ b/data-processing/entities/definitions/commands/detect_definitions.py
@@ -320,7 +320,7 @@ def search_symbol_nickname(
 
     # Union set of POS tags in nicknames. Feel free to add more if you have new patterns.
     # POS Tag Expansions : https://www.ling.upenn.edu/courses/Fall_2003/ling001/penn_treebank_pos.html.
-    UNION_POS_LIST = ["DT", "JJ", "NN", "NNS", "NNP", "NNPS"]
+    UNION_POS_LIST = ["DT", "JJ", "NN", "NNS", "NNP", "NNPS", ":"]
     pos_indexes = []
     pos_tags = []
     pos_ranges = []
@@ -340,11 +340,15 @@ def search_symbol_nickname(
         pos_tags.append(pos[current_index])
         if direction == "RIGHT":
             current_index += 1
+            if current_index >= len(pos):
+                break
         else:
             current_index -= 1
+            if current_index < 0:
+                break
 
-    # Ignore when the POS tag is [DT] or [IN].
-    if len(pos_tags) == 1 and (pos_tags[0] == "DT"):
+    # Ignore when the POS tag is among ["DT", ":"].
+    if len(pos_tags) == 1 and (pos_tags[0] in ["DT", ":"]):
         return None
 
     if len(pos_indexes) > 0:
@@ -357,11 +361,11 @@ def search_symbol_nickname(
             nickname_ranges = list(pos_ranges)
             nickname_tags = list(pos_tags)
 
-        # Skip 'DT' or 'IN' as first or last nickname tokens.
-        if nickname_tags[0] == "DT":
+        # Skip ["DT", ":"] as first or last nickname tokens.
+        if nickname_tags[0] in ["DT", ":"]:
             nickname_ranges = nickname_ranges[1:]
             nickname_indexes = nickname_indexes[1:]
-        elif nickname_tags[-1] == "DT":
+        elif nickname_tags[-1] in ["DT", ":"]:
             nickname_ranges = nickname_ranges[:-1]
             nickname_indexes = nickname_indexes[:-1]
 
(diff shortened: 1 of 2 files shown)


data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -377,6 +381,24 @@ class SymbolNickname(NamedTuple):
     nickname_indexes: List[int]
     nickname_ranges: List[CharacterRange]
 
+def get_filtered_symbol_nickname_pairs(
```
andrewhead, line 384: Nit pick: maybe this could be more succinctly named the following?

```suggestion
def filter_symbol_nickname_pairs(
```

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -377,6 +381,24 @@ class SymbolNickname(NamedTuple):
...
+    # Filter symbol-nickname pairs based on the following conditions.
+    # 1. The nickname of a symbol is another symbol.
+    filtered_pairs: List[SymbolNickname] = []
+    for sn in pairs:
+        nickname_start = min(
+            [nickname_range.start for nickname_range in sn.nickname_ranges]
+        )
+        nickname_end = (
+            max([nickname_range.end for nickname_range in sn.nickname_ranges]) + 1
+        )
```
andrewhead, line 397: nit pick: Converting text ranges into text like something the nickname filtering function shouldn't have to take care of. Might it make sense to extract the text for a nickname when the SymbolNickname is getting created in `get_symbol_nickname_pairs`, and just inspect the text here?

data-processing/tests/test_extract_definitions.py
```diff
@@ -85,6 +85,70 @@ def test_extract_nicknames_from_before_symbols():
     assert nickname1.definition_text == "timestep"
 
 
+def test_extract_nicknames_symbols_separated_by_colon():
+    text = "The agent acts with SYMBOL in each timestep SYMBOL, where SYMBOL : policy."
```
andrewhead, line 89: Can this sentence be made even shorter, e.g., something like "We defined SYMBOL: policy."? A shorter sentence minimizes the things that could accidentally yield other nicknames as we extend the nickname extractor. For instance, we already check on the pattern "timestep SYMBOL" in another test, so including it here makes more work for us if we change the behavior of the nickname extractor later.

data-processing/tests/test_extract_definitions.py
```diff
@@ -85,6 +85,70 @@ def test_extract_nicknames_from_before_symbols():
...
+    nickname0 = symbol_nickname_pairs[0]
+    assert nickname0.term_text == "t"
+    assert nickname0.definition_text == "timestep"
+
+    nickname1 = symbol_nickname_pairs[1]
+    assert nickname1.term_text == r"\pi"
+    assert nickname1.definition_text == "policy"
+
+def test_extract_nicknames_symbols_filter():
+    text = "The agent acts with SYMBOL SYMBOL in each timestep SYMBOL."
```
andrewhead, line 125: Same suggestion here to minimize complexity of this sentence to make the test case less brittle if we change parts of the nickname extractor in the future:

```suggestion
    text = "The agent acts with SYMBOL SYMBOL."
```

andrewhead (commented):
This is looking good so far!

andrewhead:
Vivek, here are a couple of patterns I saw in the ELMo paper (arXiv ID 1802.05365):

1. "... nickname , SYMBOL, ..." (with comma)

image

image

2. "... nickname (SYMBOL) ..."

image

Did you see any of these cases when you inspected your sample of papers? If they're sufficiently common, it would be great if we could incorporate them into this PR. If they are not that common, we can probably skip for now.

data-processing/entities/definitions/commands/detect_definitions.py
nuwandavek: Yeah, this is true. But I was wondering if we want to add other filters as well that might not just depend on the nickname text. But we can maybe have a boolean function.

nuwandavek:
> Vivek, here are a couple of patterns I saw in the ELMo paper (arXiv ID 1802.05365):
> 
>     1. "... nickname , SYMBOL, ..." (with comma)
> 
> 
> image
> 
> image
> 
>     1. "... nickname (SYMBOL) ..."
> 
> 
> image
> 
> Did you see any of these cases when you inspected your sample of papers? If they're sufficiently common, it would be great if we could incorporate them into this PR. If they are not that common, we can probably skip for now.

@andrewhead  I added the parentheses case. That makes sense. But I am slightly skeptical of adding the comma case. Though this may be a valid case sometimes, there are many many examples with "... we use L layers, V vectors, K keys ...". Here using the comma heuristic is wrong. I think the example you mentioned requires reasoning over a more global span, than just the local heuristic.

data-processing/entities/definitions/commands/detect_definitions.py
andrewhead: Fair enough. Do you think they would depend on the character ranges? Maybe it's worth us still attaching the text to the SymbolNickname to support easier checks in this filter function, while keeping the range properties on the nickname too in case we want to add that in the future..

andrewhead:
> But I am slightly skeptical of adding the comma case. Though this may be a valid case sometimes, there are many many examples with "... we use L layers, V vectors, K keys

I see, this is a good point; it would require more global reasoning (for instance, checking to see that the clause ended after the symbol to the right of the comma). I can see that that is probably out of scope for now.

Thanks for thinking through it, and for also including the case with the parentheses!

andrewhead (commented):
This gets my approval. Let's decide whether we want to attach a `.text` property to the SymbolNickname and, if so, we'll make the change, rebase, the merge!

Can you do a rebase on top of master?

dykang:
This looks great to me! Thanks, @nuwandavek  for your hard work!

dykang (approved):

PR #179, opened 2020-12-24
[DON'T MERGE] Features for more efficient annotation of ground truth entity locations

andrewhead:
This PR is not meant to be merged into the main branch, but rather to serve as a place for to review a branch used to create a dataset for evaluating the pipeline. @kyleclo and maybe I will be using this branch to make annotations of bounding boxes of symbols and sentences.

This first commit offers an "rapid annotation mode" in the entity creation toolbar. Turning on rapid annotation mode then supports rapid annotation in two ways:

1. It colorizes all boxes for citations red and boxes for symbols blue
2. It allows you to create bounding boxes that overlap each other (earlier, this was finnicky, as it when you click on an existing bounding box, it selects that box and cancels the creation of a new box).

Rapid Annotation Mode

This code is a bit of a hack, though was left messy as it's not intended to be merged into the main branch.

### DIFF
--- a/ui/src/ScholarReader.tsx
+++ b/ui/src/ScholarReader.tsx
@@ -1,23 +1,38 @@
+import classNames from "classnames";
+import React from "react";
 import * as api from "./api/api";
-import AppOverlay from "./components/overlay/AppOverlay";
+import {
+  Entity,
+  EntityCreateData,
+  EntityUpdateData,
+  isCitation,
+  isEquation,
+  isSymbol,
+  isTerm,
+  Paper,
+  Symbol,
+} from "./api/types";
 import Control from "./components/control/Control";
-import DefinitionPreview from "./components/preview/DefinitionPreview";
-import { Drawer, DrawerContentType } from "./components/drawer/Drawer";
-import EntityAnnotationLayer from "./components/entity/EntityAnnotationLayer";
 import EntityCreationCanvas from "./components/control/EntityCreationCanvas";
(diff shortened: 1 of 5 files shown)


andrewhead:
The second commit is a "power deletion" mode which allows the deletion of annotations with a single click. Like with rapid annotation mode, all existing annotations are highlighted in color to make it easier to see what annotations are faulty and therefore should be deleted:

Power Deletion Mode

Note that power deletion mode only works when rapid annotation mode is disabled.

andrewhead:
I just made a few more updates to this branch to support a few more use cases in annotation, particularly with annotating sentences. The updates are:

1. I made the entity creation toolbar vertical rather than horizontal
2. Highlighting existing entities is now controlled with the "Highlight Entities" toggle, instead of "Power add" or "Power delete"
3. It is now possible to group multiple boxes into one entity (like, if you want to make one sentence from multiple boxes). To do this, do the following:

  * Turn off "Power add"
  * Shift-click the entities you want to group, holding down Shift the whole time
  * Click on the "Group Entities" button in the entity creation toolbar

Create Sentence

kyleclo:
hi @mjlangan!  we've been using this annotation tool to curate "gold" bounding boxes & perform manual evaluation of the system.  we were wondering if it's possible to merge this into what we have in master now?   what do you think?  this "annotation tool" interacts w/ the API in a manner that we're unsure if we want as part of the main tool.

PR #182, opened 2020-12-30
Update pipeline log messages: remove verbose messages; add timing messages

andrewhead:
I'm considering making updates to the pipeline to make it faster to process papers. To determine the effort required to speed up the pipeline, I'd like to collect more detailed timing information about key steps in the paper processing pipeline.

This PR adds log statements to the commands for locating entities that can be used to measure the time it takes to:

1. Insert color macros into the TeX
2. Compile the TeX
3. Raster the TeX to images
4. Difference the images
5. Locate entities in the image diffs

These steps will be performed up to thousands of times for a single paper, so strategic improvements to pipeline performance can be aided by knowing which one of these steps usually takes the longest.

Log statements are added both at the start and end of each of the steps above to make it possible for our log analysis scripts to measure the of that step as the difference in timestamps between the start and end log message.

I have also removed a couple of overly verbose log messages from the project which have never really served a purpose in helping me (or as far as I know, anyone else) debug.

### DIFF
--- a/data-processing/common/commands/locate_entities.py
+++ b/data-processing/common/commands/locate_entities.py
@@ -13,23 +13,14 @@ from common.colorize_tex import ColorizedTex, colorize_entities
 from common.commands.base import ArxivBatchCommand
 from common.commands.compile_tex import save_compilation_result
 from common.commands.raster_pages import raster_pages
-from common.compile import (
-    compile_tex,
-    get_compiled_tex_files,
-    get_last_autotex_compiler,
-    get_last_colorized_entity_id,
-)
+from common.compile import (compile_tex, get_compiled_tex_files,
+                            get_last_autotex_compiler,
+                            get_last_colorized_entity_id)
 from common.diff_images import diff_images_in_raster_dirs
 from common.locate_entities import locate_entities
-from common.types import (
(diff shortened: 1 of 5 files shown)


kyleclo (approved):
lgtm

rreas:
Are you planning to run these locally or do you want the output on the S2 pipeline side as well? I'm mainly wondering if we should ignore DEBUG if this is going to be emitted 1k+ times per paper. Eventually that will add up in terms of $ on CloudWatch storage / retrieval.

rreas (approved):

andrewhead:
> Are you planning to run these locally or do you want the output on the S2 pipeline side as well? I'm mainly wondering if we should ignore DEBUG if this is going to be emitted 1k+ times per paper. Eventually that will add up in terms of $ on CloudWatch storage / retrieval.

My aim was to run this locally on a small number of papers. While I could envision wanting to run this on the S2 pipeline, I think I would only need to do it for a small subset of papers at a time when we're trying to debug or profile something specific. What this suggests to me is that perhaps we should ignore DEBUG by default, though provide the ability to enabled DEBUG through a configuration or command line parameter for the deployed pipeline. What would you think of that?

PR #194, opened 2021-01-07
Add CORS origins so S2 webapp can hit ScholarPhi API

mjlangan:
This CORS config should allow the S2 webapp (when running either in a deployed instance or on a developer laptop) to make API requests to ScholarPhi.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -253,6 +253,9 @@ export const plugin = {
         return h.response({ version }).code(200);
       },
       options: {
+        cors: {
+          origin: ['*.semanticscholar.org', '*.allenai.org'],
+        },
         validate: {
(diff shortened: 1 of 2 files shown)


codeviking (approved):
LGTM;

andrewhead (approved):
Great! It sounds like hapi was not allowing this earlier. Out of curiosity, what was the fix?

mjlangan:
@andrewhead the documentation is kind of unclear as to where I'm supposed to stick the CORS config object. I initially had it set on the route's options, which resulted in that route failing to initialize and the server crashing with a very unhelpful `Error: Invalid path: /api/v0//{p*}`. It was only after I searched through the issues in their GitHub repo that I found an example of the correct placement for the CORS config.

PR #195, opened 2021-01-08
Update computation of bounding box detection accuracy

andrewhead:
This PR introduces revives scripts for computing the accuracy of the bounding boxes extracted by the pipeline.

Of particular note, the scripts for computing bounding box accuracy were extended to handle two special cases:

1. When bounding boxes in the gold set match multiple extracted bounding boxes---in this case the box in the gold set should probably only be matched to a single extracted bounding box to not overinflate the accuracy
2. When bounding boxes in the gold set should be clustered into groups (like with sentences), and matches should be determined based on whether a group of boxes in the gold set matches a group of extracted boxes.

The main change, then, is a function called `iou_per_region`. It finds the "ideal" match between two sets of groups of bounding boxes. Specifically, it:

1. Takes as input (a) a set of gold groups of bounding boxes and (b) a set of extracted groups of bounding boxes (each group corresponding to a sentence, a symbol, a citation, etc.)
2. Determines _all_ pairs from set (a) and set (b) that overlap more than a certain threshold
3. Solves a linear assignment problem to determine the maximum number of groups from set (a) that can be matched to those from set (b), without matching any box from one set to more than one from the other
4. Returns the IOU for the boxes between the two sets that were assigned to each other

### DIFF
--- a/data-processing/common/commands/compute_iou.py
+++ b/data-processing/common/commands/compute_iou.py
@@ -3,7 +3,7 @@ import os.path
 from argparse import ArgumentParser
 from collections import defaultdict
 from dataclasses import dataclass
-from typing import Dict, FrozenSet, Iterator, List, Optional, Tuple
+from typing import Dict, FrozenSet, Iterator, List, Optional, Set, Tuple
 
 from common import directories, file_utils
 from common.bounding_box import (
@@ -52,8 +52,8 @@ class IouJob:
 @dataclass(frozen=True)
 class IouResults:
     page_iou: float
-    precision: float
-    recall: float
+    precision: Optional[float]
+    recall: Optional[float]
     matches: RegionMatches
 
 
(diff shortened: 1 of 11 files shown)


data-processing/common/commands/compute_iou.py
```diff
@@ -87,69 +121,112 @@ def load(self) -> Iterator[IouJob]:
...
-            ] = {}
-
-            entity_types = [
-                EntityKeyPair("citations", "citation"),
-                EntityKeyPair("equations", "equation"),
-                EntityKeyPair("equation-tokens", "symbol"),
+            entity_keys = [
+                # EntityKeys("citations", "citation"),
+                # EntityKeys("equations", "equation"),
+                EntityKeys("symbols", "symbol"),
```
kyleclo, line 127: what's happening here w/ the commenting-out?

andrewhead: I am updating the script to only compute accuracy of symbol detection, and not the accuracy of citation and equation detection. Though I should revert this to remove the comments before merge.

kyleclo (approved):
lgtm; just had a question

PR #197, opened 2021-01-11
Pipeline: Make LaTeX coloring macros and error checking more robust

andrewhead:
This PR contains two fixes:

First, the code for the pipeline commands is made robust to strange errors that arise when a paper is compiled from multiple main files (which is in fact possible for arXiv papers).

Second, the coloring macros have been made slightly more robust. To support both TeX and LaTeX, the coloring macros have conditional logic that depends on our custom macro `\scholarifdefined...`, which can be used to see what coloring commands are available in the environment, and find out other useful information.

It turns out that the way we were checking if a command was defined had the side effect of _defining_ that command to be `\relax` if it was not yet defined. In most cases, this did not matter. For the supplemental material (`supp.tex`) of paper 1712.05055, it did matter. I believe what was happening was that the style file had a check for whether hypersetup was defined. Typically, it is not. Our code defined it (albeit to a no-op value). This changed the behavior of the style file, and thefore the appearance of the document.

This fix will not fix the problem for _all_ Tex documents, as it only works for papers that are compiled with an e-TeX-based backend. It worked for this paper, so I am hopeful it will work for many others.

### DIFF
--- a/data-processing/common/normalize_tex.py
+++ b/data-processing/common/normalize_tex.py
@@ -106,7 +106,9 @@ def expand_tex(
     # In a specific sample of about 120 conference papers, only 5 had '\include' macros, yet
     # many more had '\input' commands). Only 1 used an '\include' macro to read in text.
     # The rest of the files used '\include' macros to include macros and usepackage statements.
-    include_patterns = [
+    # XXX(andrewhead): The 'includes' patterns are currently disabled because the TeX that is
+    # being inserted in their place is incorrect (i.e., it causes compilation errors).
+    include_patterns: List[Pattern] = [
         # Pattern("include_braces", r"\\include\s*{([^}]+)}"),
         # Pattern("include", r"\\include\s+(\S+)"),
     ]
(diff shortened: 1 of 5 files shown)


rreas (approved):

data-processing/common/normalize_tex.py
```diff
@@ -107,8 +107,8 @@ def expand_tex(
     # many more had '\input' commands). Only 1 used an '\include' macro to read in text.
     # The rest of the files used '\include' macros to include macros and usepackage statements.
     include_patterns = [
-        Pattern("include_braces", r"\\include\s*{([^}]+)}"),
-        Pattern("include", r"\\include\s+(\S+)"),
+        # Pattern("include_braces", r"\\include\s*{([^}]+)}"),
```
kyleclo, line 110: can leave a note as to why these are commented out?

kyleclo (approved):
lgtm; see comments

data-processing/common/normalize_tex.py
andrewhead: Yes, done!

PR #205, opened 2021-01-19
UI: Disable declutter by default; only enable citation glosses and sy…

andrewhead:
This PR addresses issue https://github.com/allenai/scholar/issues/26218 (cc @aschokking)

It accomplishes two goals:

1. It disables the 'declutter' symbol interaction by default. As it turns out, we already had a flag in the user interface for turning on and off 'declutter.' This PR modifies the default settings such that the only features enabled are symbol search and citation abstracts.

2. It removes out some of the old settings presets that were used exclusively for our formal usability study.

A GIF showing proof of successful implementation is forthcoming.

### DIFF
--- a/ui/src/settings.ts
+++ b/ui/src/settings.ts
@@ -110,6 +110,9 @@ const PRESETS: Preset[] = [
     key: "demo",
     primerInstructionsEnabled: false,
     useDefinitionsForDiagramLabels: true,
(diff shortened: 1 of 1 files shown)


mjlangan (approved):

aschokking (approved):
👍  yay for flags

PR #206, opened 2021-01-20
Pipeline: Several bugfixes to support ScholarPhi paper

andrewhead:
Several changes were needed to make the ScholarPhi pipeline robust enough to handle the [ScholarPhi paper](https://arxiv.org/abs/2009.14237). Namely:

1. A bug was found in the citation color macros where a coloring macro was calling itself recursively. That has now been fixed (`03-load-color-commands.tex`)

2. TeX logs can apparently contain non-utf-8 characters. So the logs needed to be loaded in a way that was tolerant to unexpected non-utf-8 characters (`locate_entities.tex`)

3. The bibliography entries for this paper were written idiosyncratically---often there was a newline between the `\bibitem` macro and the `{...}` curly braces in which the key for the reference was written. This was confusing our code that parsed out reference keys from the bibliography. This has since been fixed (see `extractor.py` and `test_parse_tex.py`).

The following GIF shows citation interactions working for the ScholarPhi paper in a local version of the app:

citations_working

### DIFF
--- a/data-processing/common/commands/locate_entities.py
+++ b/data-processing/common/commands/locate_entities.py
@@ -629,9 +629,12 @@ def get_last_colorized_entity(
         )
         return None
 
-    with open(original_autogen_log_path) as file_:
+    # If TeX can process data that is not utf-8, then that non-utf-8 data can also be printed
+    # out to the AutoTeX log. Therefore, when reading the AutoTeX log, the files need to be opened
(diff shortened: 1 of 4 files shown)


andrewhead:
cc @aschokking

aschokking (approved):
👍

rreas (approved):

PR #207, opened 2021-01-20
Make /version API aware of entity data

mjlangan:
S2 uses this API to determine whether or not to display the Semantic Reader button on the PDP, and to prevent the button from displaying for papers that don't have any extracted entities, this change ensures we only report back with version data if we also have extracted entities for that paper.

### DIFF
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -116,7 +116,8 @@ export class Connection {
 
   async getPaperEntityCount(paperSelector: PaperSelector): Promise<{paper: PaperSelector, count: number} | null> {
     const idField = isS2Selector(paperSelector) ? 'p.s2_id' : 'p.arxiv_id';
-    const idValue = isS2Selector(paperSelector) ? paperSelector.s2_id : paperSelector.arxiv_id;
+    const idValue = isS2Selector(paperSelector) ? paperSelector.s2_id : `${paperSelector.arxiv_id}%`;
+    const whereClause = `${idField} ${isS2Selector(paperSelector) ? '=' : 'ilike'} ?`
     const response = await this._knex.raw<{ rows: {count: number, id: string}[] }>(`
       select count(e.*), ${idField}
       from paper p
@@ -126,7 +127,7 @@ export class Connection {
         from entity
         group by paper_id order by paper_id
       ) as maximum on maximum.paper_id = e.paper_id
-      where e.version = maximum.max_version and ${idField} = ?
+      where e.version = maximum.max_version and ${whereClause}
       group by p.s2_id
     `, [idValue]);
     if (response.rows.length > 0) {
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -114,6 +114,30 @@ export class Connection {
     return { rows, offset, size, total };
   }
 
+  async getPaperEntityCount(paperSelector: PaperSelector): Promise<{paper: PaperSelector, count: number} | null> {
+    const idField = isS2Selector(paperSelector) ? 'p.s2_id' : 'p.arxiv_id';
+    const idValue = isS2Selector(paperSelector) ? paperSelector.s2_id : paperSelector.arxiv_id;
+    const response = await this._knex.raw<{ rows: {count: number, id: string}[] }>(`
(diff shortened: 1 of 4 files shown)


api/src/db-connection.ts
```diff
@@ -114,6 +114,31 @@ export class Connection {
     return { rows, offset, size, total };
   }
 
+  async getPaperEntityCount(paperSelector: PaperSelector): Promise<{paper: PaperSelector, count: number} | null> {
```
andrewhead, line 117: Stylistic question: why return the paper selector with the count?

api/src/tests/api.test.ts
```diff
@@ -957,4 +957,81 @@ describe("API", () => {
...
+          entity_id: 1,
+          source: "test",
+          page: 0,
+          left: 0,
+          top: 0,
+          width: 0.1,
+          height: 0.05,
+        },
+      ] as BoundingBoxRow[]);
+      await knex.batchInsert("entitydata", [
```
andrewhead, line 992: I note that the inserts into `boundingbox` and `entitydata` are not strictly necessary to pass the test. Is it worth removing them for the sake of test case simplicity? Or perhaps only passing the version check if entities have bounding boxes / data?

api/src/tests/api.test.ts
```diff
@@ -957,4 +957,81 @@ describe("API", () => {
...
+      });
+
+      expect(response.statusCode).toEqual(200);
+      expect(response.result).toEqual({ version: 1 });
+    });
+
+    test("paper with no entity data", async () => {
+      await knex("paper").insert({
+        s2_id: "s2id",
+        arxiv_id: "1111.1112v1",
```
andrewhead, line 1016: nit pick; the use of a qualified arxiv_id here ("1111.1112v1") versus unqualified below (1111.1112) results in this test testing two small pieces of logic (i.e., that 1111.1112 matches the qualified arxiv_id, but then the version check fails because there are no entities). I wonder if it makes sense for us to decouple those two pieces of logic in this test and just use the same arxiv ID in both the database and the query.

api/src/db-connection.ts
```diff
@@ -114,6 +114,31 @@ export class Connection {
...
+    const idValue = isS2Selector(paperSelector) ? paperSelector.s2_id : `${paperSelector.arxiv_id}%`;
+    const whereClause = `${idField} ${isS2Selector(paperSelector) ? '=' : 'ilike'} ?`
+    const response = await this._knex.raw<{ rows: {count: number, id: string}[] }>(`
+      select count(e.*), ${idField}
+      from paper p
+      join entity e on e.paper_id = p.s2_id
+      join (
+        select paper_id, max(version) as max_version
+        from entity
+        group by paper_id order by paper_id
```
andrewhead, line 128: out of curiosity, what is the purpose of ordering by paper_id?

andrewhead (approved):
lgtm! I left a few suggestions throughout that might provide opportunities to simplify.

api/src/tests/api.test.ts
mjlangan: I'll give that a try. It's hard to tell what setup is necessary with these heavyweight integration tests that require actual data in the DB to test API behaviors. Ideally our tests would allow me to mock the DB connection and write unit tests on just the API layer instead.

api/src/tests/api.test.ts
mjlangan: S2 does not store the version when we record the arXiv ID of a paper, so these tests follow the way S2 will be interacting with the API.

api/src/db-connection.ts
mjlangan: In this case, it's a leftover from me figuring out what the queries should be :)

api/src/db-connection.ts
mjlangan: Not necessarily needed here.

api/src/tests/api.test.ts
andrewhead: Fair enough. I would be interested in discussing how make the tests less heavy-handed. Admittedly I did want to have tests for the SQL queries since some of them get pretty complex. That said, it is annoying that the tests require so much setup.

PR #208, opened 2021-01-20
UI: Add guards to sorting / display logic when an entity has no bounding boxes

andrewhead:
It sometimes happens that the pipeline uploads entities that have no bounding boxes at all. While most of the UI code has guards to check whether an entity has some bounding boxes before attempting to do something with those boxes, some of it does not. This leads to exceptions.

This PR introduces two small fixes in the form of guards in the UI code that check for the presence of bounding boxes for an entity before attempting to do something with those bounding boxes. In particular, the guards appear before:

1. Sorting two entities based on the position of the bounding boxes
2. Rendering a preview of a symbol

Guard #1 has been tested to fix the following exception for paper 2101.05042v1 in a local version of the UI with the fixes incorporated:

image

Clicking on a symbol in that paper now no longer results in this exception.

cc @kyleclo, who authored the first version of this fix.

### DIFF
--- a/ui/src/components/preview/SymbolPreview.tsx
+++ b/ui/src/components/preview/SymbolPreview.tsx
@@ -1,8 +1,7 @@
-import PaperClipping from "./PaperClipping";
-import { Sentence, Symbol } from "../../api/types";
-
 import { PDFDocumentProxy } from "pdfjs-dist";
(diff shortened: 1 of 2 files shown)


aschokking (approved):
Looks good to me

PR #211, opened 2021-01-23
Fix s2 library API call

mjlangan:
The request payload of the library API on S2 now expects an array of numbers named `folders` instead of an array of strings named `tags`.


andrewhead (approved):
Looks great!

PR #213, opened 2021-01-26
Remove the "Cite" button

mjlangan:
Ref https://github.com/allenai/scholar/issues/26346

Also sets a minimum width on the PaperSummary component so it doesn't get too narrow.
image


andrewhead (approved):
lgtm!

PR #214, opened 2021-01-27
Link to S2 when users save to library when not logged in

mjlangan:
Ref https://github.com/allenai/scholar/issues/26347

### DIFF
--- a/ui/src/components/entity/citation/PaperSummary.tsx
+++ b/ui/src/components/entity/citation/PaperSummary.tsx
@@ -189,7 +189,7 @@ export default class PaperSummary extends React.PureComponent<Props, State> {
           {this.state.showLoginMessage && (
             <React.Fragment>
               Before you can save papers to your library, you must be logged into Semantic Scholar.
-              Visit <ExternalLink href="https://www.semanticscholar.org" allowReferrer>https://semanticscholar.org</ExternalLink> to log in.
+              Visit <ExternalLink href="https://www.semanticscholar.org/me/library" allowReferrer>Semantic Scholar</ExternalLink> to log in.
               Then refresh this page and try again.
             </React.Fragment>
           )}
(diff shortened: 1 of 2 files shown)


ui/src/components/entity/citation/PaperSummary.tsx
```diff
@@ -199,7 +185,14 @@ export default class PaperSummary extends React.PureComponent<Props, State> {
         </div>
 
         <div className="paper-summary__library-error">
-          {this.state.errorMessage && this.state.errorMessage}
+          {!!this.state.errorMessage && this.state.errorMessage}
```
mjlangan, line 188: This change fixes a common issue we ran into in S2--if `this.state.errorMessage` is the string `'0'`, this will render a `0` on the page.

ui/src/components/entity/citation/PaperSummary.tsx
```diff
@@ -199,7 +185,14 @@ export default class PaperSummary extends React.PureComponent<Props, State> {
...
+          {this.state.showLoginMessage && (
+            <React.Fragment>
+              Before you can save papers to your library, you must be logged into Semantic Scholar.
+              Visit <ExternalLink href="https://www.semanticscholar.org" allowReferrer>https://semanticscholar.org</ExternalLink> to log in.
```
aschokking, line 192: propose a slight tweak, link to https://www.semanticscholar.org/me/library instead so people will see a sign in page right away

mjlangan: Sounds good. Since that makes the URL longer I'll also just rename the link to "Semantic Scholar" in the UI.

ui/src/components/entity/citation/PaperSummary.tsx
andrewhead: excellent, and great catch! thank you for fixing this.

andrewhead (approved):

PR #215, opened 2021-01-27
Rename ScholarPhi -> Semantic Reader in public-facing places

mjlangan:
Ref https://github.com/allenai/scholar/issues/26407

Since "Semantic Reader" is the official name for the app, this changes some publicly-facing parts of the code to use the new name. It also makes me think we might want to rename the Slack channel while we're at it.


andrewhead:
LGTM. I would also like to point to the tab title here, which may be worth considering changing as part of this PR as well.

https://github.com/allenai/scholarphi/blob/2b6e2e8aaac77656623312e21447f801842df447/ui/public/viewer.html#L35

andrewhead (approved):

PR #216, opened 2021-01-28
Symbol parsing cleanup

andrewhead:
This PR contributes two changes.

1. Detect operator symbols (in case we eventually want to show explanations for them too)
2. Identify function symbols earlier in the pipeline---i.e., in an earlier stage in which the MathML for an equation is cleaned, rather than a later stage where that MathML is parsed for symbols. This may seem not useful, though it is a cleanup that will make it easier for me to make some other symbol segmentation improvements soon.

Because operator symbols will now be uploaded to the database, we need a way to filter them out of the UI until we decide we want to show them. Therefore, there are changes in the API and UI:

* API: extend the symbol attributes type to have a `type` field, which can be set to "identifier", "function", or "operator
* UI: only render annotations for symbols that are of type "identifier" or "function"

### DIFF
--- a/ui/src/api/types.ts
+++ b/ui/src/api/types.ts
@@ -206,6 +206,7 @@ export interface Symbol extends BaseEntity {
 
 export interface SymbolAttributes extends BaseEntityAttributes {
   tex: string | null;
+  type: "identifier" | "function" | "operator";
   mathml: string | null;
   mathml_near_matches: string[];
   is_definition: boolean | null;
--- a/ui/src/components/entity/EntityAnnotationLayer.tsx
+++ b/ui/src/components/entity/EntityAnnotationLayer.tsx
@@ -316,6 +316,14 @@ class EntityAnnotationLayer extends React.Component<Props, {}> {
             );
           } else if (isSymbol(entity)) {
             /*
+             * Only show annotations for identifiers and functions (i.e., not operators).
+             */
+            const ANNOTATED_SYMBOL_TYPES = ["identifier", "function"];
+            if (ANNOTATED_SYMBOL_TYPES.indexOf(entity.attributes.type) === -1) {
+              return null;
+            }
+
+            /*
              * If the symbol appears in a selected equation, don't render it; the
              * equation diagram provides its own targets to click.
              */
--- a/ui/src/components/entity/EntityAnnotationLayer.tsx
+++ b/ui/src/components/entity/EntityAnnotationLayer.tsx
@@ -316,7 +316,8 @@ class EntityAnnotationLayer extends React.Component<Props, {}> {
             );
           } else if (isSymbol(entity)) {
             /*
-             * Only show annotations for identifiers and functions (i.e., not operators).
+             * Only show annotations for identifiers and functions (i.e., not operators). In
+             * the future, this filter could be exposed in the admin control panel.
              */
             const ANNOTATED_SYMBOL_TYPES = ["identifier", "function"];
             if (ANNOTATED_SYMBOL_TYPES.indexOf(entity.attributes.type) === -1) {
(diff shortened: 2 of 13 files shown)


kyleclo:
1. seems reasonable.  this decouples "we can identify stuff" from "show this & not this".  should the latter be determined by the UI though, and not a final pipeline step that marks certain symbols in the DB as "show/not show"?
2. can u give example of how identifying function symbols earlier in pipeline helps?

ui/src/api/types.ts
```diff
@@ -162,6 +162,7 @@ export interface Symbol extends BaseEntity {
 
 export interface SymbolAttributes extends BaseEntityAttributes {
   tex: string | null;
+  type: "identifier" | "function" | "operator";
```
kyleclo, line 165: will this ever get extended?  if so, can  you make a note here what other code need to get changed?

ui/src/components/entity/EntityAnnotationLayer.tsx
```diff
@@ -309,6 +308,14 @@ class EntityAnnotationLayer extends React.Component<Props, {}> {
               />
             );
           } else if (isSymbol(entity)) {
+            /*
+             * Only show annotations for identifiers and functions (i.e., not operators).
+             */
+            const ANNOTATED_SYMBOL_TYPES = ["identifier", "function"];
```
kyleclo, line 314: do we have a config for the UI that determines this?

andrewhead: No, we do not 😞  Do you think we should have one?

ui/src/api/types.ts
andrewhead: Sure thing! Good thinking 👍

andrewhead:
> can u give example of how identifying function symbols earlier in pipeline helps?

Sure thing! It is not an accuracy thing, but rather something that is making the code easier for me to reason about.

Not so long ago, the pipeline detected symbols in one big, complex step: it traversed a MathML parse tree, and then wherever it found a node that it thought was a symbol, it created a cleaned up version of the subtree for that symbol (for instance, removing spaces, or merging consecutive identifiers like 'm', 'a', 'x' into 'max').

This made the code hard to maintain, because it meant that the MathML parse tree would change in the middle of it being processed, as nodes in it were getting cleaned. So the code was refactored to find symbols in two steps:

Step 1. Transform the MathML tree into a canonicalized form
Step 2. Visit nodes in the MathML tree to find symbols

For functions, I wanted to transform the tree to group together nodes in a function under a new MathML element (i.e., under a `mrow` element), because such a representation of the function can be shown on its own in the UI. Because this required a transformation of the MathML tree, I split it out into Step 1, where the tree is modified, rather than Step 2.

andrewhead:
> seems reasonable. this decouples "we can identify stuff" from "show this & not this". should the latter be determined by the UI though, and not a final pipeline step that marks certain symbols in the DB as "show/not show"?

Yeah, I agree that it should be determined by the UI. Currently, the logic for filtering out what symbols are shown are in the UI (in particular, at the line in `EntityAnnotationLayer.tsx` you commented on). Do you feel even more choices should be given to the UI about what to show or not?

ui/src/api/types.ts
andrewhead: @kyleclo I thought it was more in the spirit of your suggestion to add the note about extensions to `data-processing/common/types.py`, where one is likely to add new names of types of symbols. Does my change to the documentation there satisfy your desire for documentation about extensions?

kyleclo: yea that's fine

ui/src/components/entity/EntityAnnotationLayer.tsx
kyleclo: yeah probably, but it seems good to do this for now.  maybe add a TODO for migrating to config at later date if it ever becomes necessary

PR #218, opened 2021-01-29
Pipeline: save space by deleting directories of compiled sources and …

andrewhead:
Addresses https://github.com/allenai/scholar/issues/26111

Machines that the pipeline runs on frequently run out of space, particularly for papers with lots of symbols. This PR seeks to remove the major source of out-of-space errors---namely, the creation of many copies of the TeX source directories and rastered images from a paper.

By default, the pipeline now deletes copies of TeX source directories and page rasters as soon as it is done using them. Now, at any given time, there should be only one set of copies of TeX sources and rastered images.

The reason all of these files were generated and stored in the first place was to aid with debugging of the pipeline. To allow us to continue debugging the pipeline, a flag `--keep-intermediate-files` has been added. When set, the copies of the TeX sources and the page rasters will not be deleted, so they can be inspected during debugging.

### DIFF
--- a/data-processing/common/commands/locate_entities.py
+++ b/data-processing/common/commands/locate_entities.py
@@ -13,23 +13,14 @@ from common.colorize_tex import ColorizedTex, colorize_entities
 from common.commands.base import ArxivBatchCommand
 from common.commands.compile_tex import save_compilation_result
 from common.commands.raster_pages import raster_pages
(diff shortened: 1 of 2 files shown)


aschokking (approved):
Looks nice and clean, thanks for the detailed write up and arg doc!

PR #219, opened 2021-01-30
[WIP] Expand TeX macros in math environments as a core stage of the pipeline

andrewhead:
(Not yet ready for review)


andrewhead:
Andrew's personal notes (mostly old).

Extra steps:
- [ ] Handle macros that add expansion to pushback and use them as arguments to other macros (even if just ignoring this macros altogether)
- [ ] ~Add space before a macro when it follows another macro (so that its letters do not get merged into the earlier macro)---see if it is possible to look backwards in LaTeXML to the category code of the last character~
- [x] Do not expand 'num' (robust fix---do not expand macros that are not defined in project files. more brittle (but faster) fix---do not expand 'num') (I think we do not want to expand nested macros that were not defined in project files, either, e.g., operatorname)
  - Robus fix would also address not trying to expand '\%' and '\quad' and '\dots' and '\left' and '\end{align}', '\]' and '\textnormal'

What is going on with (in paper 1811.12359):
*  the `\rz` expansion? Maybe it is an issue of reading expansion tokens that are not part of the expansion.
* the '\supp' expansion? Continually expands to the expansion of the control sequence _after_ it, and keeps the next control sequence from expanding
* the '\E' at the beginning of line 1109 (should have been expanded)
* missing '\KL' expansion in the middle of the line of line 1127 (also not expanded at the beginning on line 1229)
* missing '\rz' expansion on line 1176
* Expansion of '\argmax' on line 1178. DeclareMathOperator probably needs a more sophisticated treatment.

And also in paper 1906.08632:
* "\Emu" on the start of line 892.

Continue from line 1127 in the diff

andrewhead:
With regard to the upcoming task: 

> If a macro expands to include another macro and an argument to that macro, there is a chance the argument may be duplicated. This case needs a test case and may require updates to our macro expansion code.

Here are some personal notes with regard to how to implement the fix:

```
    # When getting the argument, check if it appears in the TeX or not (using source name).
    # If not, then find the first argument _after_ the control sequence with a matching
    # object ID, and remove it.
```

data-processing/common/commands/expand_macros.py
```diff
@@ -0,0 +1,191 @@
...
+        absolute_tex_path = os.path.realpath(os.path.abspath(tex_path))
+
+        # Remove '.tex' extension from the name of the TeX file.
+        tex_name = re.sub(r"\.tex$", "", item.tex_file.path)
+
+        latexml_bin = os.path.join(
+            os.path.abspath(os.path.join("perl", "latexml", "bin", "latexml"))
+        )
+        if not os.path.exists(latexml_bin):
+            logging.warning(  # pylint: disable=logging-not-lazy
```
aschokking, line 94: Naively I'd assume this should cause the pipeline to fail if the box it's running on isn't setup correctly? If someone didn't want to run this step they could pull it out of the pipeline themselves right?

data-processing/common/commands/expand_macros.py
```diff
@@ -0,0 +1,191 @@
+import logging
```
aschokking, line 1: consider putting a blurb at the top here, maybe using info from your pr description, about what macros are and why they need to be expanded

data-processing/common/expand_macros.py
```diff
@@ -0,0 +1,492 @@
...
+    arguments (e.g., all of '\\c{X}' for the macro '\\c' defined as '\\def\\c#1{\\mathcal{C}}')).
+
+    For 'start_line' and 'end_line', the first line in a file is 1 (not 0). For 'start_col' and
+    'end_col', the first character on the line is at col 0 (not 1, in contrast to the line number).
+    """
+
+    expansion: bytes
+
+
+def detect_expansions(
```
aschokking, line 228: totally optional, but this is quite a beefy function, I wonder if it would be possible to break it up into smaller chunks so the overall shape of what it's doing could be seen at a glance and then you could dive into the broken out functions to see details of how a particular type is processed for instance?

data-processing/common/expand_macros.py
```diff
@@ -0,0 +1,492 @@
...
+        line_start += len(line)
+
+    # Apply expansions (in reverse order)
+    contents_copy = bytes(contents)
+    for ei, expansion in enumerate(expansions_sorted):
+        macro_start = start_offsets[ei]
+        macro_end = end_offsets[ei]
+        if macro_start == -1 or macro_end == -1:
+            continue
+        # Do not replace macro if the text to be expanded does not start with the macro name.
```
aschokking, line 480: would this indicate an error we should warn about or can come up normally?

data-processing/perl/setup.sh
```diff
@@ -0,0 +1,9 @@
+# Fetch sources for AI2 clone of LaTeXML.
+git clone https://github.com/allenai/latexml latexml
```
aschokking, line 2: I was just wondering if we'd ever have a breaking change merge to our fork of latexml that then might then break the processor until it was updated?

data-processing/tests/test_expand_macros.py
```diff
@@ -0,0 +1,247 @@
+from typing import List
+
+from common.expand_macros import Expansion, detect_expansions, expand_macros
+
+
+def to_bytes(*lines: List[str]) -> bytes:
+    return "\n".join(lines).encode()
+
+
+def test_detect_simple_macro():
```
aschokking, line 10: yay tests! ❤️

data-processing/perl/setup.sh
```diff
@@ -0,0 +1,9 @@
+# Fetch sources for AI2 clone of LaTeXML.
```
aschokking, line 1: does `data-processing/Dockerfile` need to get updated to call this script?

aschokking (commented):
There's a lot going on here, exciting to see this coming together! Appreciate all the detailed comments and unit tests! 

I had a few smaller questions/comments for you.

data-processing/common/expand_macros.py
andrewhead: You are right, this would be a good thing to warn about (I would be interested in knowing if this issue ever occurs---right now I am only speculating that it could). I will add a warning 👍

data-processing/common/expand_macros.py
andrewhead: I was thinking about this! I will take a stab at breaking this into a few functions, probably as part of a class (so that state can be shared). If I can do it in a time-boxed interval, we can review it as an alternative.

data-processing/common/commands/expand_macros.py
andrewhead: Good thinking!

data-processing/perl/setup.sh
andrewhead: This is a great point. It could certainly happen.

I wonder what types of merge workflows could prevent this from happening. Maybe an explicit tag of the LaTeXML repo is cloned, so that there is no unexpected changes in the behavior of the processor as the LaTeXML branch updates?

data-processing/tests/test_expand_macros.py
andrewhead: :-) 

A haiku
Unit tests are like
hard timber walls of a house
on a blust'ry day.

data-processing/perl/setup.sh
andrewhead: It sure does! Great catch.

data-processing/common/commands/expand_macros.py
andrewhead: Yes, this is a good point. I will add an exception in the style of those you and Chris have been adding lately.

data-processing/perl/setup.sh
aschokking: yup explicitly pinning the tag/commit of the latexXML repo here at cloning time is probably good enough 👍

data-processing/tests/test_expand_macros.py
aschokking: Haikus are easy
But sometimes they don't make sense
Refrigerator

andrewhead: 👏 😆

data-processing/common/expand_macros.py
andrewhead: @aschokking hopefully the changes below (from monolithic method to class with half a dozen methods) makes the code a bit easier to read / navigate by chunking it up. Want to take a look?

aschokking: Thanks Andrew, `detect_expansions ` is definitely a lot easier to read now!

data-processing/perl/setup.sh
andrewhead: good thinking!

PR #220, opened 2021-01-31
HEDDEx (Definition extractor) new model integration (v3)

nuwandavek:
## Changes 
- Evaluation of term-def, abbr-exp, sym-nick by their respective models
- Changes in the model trainer to store the prediction_type
- Change DefinitionDetectionModel to store all 3 models together and predict accordingly

## ToDo before merging
- Replace the placeholder model download links with the correct ones
- Retain necessary nickname code
- Delete commented legacy code (abbreviation and nicknames)
- Update filter heuristics for abbr-exp and sym-nick (heuristic_filters function in trainer)
- Update tests top reflect (and test) the above changes

### DIFF
--- a/data-processing/entities/definitions/commands/detect_definitions.py
+++ b/data-processing/entities/definitions/commands/detect_definitions.py
@@ -153,6 +153,18 @@ def consolidate_keyword_definitions(
                     sum(definition_confidences) / len(definition_confidences)
                 )
             term_ranges, definition_ranges = [], []
+        
+    if len(term_ranges) > 0:
+        terms.append(term_ranges)
+        term_confidence_list.append(
+            sum(term_confidences) / len(term_confidences)
+        )
+    if len(definition_ranges) > 0:
+        definitions.append(definition_ranges)
+        definition_confidence_list.append(
+            sum(definition_confidences) / len(definition_confidences)
+        )
+    term_ranges, definition_ranges = [], []
 
     # Match pairs of term and definitions sequentially ( O T O O D then (T, D)). This
     # does not handle multi-term / definitions pairs.
@@ -664,6 +676,7 @@ class DetectDefinitions(
                         symbol_texs = get_symbol_texs(
                             s.legacy_definition_input, s.with_formulas_marked
                         )
+                        #--- This is V2 code for getting symbol_nickname_pairs and abbreviation_pairs---
                         # if symbol_texs is None:
                         #     symbol_nickname_pairs = []
                         # else:
@@ -679,6 +692,7 @@ class DetectDefinitions(
                         #     sentence_features["tokens"],
                         #     model.nlp,
                         # )
+                        #--- end ---
 
                         # Only process slots when they include both 'TERM' and 'DEFINITION'.
                         if "TERM" not in termdef_sentence_slots or "DEF" not in termdef_sentence_slots:
--- a/data-processing/entities/definitions/commands/detect_definitions.py
+++ b/data-processing/entities/definitions/commands/detect_definitions.py
@@ -113,11 +113,12 @@ def get_symbol_texs(
     return dict(zip(symbol_starts, symbol_texs))
 
 
-def consolidate_term_definitions(
+def consolidate_keyword_definitions(
     text: str,
     tokens: List[str],
     slot_predictions: List[str],
     slot_prediction_confidences: List[float],
+    prediction_type : str
 ) -> List[TermDefinitionPair]:
 
     # Make index from tokens to their character positions.
@@ -134,13 +135,13 @@ def consolidate_term_definitions(
     for (slot_label, slot_confidence, range_) in zip(
         slot_predictions, slot_prediction_confidences, ranges
     ):
-        if slot_label == "TERM":
+        if slot_label == 'TERM':
             term_ranges.append(range_)
             term_confidences.append(slot_confidence)
-        if slot_label == "DEF":
+        if slot_label == 'DEF':
             definition_ranges.append(range_)
             definition_confidences.append(slot_confidence)
-        if slot_label == "O":
+        if slot_label == 'O':
             if len(term_ranges) > 0:
                 terms.append(term_ranges)
                 term_confidence_list.append(
@@ -183,8 +184,14 @@ def consolidate_term_definitions(
             max([definition_range.end for definition_range in definition_rangelist]) + 1
         )
 
-        term_type = "symbol" if "SYMBOL" in text[term_start:term_end] else "term"
-        definition_type = "definition"
+        all_definition_types = {
+        'TERM-DEF' : ['term','definition'],
+        'ABBR-EXP' : ['abbreviation','expansion'],
+        'SYM-NICK' : ['symbol','nickname']
+        }
+
+        term_type = all_definition_types[prediction_type][0]
+        definition_type = all_definition_types[prediction_type][1]
 
         pair = TermDefinitionPair(
             term_start=term_start,
@@ -582,7 +589,8 @@ class DetectDefinitions(
             return
 
         # Load the pre-trained definition detection model.
-        model = DefinitionDetectionModel()
+        prediction_types =  ['TERM-DEF', 'ABBR-EXP', 'SYM-NICK']
+        model = DefinitionDetectionModel(prediction_types)
 
         definition_index = 0
         features = []
@@ -619,19 +627,33 @@ class DetectDefinitions(
 
                     # Detect terms and definitions in each sentence with a pre-trained definition
                     # extraction model, from the featurized text.
-                    intents, slots, slots_confidence = model.predict_batch(
-                        cast(List[Dict[Any, Any]], features)
+                    termdef_intents, termdef_slots, termdef_slots_confidence = model.predict_batch(
+                        cast(List[Dict[Any, Any]], features), 'TERM-DEF'
                     )
+                    abbrexp_intents, abbrexp_slots, abbrexp_slots_confidence = model.predict_batch(
+                        cast(List[Dict[Any, Any]], features), 'ABBR-EXP'
+                    )
+                    symnick_intents, symnick_slots, symnick_slots_confidence = model.predict_batch(
+                        cast(List[Dict[Any, Any]], features), 'SYM-NICK'
+                    )
+                    
 
                     # Package extracted terms and definitions into a representation that's
                     # easier to process.
                     for (
                         s,
                         sentence_features,
-                        intent,
-                        sentence_slots,
-                        sentence_slots_confidence,
-                    ) in zip(sentences, features, intents, slots, slots_confidence):
+                        termdef_intent,
+                        termdef_sentence_slots,
+                        termdef_sentence_slots_confidence,
+                        abbrexp_intent,
+                        abbrexp_sentence_slots,
+                        abbrexp_sentence_slots_confidence,
+                        symnick_intent,
+                        symnick_sentence_slots,
+                        symnick_sentence_slots_confidence,
+                    ) in zip(sentences, features, termdef_intents, termdef_slots, termdef_slots_confidence,abbrexp_intents, 
+                            abbrexp_slots, abbrexp_slots_confidence, symnick_intents, symnick_slots, symnick_slots_confidence):
                         # Extract TeX for each symbol from a parallel representation of the
                         # sentence, so that the TeX for symbols can be saved.
                         # Types of [term and definition] pairs.
@@ -642,37 +664,60 @@ class DetectDefinitions(
                         symbol_texs = get_symbol_texs(
                             s.legacy_definition_input, s.with_formulas_marked
                         )
-                        if symbol_texs is None:
-                            symbol_nickname_pairs = []
-                        else:
-                            symbol_nickname_pairs = get_symbol_nickname_pairs(
-                                s.legacy_definition_input,
-                                sentence_features["tokens"],
-                                sentence_features["pos"],
-                                symbol_texs,
-                            )
+                        # if symbol_texs is None:
+                        #     symbol_nickname_pairs = []
+                        # else:
+                        #     symbol_nickname_pairs = get_symbol_nickname_pairs(
+                        #         s.legacy_definition_input,
+                        #         sentence_features["tokens"],
+                        #         sentence_features["pos"],
+                        #         symbol_texs,
+                        #     )
 
-                        abbreviation_pairs = get_abbreviations(
-                            s.legacy_definition_input,
-                            sentence_features["tokens"],
-                            model.nlp,
-                        )
+                        # abbreviation_pairs = get_abbreviations(
+                        #     s.legacy_definition_input,
+                        #     sentence_features["tokens"],
+                        #     model.nlp,
+                        # )
 
                         # Only process slots when they include both 'TERM' and 'DEFINITION'.
-                        if "TERM" not in sentence_slots or "DEF" not in sentence_slots:
+                        if "TERM" not in termdef_sentence_slots or "DEF" not in termdef_sentence_slots:
                             term_definition_pairs = []
                         else:
-                            term_definition_pairs = consolidate_term_definitions(
+                            term_definition_pairs = consolidate_keyword_definitions(
                                 s.legacy_definition_input,
                                 sentence_features["tokens"],
-                                sentence_slots,
-                                sentence_slots_confidence,
+                                termdef_sentence_slots,
+                                termdef_sentence_slots_confidence,
+                                "TERM-DEF"
+                            )
+
+                        if "TERM" not in abbrexp_sentence_slots or "DEF" not in abbrexp_sentence_slots:
+                            abbreviation_expansion_pairs = []
+                        else:
+                            abbreviation_expansion_pairs = consolidate_keyword_definitions(
+                                s.legacy_definition_input,
+                                sentence_features["tokens"],
+                                abbrexp_sentence_slots,
+                                abbrexp_sentence_slots_confidence,
+                                "ABBR-EXP"
+                            )
+                        
+                        if "TERM" not in symnick_sentence_slots or "DEF" not in symnick_sentence_slots:
+                            symbol_nickname_pairs = []
+                        else:
+                            symbol_nickname_pairs = consolidate_keyword_definitions(
+                                s.legacy_definition_input,
+                                sentence_features["tokens"],
+                                symnick_sentence_slots,
+                                symnick_sentence_slots_confidence,
+                                "SYM-NICK"
                             )
 
                         pairs = (
                             term_definition_pairs
                             + symbol_nickname_pairs
-                            + abbreviation_pairs
+                            + abbreviation_expansion_pairs
                         )
 
                         for pair in pairs:
@@ -768,7 +813,7 @@ class DetectDefinitions(
                                 text=pair.definition_text,
                                 context_tex=sentence.context_tex,
                                 sentence_id=sentence.id_,
-                                intent=bool(intent),
+                                # intent=bool(intent),
                                 confidence=definition_confidence,
                             )
                             definitions[definition_id] = definition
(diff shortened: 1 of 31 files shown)


data-processing/entities/definitions/model/model/modeling_joint_roberta.py
```diff
@@ -49,23 +53,26 @@ def __init__(self, config: RobertaConfig, args: Any, intent_label_lst: List[str]
...
 
-        if args.use_crf:
-            self.crf = CRF(num_tags=self.num_slot_labels, batch_first=True)
+        for pred_type in tasks:
+            self.slot_classifiers[pred_type] = SlotClassifier(hidden_size, self.num_slot_labels_dict[pred_type], args.dropout_rate)
+
+            if args.use_crf:
+                self.crfs[pred_type] = CRF(num_tags=self.num_slot_labels_dict[pred_type], batch_first=True)
+        
+        self.intent_classifiers = nn.ModuleDict(self.intent_classifiers)
```
andrewhead, line 65: out of curiosity, what does wrapping these attributes in moduledicts do?

data-processing/entities/definitions/model/model/modeling_joint_roberta.py
```diff
@@ -108,56 +115,56 @@ def forward(
...
+            intent_label_id = intent_label_ids[data_type]
+            if intent_label_id is not None:
+                if self.num_intent_labels_dict[data_type] == 1:
+                    intent_loss_fct = nn.MSELoss()
+                    intent_loss = intent_loss_fct(intent_logits[data_type].view(-1), intent_label_id.view(-1))
+                else:
+                    intent_loss_fct = nn.CrossEntropyLoss()
+                    intent_loss = intent_loss_fct(intent_logits[data_type].view(-1, self.num_intent_labels_dict[data_type]), intent_label_id.view(-1))
+                total_loss += intent_loss
+                total_loss_dict["intent_{}".format(data_type)] = intent_loss
```
andrewhead, line 141: have you heard of f strings (e.g., `f"intent_{data_type}"`)? They are a nifty Python feature that provides a subset of the string formatting functionality. Not necessary to change, just wanted to share.

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -619,19 +639,33 @@ def process(
 
                     # Detect terms and definitions in each sentence with a pre-trained definition
                     # extraction model, from the featurized text.
-                    intents, slots, slots_confidence = model.predict_batch(
-                        cast(List[Dict[Any, Any]], features)
+                    termdef_intents, termdef_slots, termdef_slots_confidence = model.predict_batch(
```
andrewhead, line 642: Is there ever a circumstance where we will not want to just get everything from `model.predict_batch` for each of the three tasks at the same time? Could we just make model.predict_batch return predictions for all three tasks at once? (Mostly curious. I realize this could potentially require a big change, which might not be worth it.)

data-processing/tests/test_extract_definitions.py
```diff
@@ -5,15 +5,15 @@
...
+    consolidate_keyword_definitions,
     get_abbreviations,
     get_symbol_nickname_pairs,
 )
 
 
 @pytest.fixture(scope="module")
 def model():
-    model = DefinitionDetectionModel()
+    model = DefinitionDetectionModel(['TERM-DEF', 'ABBR-EXP', 'SYM-NICK'])
```
andrewhead, line 16: could these args just be hard-coded into the DefinitionDetectionModel? Is there a circumstance under which we would want to change the task names?

data-processing/tests/test_extract_definitions.py
```diff
@@ -84,6 +123,38 @@ def test_extract_nicknames_from_before_symbols():
     assert nickname1.term_text == "t"
     assert nickname1.definition_text == "timestep"
 
+# @pytest.mark.slow
+# def test_model_extracts_nickname_symbol_separated_by_colon(model: DefinitionDetectionModel):
+#     prediction_type = 'SYM-NICK'
+#     features = model.featurize("The agent acts with a policy : SYMBOL")
+#     intents, slots, _ = model.predict_batch([features], prediction_type)
+#     assert intents[0]
+#     assert slots[0] == ["O", "O", "O", "O", "O", "DEF", "O", "TERM"]
```
andrewhead, line 132: Is this one of the buggy tests with the current model? I think for such tests, our approach should be to assign a tag to skip it, instead of commenting it out.

data-processing/tests/test_extract_definitions.py
```diff
@@ -137,6 +240,34 @@ def test_extract_nicknames_symbols_parentheses():
...
+def test_model_extracts_nickname_symbol_filter(model: DefinitionDetectionModel):
+    prediction_type = 'SYM-NICK'
+    features = model.featurize("The agent acts with SYMBOL SYMBOL")
+    intents, slots, _ = model.predict_batch([features], prediction_type)
+    assert intents[0] == 0
+    assert slots[0] == ["O", "O", "O", "O", "O", "O"]
+
+
+@pytest.mark.slow
+def test_extract_symbol_nickname_pairs_case_4():
```
andrewhead, line 253: nit pick: give test more descriptive name indicating the specific behavior being tested

data-processing/tests/test_extract_definitions.py
```diff
@@ -159,6 +290,14 @@ def test_extract_nicknames_symbols_filter():
     assert len(symbol_nickname_pairs) == 0
 
 
+# @pytest.mark.slow
+# def test_model_extracts_nickname_after_symbol(model: DefinitionDetectionModel):
+#     prediction_type = 'SYM-NICK'
+#     features = model.featurize("The architecture consists of SYMBOL dense layers trained with SYMBOL learning rate")
+#     intents, slots, _ = model.predict_batch([features], prediction_type)
+#     assert intents[0]
+#     assert slots[0] == ["O", "O", "O", "O", "TERM", "DEF", "DEF", "O", "O", "TERM", "DEF", "DEF"]
```
andrewhead, line 299: Lots of commented out tests, will these be removed / fixed before merge?

data-processing/tests/test_extract_definitions.py
```diff
@@ -219,6 +366,41 @@ def test_extract_nickname_for_th_index_pattern():
...
+            ("CNN", "TERM", 1.0),
+            (")", "O", 1.0),
+            ("based", "O", 1.0),
+            ("architecture", "O", 1.0),
+        ]
+    )
+
+    abbreviation_expansion_pairs = consolidate_keyword_definitions(
+        text, tokens, slots, confidences, prediction_type
+    )
```
andrewhead, line 400: @nuwandavek this part of the test feels more like a test of `consolidate_keyword_definitions` than it does of the model. In fact, the model is not even getting invoked here. Do we even need this test? Maybe just make a couple of dedicated tests for `consolidate_keyword_definitions` and call it a day?

data-processing/tests/test_extract_definitions.py
```diff
@@ -244,6 +426,14 @@ def test_extract_abbreviation_acronym(model):
...
+@pytest.mark.slow
+def test_model_extract_abbreviation_shortened_word(model: DefinitionDetectionModel):
+    prediction_type = 'ABBR-EXP'
+    features = model.featurize("We propose Conductive Networks (CondNets)")
+    intents, slots, _ = model.predict_batch([features], prediction_type)
+    assert intents[0]
+    assert slots[0] == ["O", "O", "DEF", "DEF", "O", "TERM", "O"]
+
 @pytest.mark.slow
 def test_extract_abbreviation_shortened_word(model):
```
andrewhead, line 438: this file is starting to get a bit long; can you take a quick look and see if any of the tests are redundant, and if so, remove them?

data-processing/entities/definitions/model/trainer.py
```diff
@@ -344,325 +284,126 @@ def evaluate_from_input(self, dataset: TensorDataset) -> Tuple[List[int], List[L
...
+
+        slot_label_map = {}
+        slot_label_to_idx_map = {}
+        for data_type, slot_labels in self.slot_label_dict.items():
+            slot_label_map[data_type] = {i: label for i, label in enumerate(slot_labels)}
+            slot_label_to_idx_map[data_type] = {label: i for i, label in enumerate(slot_labels)}
+        pos_label_map = {i: label for i, label in enumerate(self.pos_label_lst)}
+        pos_label_to_idx_map = {label: i for i, label in enumerate(self.pos_label_lst)}
+
+
```
andrewhead, line 367: nit pick: make sure to run code formatter on all this stuff before merge

andrewhead (commented):
Looking good, I left a set of comments.

data-processing/tests/test_extract_definitions.py
```diff
@@ -13,7 +13,7 @@
 
 @pytest.fixture(scope="module")
 def model():
-    model = DefinitionDetectionModel(['TERM-DEF', 'ABBR-EXP', 'SYM-NICK'])
+    model = DefinitionDetectionModel(['AI2020', 'DocDef2', 'W00'])
```
andrewhead, line 16: What does it mean for the tasks to have the names 'AI2020', 'DocDef2' and 'W00', when the tasks are named 'TERM-DEF', 'ABBR-EXP', and 'SYM-NICK' elsewhere?

andrewhead (changes requested):

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -642,37 +676,62 @@ def process(
...
+                        #         sentence_features["pos"],
+                        #         symbol_texs,
+                        #     )
+
+                        # abbreviation_pairs = get_abbreviations(
+                        #     s.legacy_definition_input,
+                        #     sentence_features["tokens"],
+                        #     model.nlp,
+                        # )
+                        #--- end ---
```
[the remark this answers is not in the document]
nuwandavek: I think Dongyeop was saying that we can have it till we have tested the v3 system on the frontend and made sure everything works fine? 
@dykang do correct me if I'm mistaken

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -642,37 +676,62 @@ def process(
...
+                        # Only process slots when they include both 'TERM' and 'DEFINITION'.
+                        if "TERM" not in termdef_sentence_slots or "DEF" not in termdef_sentence_slots:
+                            term_definition_pairs = []
                         else:
-                            symbol_nickname_pairs = get_symbol_nickname_pairs(
+                            term_definition_pairs = consolidate_keyword_definitions(
                                 s.legacy_definition_input,
                                 sentence_features["tokens"],
-                                sentence_features["pos"],
-                                symbol_texs,
```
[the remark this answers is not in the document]
nuwandavek: I think we still need this (as we swap placeholder text with the actual tex once we have all pairs of termdef, abbrexp, symnick

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -768,7 +827,7 @@ def process(
                                 text=pair.definition_text,
                                 context_tex=sentence.context_tex,
                                 sentence_id=sentence.id_,
-                                intent=bool(intent),
+                                # intent=bool(intent),
```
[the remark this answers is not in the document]
nuwandavek: I think intent was always redundant. We generate term-def pairs only when the intent is 1. We can get rid of this, unless it is required elsewhere downstream. Is it?

data-processing/entities/definitions/model/configuration.py
```diff
@@ -15,6 +15,12 @@ class ModelArguments:
             "help": "The model checkpoint for weights initialization. Leave None if you want to train a model from scratch."
         },
     )
+    prediction_type: Optional[str] = field(
+        default=None,
+        metadata={
+            "help": "The model prediction type. One of [TERM-DEF, ABBR-EXP, SYM-NICK]"
```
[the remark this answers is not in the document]
nuwandavek: Yes, this is correct, albeit this is only temporary. We are doing this because there is a separate model for each task (which in turn can potentially predict all 3 at once)

data-processing/entities/definitions/model/load_data.py
```diff
@@ -555,6 +555,80 @@ def load_and_cache_example(
...
+    processor = DefProcessor(args)
+
+    examples = []
+    raw = []
+    for d in data:
+        example = InputExample(
+            guid="one",
+            words=d["tokens"],
+            intent_label=1,  # fakse intent label
+            slot_labels=[1] * len(d["tokens"]),  # fake slot labels
```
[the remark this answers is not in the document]
nuwandavek: That is correct, these are fake labels (used for prediction when no labels exists, I think). I had this doubt too, but I think this is correct.

data-processing/entities/definitions/model/model/modeling_joint_roberta.py
nuwandavek: This is the pytorch API for having dynamic layers that are not explicitly defined. As you can see here, the layers are initialized based on the tasks.

data-processing/entities/definitions/model/model/modeling_joint_roberta.py
nuwandavek: Yes! I love fstrings! This was legacy code from the trainer repo that I just got lazy to change. 😛

data-processing/entities/definitions/commands/detect_definitions.py
nuwandavek: I think the expectation is for the 3 model paradigm to go away quickly. So, my changes mostly are to make sure that change will be very very easy (just deleting 2 of the 3 models)

data-processing/tests/test_extract_definitions.py
nuwandavek: These are cases where the tests currently fail. How should we handle that? Skip as you mentioned?

data-processing/tests/test_extract_definitions.py
nuwandavek: Yeah, that sounds good!

data-processing/tests/test_extract_definitions.py
nuwandavek: Will do!

data-processing/tests/test_extract_definitions.py
nuwandavek: It is (['AI2020', 'DocDef2', 'W00']) everywhere now. This is because the datasets are named that, and those names carry over to the model. But I agree that scholarphi should not have anything to do with the dataset names.
How should we handle this going forward @dykang ?

data-processing/entities/definitions/commands/detect_definitions.py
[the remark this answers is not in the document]
dykang: +1 keeping them for later usages. We may have a possibility to combine these off-the-shelf systems with neural systems in the future.

data-processing/tests/test_extract_definitions.py
dykang: Good point. There are some inquiries about our HEDDEx model, and I pointed them to see codes in this repo for now. In that regard, I believe we should keep the dataset names clear like as they are. But, I agree that they should be in a sort of configuration file in the future.

data-processing/entities/definitions/commands/detect_definitions.py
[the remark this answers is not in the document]
andrewhead: 🤔  I am not sure. Maybe run a search for intent across the project?

data-processing/entities/definitions/model/load_data.py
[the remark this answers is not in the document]
andrewhead: Let us give the fake labels a value that more accurately demonstrates that the data is invalid (e.g., -1 if it can be an int, or 'unknown' if a string)

data-processing/tests/test_extract_definitions.py
andrewhead: Yup!

andrewhead (commented):
Thank you for the updates, @nuwandavek! I left a few updated comments throughout. Have you managed to run the extractor on an arXiv paper and see that it extracts definitions of the three types?

data-processing/entities/definitions/model/load_data.py
```diff
@@ -555,6 +555,80 @@ def load_and_cache_example(
     )
     return dataset
 
+def load_and_cache_example_batch_raw(
```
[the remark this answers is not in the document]
nuwandavek: This returns the raw data as well (along with the tensor dataset). Raw data is needed in the filtering heuristics stage. I could have edited the old function to do this but it would break at places and would require a change to the old code as well.

data-processing/entities/definitions/model/load_data.py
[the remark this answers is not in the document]
nuwandavek: I'm using -1 for this

nuwandavek:
I've resolved all issues except for the formatting issue. Yes, I've run it on that sample short paper!

data-processing/entities/definitions/model/load_data.py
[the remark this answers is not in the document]
nuwandavek: Actually I cannot use -1 😱 
It triggers a CUDA error. The intent label has to be one of the 2 possible values (0 or 1), and the slot label has to be one of 6 possible values (0 to 5)

data-processing/entities/definitions/commands/detect_definitions.py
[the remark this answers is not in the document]
andrewhead: Sounds good! I left in the functions, though removed these commented-out function calls, and added some documentation indicating that the functions are now legacy (though left them in if we want to experiment with them in the future).

data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -926,11 +923,14 @@ def process(
...
         # Detect all other references to the defined terms. Detect references to textual
-        # terms, abbreviations, and symbols.
+        # terms and abbreviations. References to symbols need not be found here; they
+        # will be detected automatically in the symbol extraction code.
         term_index = 0
 
         for tex_path, file_contents in item.tex_by_file.items():
-            term_extractor = PhraseExtractor(term_phrases + abbreviations + symbol_nicks)
+            term_extractor = PhraseExtractor(
+                term_phrases + abbreviations
```
nuwandavek, line 932: @andrewhead Is there a reason we don't want symbol-nick definitions to get aggregated as well?

andrewhead: @nuwandavek yeah and great question! (I explain in a bit more detail in the comment just above this block)---all symbol usages will get detected automatically in the symbol extraction part of the pipeline, making detection of references to symbols in this part of the code redundant.

data-processing/entities/definitions/model/trainer.py
```diff
@@ -344,325 +284,122 @@ def evaluate_from_input(self, dataset: TensorDataset) -> Tuple[List[int], List[L
...
-                    inputs["intent_label_ids"].detach().cpu().numpy(),
-                    axis=0,
-                )
+            # Predict intent
+            intent_probs = {}
+            for data_type, logits in intent_logits.items():
+                intent_probs[data_type] = torch.softmax(logits, dim=1).detach().cpu().numpy()
+                gold_intent_label_dict[data_type] = inputs["intent_label_ids"][data_type].detach().cpu().numpy()
+                intent_preds[data_type] = logits.detach().cpu().numpy()
+                intent_conf[data_type] = intent_probs[data_type][:,1]
```
andrewhead, line 320: Hi @dykang and @nuwandavek! These two lines are overwriting the `intent_preds` and `intent_conf` arrays with each batch processed, instead of appending to them. The same might also happening for the other dictionaries of: gold_slot_labels_ids, slot_preds, slot_conf, and gold_intent_label_dict.

The result is that if someone passes in a batch of > 16 examples, they will get `NUM_EXAMPLES % 16` predictions. Could one of you take a look at this and offer a fix?

andrewhead (changes requested):

nuwandavek:
@andrewhead  : The fix works! You're right. We were actually assuming that the batch size is the same in both places

andrewhead (approved):
I have verified that the fix works on my machine. Thank you, Vivek!

PR #223, opened 2021-02-02
Change papers view sort to be by arxiv id desc

aschokking:
Ref https://github.com/allenai/scholar/issues/26501
So recent papers are on top. Previous sort was by total entity count.

After change:
image


kyleclo (approved):
lgtm

andrewhead (commented):
Excellent! lgtm 👍

PR #224, opened 2021-02-03
Change entity quality check to look for citations (instead of any entity)

aschokking:
Ref https://github.com/allenai/scholar/issues/26207

Since we might have useless on their own entities like sentences we wanted to tighten up this check we use to decide when to show the reader link on the s2 side.


mjlangan (approved):
👍

PR #225, opened 2021-02-03
Update root readme: Add little quickstart blurb

aschokking:
It took me longer than it should have to get the ui + api working locally at first. This was because I got lost in the different READMEs and what the ui server was or wasn't doing. This might help the next person get rolling a little faster 🤞

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,11 @@ npm install
 npm start
 ```
 
-However, if you want to run the api locally as well you'll have to:
-- Do the 1 time setup instructions for the api detailed in [api/README.md#one-time-setup](api/README.md#one-time-setup)
-- Run the api server in /api via `npm start`
-- In another terminal, run the reader app via `npm run dev`
+See [ui/README.md](ui/README.md) for more details on how to use the reader.
+
+However, if you want to run the *api* locally as well you'll have to:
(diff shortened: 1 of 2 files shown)


andrewhead:
Lgtm! Maybe direct readers to the `ui/README.md` as well, which has additional context about what to expect to see when the reader app launches and building the app?

README.md
```diff
@@ -16,6 +16,23 @@ papers and bounding boxes for those entities.
...
+In order to run the reader app locally (pointing to the production api) you only need to do 
+
+```bash
+cd ui
+npm install 
+npm start
+```
+
+However, if you want to run the api locally as well you'll have to:
+- Do the 1 time setup instructions for the api detailed in [api/README.md#one-time-setup](api/README.md#one-time-setup)
```
andrewhead, line 30: Readers should also do the `npm install` in the API directory before following the instructions in one-time setup

andrewhead (approved):

README.md
aschokking: yeah that's fair, I was trying to balance how much detail to put in here but you're right that one is worth including

aschokking: Changed this up so the api/README.md covered the one time setup in the right spot and just referenced those instructions as a whole

PR #226, opened 2021-02-03
Raise exception if references not found on s2

aschokking:
Ref https://github.com/allenai/scholar/issues/26519

When we try to process a paper before S2 has it's references available it means the final result won't have citations processed (which are a core part of the scholarPhi value prop), in which case we won't even make the reader link available on the PDP.

This PR proposes to fail pipeline jobs where the paper's references cannot be found on the S2 public api. The short-run goal is to mark these jobs as failed so we have a more accurate picture of our overall success rate. These should happen less frequently as well now that we're delaying queuing up arxiv papers by 1 day.

Medium term we should try to communicate the error was retryable so the outside pipeline can retry the job later. @rreas I'm interested in your thoughts on that communication medium, exit code?

Longer term we should split this out into it's own stage that runs independently but that's a much larger task.

### DIFF
--- a/data-processing/common/commands/fetch_s2_data.py
+++ b/data-processing/common/commands/fetch_s2_data.py
@@ -67,15 +67,14 @@ class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
                 references.append(reference)
             
             if not references:
+                # References are required to process citations, mark job as failed
                 raise S2ReferencesNotFoundException()
 
             s2_metadata = S2Metadata(s2_id=data["paperId"], references=references)
             logging.debug("Fetched S2 metadata for arXiv paper %s", item)
             yield s2_metadata
         else:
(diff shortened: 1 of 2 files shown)


data-processing/common/commands/fetch_s2_data.py
```diff
@@ -13,6 +13,16 @@
 """ Time to wait between consecutive requests to S2 API. """
 FETCH_DELAY = 3  # seconds
 
+class S2PaperNotFoundException(Exception):
```
rreas, line 16: Good idea to use custom error class. Consider introducing a parent Exception class like S2Error that these inherit from. The parent lets consumers of this class catch all S2-based errors without having to explicitly enumerate them.

rreas (approved):

andrewhead (approved):
@aschokking this looks great! Thanks for the test of the exception 👍

PR #227, opened 2021-02-04
Pipeline: Convenience script for fetching and unpacking sources for single arXiv ID

andrewhead:
I frequently want to download and unpack the TeX sources for a single arXiv paper quickly when I am debugging issues of missing entities. The current workflow to do so requires calling a subset of the pipeline via `python scripts/run_pipeline.py --arxiv-ids <ARXIV_ID> --commands fetch-arxiv-ids` and then looking at the unpacked sources in `data/**-sources/<ARXIV_ID>`.

This was feeling a bit cumbersome, so I wrote a convenience script for fetching just a single project and unpacking it to `./tmp/<ARXIV_ID>`. The new invocation is:

```bash
python scripts/fetch_arxiv_sources.py <ARXIV_ID>
```

I am submitting for review as I would be happy to merge it to main if it was thought that this would be of general interest. That said, this might be an optimization for just one person (me) ^_^ Would love to hear your thoughts---does this belong in the ScholarPhi repo?

### DIFF
--- a/data-processing/scripts/fetch_arxiv_sources.py
+++ b/data-processing/scripts/fetch_arxiv_sources.py
@@ -1,0 +1,46 @@
+import argparse
+import os
+
+from common import directories
+from common.fetch_arxiv import fetch_from_arxiv
+from common.unpack import unpack, unpack_archive
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        "Fetch and unpack sources for a single arXiv paper."
+    )
+    parser.add_argument(
+        "arxiv_id",
+        help="The arXiv ID for a paper. May include version number (i.e., 'v1', 'v2', etc.)",
+    )
+    parser.add_argument(
+        "--output-dir",
+        help="Directory into which the arXiv sources will be fetched.",
+        default="tmp",
+    )
+
+    args = parser.parse_args()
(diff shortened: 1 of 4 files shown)


data-processing/scripts/fetch_arxiv_sources.py
```diff
@@ -0,0 +1,46 @@
...
+    parser = argparse.ArgumentParser(
+        "Fetch and unpack sources for a single arXiv paper."
+    )
+    parser.add_argument(
+        "arxiv_id",
+        help="The arXiv ID for a paper. May include version number (i.e., 'v1', 'v2', etc.)",
+    )
+    parser.add_argument(
+        "--output-dir",
+        help="Directory into which the arXiv sources will be fetched.",
```
aschokking, line 18: consider noting the sources will be stored in a  arxiv id subfolder of this location (so no need to make your own outputdir the arxiv id)

aschokking (approved):
Seems generally useful, I'm all for including it 👍

data-processing/scripts/fetch_arxiv_sources.py
andrewhead: Good thinking! Will do 👍

PR #229, opened 2021-02-08
Use bearer token check to restrict write-apis

aschokking:
Fixes https://github.com/allenai/scholar/issues/25340

After this change, calls to the 2 apis that allow directly modifying data in the db will require an access token be set in the request. I've updated the UI app so it will try to read the token from a cookie named `readerAdminToken` that people can set locally. We can share the token value in slack or something.

If directly hitting the api, the token can be set in the query string like:
`?access_token={token}`

Or via headers like:
`Authorization: Bearer {token}`

The token checking is handled by an off the shelf hapi plugin I've added to the stack called `hapi-auth-bearer-token`.

The token is stored in the skiff secrets store and can be updated there at will. When running locally a static token value of `"local"` is used.

## Longer term

Longer term we could create users with passwords in the db or piggy back on google oauth or something but that's a little trickier to do in a way that this open source projects remains accessible for others to run locally.

### DIFF
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -11,6 +11,7 @@ nconf
       port: 5432,
       schema: "public",
     },
+    adminToken: "local"
   });
 
 process.on("unhandledRejection", (err) => {
--- a/api/src/index.ts
+++ b/api/src/index.ts
@@ -11,7 +11,6 @@ nconf
       port: 5432,
       schema: "public",
     },
-    adminToken: "local"
   });
 
 process.on("unhandledRejection", (err) => {
--- a/ui/src/api/api.ts
+++ b/ui/src/api/api.ts
@@ -1,4 +1,5 @@
 import axios, { AxiosResponse } from "axios";
+import cookie from "cookie";
 import { addLibraryEntryUrl, userInfoUrl } from "./s2-url";
 import { UserInfo, UserLibrary } from "../state";
 import {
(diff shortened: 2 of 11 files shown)


api/src/index.ts
```diff
@@ -11,6 +11,7 @@ nconf
       port: 5432,
       schema: "public",
     },
+    adminToken: "local"
```
mjlangan, line 14: With it being initialized this way, does this mean that if the prod config somehow loses the admin token entry, the public site will fall back to `local`? It might be better to have those APIs more or less disabled unless the admin token is configured in the secrets file, just to be safe.

ui/src/api/api.ts
```diff
@@ -12,6 +13,12 @@ import {
   PaperIdWithEntityCounts,
 } from "./types";
 
+const token = cookie.parse(document.cookie)["readerAdminToken"];
```
andrewhead, line 16: How does one set this adminToken locally? Would they open the cookie editor in the browser dev tools? I wonder if it we possible to let us set the reader admin token through environment variables?

andrewhead, line 16: How does one set this adminToken locally? Would they open the cookie editor in the browser dev tools? I wonder if it we possible to let us set the reader admin token through environment variables?

api/src/index.ts
aschokking: That's a good point. The local story gets a little more gnarly then but I guess we can put it in part of the 1 time setup that people set the value in their config.

ui/src/api/api.ts
aschokking: > How does one set this adminToken locally? Would they open the cookie editor in the browser dev tools?

Right setting it in the browser dev tools is the easiest place to start. Can just add a new cookie by hand with key `readerAdminToken` and value the secret value we can share over slack or what have you.

> I wonder if it we possible to let us set the reader admin token through environment variables?

Do you mean setting it in the server or the browser? If the latter, browsers don't have access to system environment variables.

api/src/index.ts
aschokking: Updated

mjlangan (approved):
LGTM 👍

PR #231, opened 2021-02-11
TeX: Fix up exception from parsing symbols with scripts

andrewhead:
In a recent change to the symbol parsing logic, I introduced several bugs that manifested once applying the pipeline to papers that make heavy use of subscripted / superscripted symbols (which seems to be a lot of papers!)

This PR fixes those bugs as discovered when processing paper `2002.06755v1`. The paper now processes successfully with symbols extracted.


aschokking (approved):
Looks good, thanks for the plentiful comments

aschokking:
We should probably do another pipeline deploy after this goes in

aschokking:
Talked to Andrew offline about merging this on his behalf

PR #232, opened 2021-02-16
Restore native link functionality

aschokking:
Fixes https://github.com/allenai/scholar/issues/26381

This puts the native link functionality back in to the reader so things like links to other pages or internal links to headers work again.

@andrewhead I wanted to understand if there were overlapping issues you ran into before where these link annotations and our own would fight, do we need to bump these down in the z-order or something for that case or do you think this'll be okay?

Before:
Example from the [PAWLS paper](https://scholarphi.semanticscholar.org/?file=https://arxiv.org/pdf/2101.10281v1.pdf) where this links to the app aren't clickable today:
image

After: 
They are clickable and have a highlight hover:
image

### DIFF
--- a/ui/src/style/pdf-viewer.less
+++ b/ui/src/style/pdf-viewer.less
@@ -29,9 +29,8 @@
  * border styles. See the pdf.js code setting border styles here:
  * https://github.com/mozilla/pdf.js/blob/474fe1757e654e67d20a53774edb293aa073ce0b/src/display/annotation_layer.js#L19
  */
-.annotationLayer .linkAnnotation {
+.annotationLayer .linkAnnotation > a {
   border: none !important;
-  display: none !important;
 }
 
 /**


ui/src/style/pdf-viewer.less
```diff
@@ -29,9 +29,8 @@
  * border styles. See the pdf.js code setting border styles here:
  * https://github.com/mozilla/pdf.js/blob/474fe1757e654e67d20a53774edb293aa073ce0b/src/display/annotation_layer.js#L19
  */
-.annotationLayer .linkAnnotation {
+.annotationLayer .linkAnnotation > a {
```
mjlangan, line 32: Unless this is an unintended bug in the CSS overrides, a comment here would be useful so we know that this is necessary for enabling pdf.js's default link handling.

mjlangan (approved):
Seems a little spooky/magical that a CSS change makes the pdf.js viewer's links clickable again, but works for me.

ui/src/style/pdf-viewer.less
aschokking: Good question, the `display: none` was actually what was causing the click targets from pdf.js to go away, so this PR is removing the hack that was preventing the default link handling. The current comment that is here does actually apply well to the border override that we still want.

Given there isn't anything about native link handling involved anymore, I wasn't sure if it made sense to comment about it but interested in thoughts.

mjlangan: Ah, I see, so we're removing something that was intentionally disabling it. Less important to me in this case, then.

andrewhead (approved):
Looks good! @aschokking yes, there will be overlapping issues. In many papers, citations will have hyperlinks that take a reader to the end of the paper that pdf.js will render in a similar place as our own links.

One of the challenges of the overlapping bounding boxes is that, if the two boxes are slightly different, clicking on two different places in a citation could then have different interactive behavior. A long-term solution to this might be iterate over the ScholarPhi bounding boxes and the pdf.js bounding boxes, and disable those in pdf.js that overlap with ScholarPhi's.

An easier stopgap solution I think is to do as you mentioned, and make sure that ScholarPhi bounding boxes appear on top.

aschokking:
> One of the challenges of the overlapping bounding boxes is that, if the two boxes are slightly different, clicking on two different places in a citation could then have different interactive behavior. A long-term solution to this might be iterate over the ScholarPhi bounding boxes and the pdf.js bounding boxes, and disable those in pdf.js that overlap with ScholarPhi's.
> 
> An easier stopgap solution I think could be to do as you mentioned, and make sure that ScholarPhi bounding boxes appear on top.

@andrewhead do you happen to know of an example paper off the top of your head that has this overlap behavior I could use to test that out? I can do some random sampling but otherwise I think we might have to deploy it as is and wait for someone to report an issue on a specific paper?

PR #233, opened 2021-02-17
Log entity load time in heap

aschokking:
Fixes https://github.com/allenai/scholar/issues/26776

This PR logs the amount of time it takes to fetch all the entities/citations to heap. This will give us a helpful view on this as we make improvements.

Example request browser is making to heap with this data in it:
image


mjlangan (approved):
👍

PR #235, opened 2021-02-18
Get citations by default & add showAll

mjlangan:
This change allows the entities APIs to fetch only entities of specified types. Multiple types can be specified using a comma-separated list, or all can be fetched by specifying `all`. The default when this query param is not present is `["citation"]` so the UI will by default only request citations as it does not include a type param in the query string. The full experience can be "unlocked" by appending `&showAll=true` to the URL in your browser.


api/src/types/validation.ts
```diff
@@ -254,3 +255,22 @@ export const loadedEntity = Joi.object({
   attributes: attributes.required(),
   relationships: relationships.required(),
 });
+
+export const apiEntityTypes = Joi.string().custom((raw: string, helpers) => {
+  // Special shortcut case for all entities
+  if (raw.toLowerCase() === 'all') {
```
aschokking, line 261: should 'all' be an exported const?

ui/src/ScholarReader.tsx
```diff
@@ -726,14 +727,28 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
...
+
   loadDataFromApi = async (): Promise<void> => {
     if (this.props.paperId !== undefined) {
       if (this.props.paperId.type === "arxiv") {
         this.setState({
           areCitationsLoading: true
         });
+        let getAllEntities = false;
+        const qs = this.parseQueryString();
+        console.log(qs);
```
aschokking, line 746: remove debug logging?

ui/src/api/api.ts
```diff
@@ -61,9 +61,12 @@ export async function getPapers(s2Ids: string[]) {
   return Promise.all(promises).then((responses) => responses.flat() as Paper[]);
 }
 
-export async function getEntities(arxivId: string) {
+export async function getEntities(arxivId: string, getAllEntities?: boolean) {
+  const params = getAllEntities ? {
+    type: 'all'
```
aschokking, line 66: make const?

aschokking (approved):
Looks good 👍

ui/src/ScholarReader.tsx
mjlangan: Whoopsies, thanks!

api/src/types/validation.ts
mjlangan: Seeing both your comments, yes 😄

api/src/api.ts
```diff
@@ -175,11 +176,16 @@ export const plugin = {
       path: "papers/{s2Id}/entities",
       handler: (request) => {
         const s2Id = request.params.s2Id;
-        return dbConnection.getEntitiesForPaper({ s2_id: s2Id });
+        // Runtime type-checked during validation.
```
andrewhead, line 179: Thank you for fixing this :-D

api/src/types/validation.ts
```diff
@@ -254,3 +255,23 @@ export const loadedEntity = Joi.object({
   attributes: attributes.required(),
   relationships: relationships.required(),
 });
+
+export const ENTITY_API_ALL = "all";
+export const apiEntityTypes = Joi.string().custom((raw: string, helpers) => {
```
andrewhead, line 260: @mjlangan it may be possible to get the same behavior here with [`Joi.alternatives()`](https://github.com/sideway/joi/blob/master/API.md#alternatives), where one alternative is the string `all`, and the other is a [`Joi.array()`](https://stackoverflow.com/a/42656623/2096369) where the items are of type string and the values are whitelisted to the entity types. Not at all necessary, though could turn this into a a couple of lines if I understand the logic correctly

ui/src/ScholarReader.tsx
```diff
@@ -726,14 +727,27 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
     });
   }
 
+  parseQueryString = (): qs.ParsedQs => {
```
andrewhead, line 730: nit pick: `index.tsx` file uses `query-string` library for similar functionality. Maybe we can use just one of these libraries or the other? https://github.com/allenai/scholarphi/blob/95ad49a46e914863d2873f19c2b9875bf4bbe94f/ui/src/index.tsx#L1

andrewhead (commented):
Overall, looking good! I am not quite finished taking a look (I will be in the early afternoon), though I wanted to share some early comments.

andrewhead (approved):
Looks good! Left a few comments throughout.

api/src/tests/api.test.ts
```diff
@@ -268,38 +269,53 @@ describe("API", () => {
...
-          sentence: { type: "sentence", id: "3" },
-          parent: { type: "symbol", id: null },
-          children: [
-            { type: "symbol", id: "4" },
-            { type: "symbol", id: "5" },
-          ],
-        },
-      } as Entity);
+      expect(response.result).toEqual({
+        data: [
```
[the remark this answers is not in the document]
mjlangan: I didn't fix it 😬 there's some weird issue with the equality assertion at the end where the correct values are there, but somehow not in the same order, so the check fails. Not really sure why it's doing it.

api/src/types/api.ts
```diff
@@ -68,6 +68,19 @@ export interface EntityGetResponse {
   data?: Entity[];
 }
 
+const entityTypes = {
```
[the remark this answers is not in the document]
mjlangan: Cool. The approach I followed here helps me define the type guard a few lines later, but it's easy enough to rework it using the const array there.

api/src/types/validation.ts
mjlangan: I had trouble trying to figure out how `Joi.array()` expects the query param to look. I knew I wanted to allow either a single string or a comma-separated list, and it was way faster for me to just write a custom function to do it than try to figure out how to express it with Joi's fluent API.

ui/src/ScholarReader.tsx
mjlangan: Somehow I missed that in the dependencies! I'll switch to using that since it does the same job and includes its own types.

PR #236, opened 2021-02-18
Consolidate routes that accept both types of IDs

mjlangan:
I redid the way we extract PaperSelectors out of the route params so that we no longer have to implement a route twice to handle both arxiv and S2 IDs. For routes that only work on arxiv IDs, there's an `arxivOnlySelector` validation option that rejects S2 IDs. I also noticed some of those routes validate a correctly-shaped arxiv ID but then only use the entity ID later in the route, indicating we can simplify the API routes (or do more validation on the paper).

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -192,18 +192,18 @@ export const plugin = {
 
     server.route({
       method: "POST",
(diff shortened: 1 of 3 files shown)


aschokking (approved):
Looks good, nice to have less duplication around this business

PR #237, opened 2021-02-23
Remove entity counts from /papers page to make it useable again

aschokking:
Fixes https://github.com/allenai/scholar/issues/26610

As the DB size has grown, the query here to fetch all papers and their counts have started to timeout our ELB at 60s (which is also way too long to wait anyway).

Since we don't value the counts enough to improve the performance of this query in this PR I propose that we drop the counts altogether. The page now loads instantly with a list of papers that have any citation entities (which is our current bar for displaying a link on the PDP as well).

Before:
image

After:
image

### DIFF
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -61,7 +61,7 @@ export class Connection {
     return await this._knex("logentry").insert(logEntry);
   }
 
-  async getAllPapers(offset: number = 0, size: number = 25, entity_type='citations'): Promise<Paginated<PaperIdInfo>> {
+  async getAllPapers(offset: number = 0, size: number = 25, entity_type: EntityType='citation'): Promise<Paginated<PaperIdInfo>> {
     type Row = PaperIdInfo & {
       total_count: string;
     }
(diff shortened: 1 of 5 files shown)


api/src/db-connection.ts
```diff
@@ -61,28 +61,14 @@ export class Connection {
...
-    type Row = PaperIdWithEntityCounts & {
-        symbol_count: string;
-        citation_count: string;
-        sentence_count: string;
-        term_count: string;
-        equation_count: string;
-        entity_count: string;
-        total_count: string;
-    };
+  async getAllPapers(offset: number = 0, size: number = 25, entity_type='citations'): Promise<Paginated<PaperIdInfo>> {
```
mjlangan, line 64: `entity_type` should be typed. You could use [EntityType](https://github.com/allenai/scholarphi/blob/master/api/src/types/api.ts#L79) here.

api/src/db-connection.ts
```diff
@@ -93,6 +79,7 @@ export class Connection {
    LEFT JOIN entity
           ON entity.paper_id = paper.s2_id
          AND entity.version = version.index
+         AND entity.type = '${entity_type}'
```
mjlangan, line 82: Would be nice to see this using knex's parameter substitution instead of inserting the string directly into the SQL query. There's an example in [getPaperEntityCount](https://github.com/allenai/scholarphi/blob/master/api/src/db-connection.ts#L118-L140)

api/src/db-connection.ts
aschokking: good call, actually highlighted a bug where the type wasn't being applied correctly too

api/src/db-connection.ts
```diff
@@ -61,55 +61,36 @@ export class Connection {
...
-    type Row = PaperIdWithEntityCounts & {
-        symbol_count: string;
-        citation_count: string;
-        sentence_count: string;
-        term_count: string;
-        equation_count: string;
-        entity_count: string;
-        total_count: string;
-    };
+  async getAllPapers(offset: number = 0, size: number = 25, entity_type: EntityType='citation'): Promise<Paginated<PaperIdInfo>> {
```
mjlangan, line 64:
```suggestion
  async getAllPapers(offset: number = 0, size: number = 25, entity_type: EntityType = 'citation'): Promise<Paginated<PaperIdInfo>> {
```
nit (spacing)

mjlangan (approved):

PR #238, opened 2021-02-23
Speed up version check query

aschokking:
Fixes https://github.com/allenai/scholar/issues/26728

Today this call takes ~3-4s to complete, leading to a delay in when a user might see the reader link appear and might have scrolled down the page and missed it. After this PR it takes ~300ms.

When I'd first added this citation count check I had mostly copied and pasted how the max version was being found in the /papers query which turns out to have been horrifically slow. We can instead grab the latest version directly from the `version` table which is tiny compared to the entity table.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -254,12 +254,14 @@ export const plugin = {
         const paperSelector = parsePaperSelector(request.params.arxivSelector);
         const version = await dbConnection.getLatestProcessedArxivVersion(paperSelector);
         const citationCount = version !== null ? await dbConnection.getPaperEntityCount(paperSelector, 'citation'): null;
-        if (version === null || citationCount === null) {
+
+        if(citationCount && citationCount > 0) {
+          return h.response({ version }).code(200);
+        } else {
           // We don't have version info for this ID, or no citations were extracted so we consider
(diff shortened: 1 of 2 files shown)


api/src/api.ts
```diff
@@ -253,8 +253,8 @@ export const plugin = {
       handler: async (request, h) => {
         const paperSelector = parsePaperSelector(request.params.arxivSelector);
         const version = await dbConnection.getLatestProcessedArxivVersion(paperSelector);
-        const citationCount = await dbConnection.getPaperEntityCount(paperSelector, 'citation');
-        if (!version || !citationCount) {
+        const citationCount = version !== null ? await dbConnection.getPaperEntityCount(paperSelector, 'citation'): null;
```
aschokking, line 256: no reason to try and get counts if the version was null

api/src/api.ts
```diff
@@ -253,13 +253,15 @@ export const plugin = {
...
+
+        if(citationCount && citationCount > 0) {
```
aschokking, line 258: this seemed easier to reason about the affirmative case

mjlangan (approved):
🏎️💨

PR #239, opened 2021-02-24
Evaluation: Fix accuracy script to ignore operator symbols

andrewhead:
Ref https://github.com/allenai/scholar/issues/26729

This script is used to compute the accuracy of a set of bounding boxes extracted for entities.

This PR updates the accuracy computation script to ignore operator symbols when computing accuracy. Our gold (hand-labeled) datasets of bounding boxes do not include operators at this time (nor do I think they will anytime soon), though our pipeline has recently been updated to detect them. This means that precision scores take a hit and are difficult to compare to past versions of the precision score.

### DIFF
--- a/data-processing/scripts/compute_accuracy.py
+++ b/data-processing/scripts/compute_accuracy.py
@@ -37,6 +37,12 @@ class AccuracyResults:
     recall: Optional[float]
 
 
+# When processing symbols, ignore those that are not of the types included in the gold set.
+# For example, operators will be ignored because they are not in the gold set. If a symbol
+# does not have an explicit type, it is assumed to be of one of the allowed types.
+GOLD_SYMBOL_TYPES = ["function", "identifier"]
+
+
 def fetch_boxes(
     arxiv_id: ArxivId, schema: str, version: Optional[int], types: List[str]
 ) -> Optional[RegionsByPageAndType]:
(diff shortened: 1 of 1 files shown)


data-processing/scripts/compute_accuracy.py
```diff
@@ -68,20 +69,54 @@ def fetch_boxes(
...
     )
+
     boxes_by_entity_db_id: Dict[str, List[BoundingBox]] = defaultdict(list)
     types_by_entity_db_id: Dict[str, str] = {}
     for row in rows:
+
+        # When processing symbols, ignore those that are not of the types included in the gold set.
+        # For example, operators will be ignored because they are not in the gold set. If a symbol
+        # does not have an explicit type, it is assumed to be of one of the allowed types.
+        GOLD_SYMBOL_TYPES = ["function", "identifier"]
```
aschokking, line 110: Not a big deal, but any reason this filter shouldn't be in the query above as a where clause instead?

aschokking, line 110: Given how important this static list is, consider putting it at the top of the file

aschokking (approved):

data-processing/scripts/compute_accuracy.py
andrewhead: Yup! I was thinking of putting it in the query above. The query was becoming pretty complex (as the desired behavior was to:

1. outer join on the entitydata table on columns where key is 'type', but only for symbols
2. if it is a symbol and it has type but is not one of the gold types, filter it out
3. for all other entity types (e.g., sentences, citations, etc.) do not impose this constraint

I thought it might be more readable without a noticeable hit to performance if I did a simple aggregation of keys and values, returned one row per entity, and did the nuanced conditionals in the Python code. Any thoughts on whether it makes sense to move into the query instead?

andrewhead: Good idea! I will do this.

aschokking: If the query would be much more gross than this logic I'm down with keeping it as is, I just wasn't sure if there was a way for it to be about the same in query form.

PR #240, opened 2021-02-24
S2 api handling

cmwilhelm:
We had limited finesse in handling different classes of
errors coming back from the S2 public API.

This changeset distinguishes between the existing
4XX series errors, including API rate limiting, as well
as capturing previously unhandled error classes, such
as 5XX-series and timeouts.

The command has also been extended to allow use
of a partner API token (as-yet unfurnished), to identify
as a "partner" with less restricted rate limits (something
we're currently seeing).

### DIFF
--- a/data-processing/common/commands/fetch_s2_data.py
+++ b/data-processing/common/commands/fetch_s2_data.py
@@ -13,12 +13,24 @@ from common.types import ArxivId, Author, Reference, S2Metadata, SerializableRef
 """ Time to wait between consecutive requests to S2 API. """
 FETCH_DELAY = 3  # seconds
 
-class S2MetadataException(Exception):
-    pass
+
+class S2ApiException(Exception):
+    """Generic exception related to calling the S2 API"""
+
+
+class S2ApiRateLimitingException(S2ApiException):
+    """
+    Caller has been rate-limited by S2's Public/Partner API.
+    """
+
+class S2MetadataException(S2ApiException):
+    """
+    Represents data availability issues when calling the S2 API.
+    """
 
 class S2PaperNotFoundException(S2MetadataException):
     """
-    The target arxiv paper could not be found on the S2 public api, 
+    The target arxiv paper could not be found on the S2 public api,
     which is a requirement for processing.
     """
 
@@ -26,6 +38,7 @@ class S2ReferencesNotFoundException(S2MetadataException):
     """
     The target arxiv paper did not have any references available via the S2 public api.
     """
+
 
 class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
     @staticmethod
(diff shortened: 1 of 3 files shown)


data-processing/common/commands/fetch_s2_data.py
```diff
@@ -27,7 +44,31 @@ class S2ReferencesNotFoundException(S2MetadataException):
...
+        self._partner_api_token = os.getenv("S2_PARTNER_API_TOKEN", None)
+
+    def _mk_api_request(self, arxivId: ArxivId) -> requests.Response:
+        # XXX(andrewhead): S2 API does not have versions of arXiv papers. I don't think this
+        # will be an issue, but it's something to pay attention to.
+        versionless_id = self._strip_arxiv_version(arxivId)
+
+        if self._partner_api_token:
+            logger.info("Calling S2's partner API")
+            return requests.get(
```
aschokking, line 60: might consider not duplicating the request but rather have the fork change the url base + headers? can avoid the duplicate logging and multiple return statements

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -44,41 +85,49 @@ def load(self) -> Iterator[ArxivId]:
...
-            s2_metadata = S2Metadata(s2_id=data["paperId"], references=references)
-            logging.debug("Fetched S2 metadata for arXiv paper %s", item)
-            yield s2_metadata
+        try:
+            resp = self._mk_api_request(item)
+
+        except Exception:
+            # Exceptions here could reflect temporary service availability
+            # or slowdown issues with S2's API.
+            raise S2ApiException()
```
aschokking, line 95: do we maybe want to capture what the exception was and log it? Right now it's swallowed?

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -44,41 +85,49 @@ def load(self) -> Iterator[ArxivId]:
...
+                logging.debug("Fetched S2 metadata for arXiv paper %s", item)
+                yield s2_metadata
+            elif resp.status_code == 404:
+                # Paper is unavailable in Public API -- potential race condition.
+                raise S2PaperNotFoundException()
+            elif resp.status_code == 429:
+                # Request rate limit has been exceeded for this account or client IP.
+                raise S2ApiRateLimitingException()
+            else:
+                raise S2ApiException()
```
aschokking, line 130: similarly here might be nice to log the status code / response content?

aschokking (approved):
Looks good, a few small suggestions

andrewhead (approved):
Looks great!

data-processing/common/commands/fetch_s2_data.py
cmwilhelm: Yeah sure, seems reasonable.

PR #242, opened 2021-02-26
Pipeline: Bug fix: fix deletion of intermediate files after entity batches that fail

andrewhead:
#218 was designed to circumvent out-of-space errors by deleting the overabundance of intermediate files produced by the pipeline (paper rasters, copies of compiled TeX, etc.) when it localizes entities.

While the PR did succeed in deleting intermediate files, it failed to delete these files when there was a failure in entity detection, like a failed LaTeX compilation.

This PR fixes this issue, ensuring that the intermediate files are always deleted, regardless of whether a failure occurred for one batch of entities.

Addresses issue: https://github.com/allenai/scholar/issues/26607

### DIFF
--- a/data-processing/common/commands/base.py
+++ b/data-processing/common/commands/base.py
@@ -136,7 +136,9 @@ def read_arxiv_ids_from_file(path: Path) -> List[ArxivId]:
         for l in arxiv_ids_file:
             comment_start = l.find("#")
             if comment_start != -1:
-                arxiv_ids.append(l[:comment_start].strip())
+                arxiv_id_or_nothing = l[:comment_start].strip()
+                if arxiv_id_or_nothing != "":
+                    arxiv_ids.append(arxiv_id_or_nothing)
             else:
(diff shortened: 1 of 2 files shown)


data-processing/common/commands/base.py
```diff
@@ -136,7 +136,9 @@ def read_arxiv_ids_from_file(path: Path) -> List[ArxivId]:
         for l in arxiv_ids_file:
             comment_start = l.find("#")
             if comment_start != -1:
-                arxiv_ids.append(l[:comment_start].strip())
+                arxiv_id_or_nothing = l[:comment_start].strip()
+                if arxiv_id_or_nothing != "":
+                    arxiv_ids.append(arxiv_id_or_nothing)
```
andrewhead, line 141: Extra small fix unrelated to this PR, where when the pipeline is getting run based on a list of arXiv IDs in a file, individual arXiv IDs can be commented out.

cmwilhelm (approved):
Seems reasonable.

PR #244, opened 2021-03-01
Retry logic and retryable exit code for Arxiv fetch failures

cmwilhelm:
We're seeing a lot of "peer closed connection" errors when requesting
data from ArXiv. This may or may not be rate limiting; however we should
not treat it as an unrecoverable error.

This changeset does two things:

1) It intercepts expected failure modes in `fetch_from_arxiv`, raising a new error type
that can be caught in a driver script and used to provide a retryable exit code to
`scholarphi-pipeline`

2) Performs an in-process retry loop with more generous backoff delay.

This won't solve our woes but will allow us to at least retry sensibly.

### DIFF
--- a/data-processing/common/commands/fetch_arxiv_sources.py
+++ b/data-processing/common/commands/fetch_arxiv_sources.py
@@ -1,15 +1,21 @@
 import time
 from argparse import ArgumentParser
+import logging
 from typing import Iterator, Optional
 
 from common.commands.base import ArxivBatchCommand
-from common.fetch_arxiv import fetch_from_arxiv, fetch_from_s3
+from common.fetch_arxiv import FetchFromArxivException, fetch_from_arxiv, fetch_from_s3
 from common.types import ArxivId
 
 DEFAULT_S3_ARXIV_SOURCES_BUCKET = "s2-arxiv-sources"
 
-""" Time to wait between consecutive requests to arXiv. """
-FETCH_DELAY = 10  # seconds
+"""Constants for dealing with retries and delays between call attempts"""
+DEFAULT_FETCH_DELAY = 10  # seconds
+BACKOFF_FETCH_DELAY = 60  # seconds
+MAX_FETCH_ATTEMPTS = 3
+
+
+logger = logging.getLogger(__name__)
 
 
 class FetchArxivSources(ArxivBatchCommand[ArxivId, None]):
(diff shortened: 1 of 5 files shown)


data-processing/common/fetch_arxiv.py
```diff
@@ -16,6 +16,13 @@
 USER_AGENT = "Andrew Head, for academic research on dissemination of scientific insight <head.andrewm@gmail.com>"
 
 
+logger = logging.getLogger(__name__)
```
andrewhead, line 19: nit pick: throughout out the code, most logging calls are the form of just `logging.func` (on the imported `logging` object) rather than `logger.func` on an initialized logger object. I am okay with using `logger` here instead. Is there an advantage?

data-processing/scripts/run_pipeline.py
```diff
@@ -86,7 +87,7 @@ def run_commands_for_arxiv_ids(
         # and subprocess calls in the commands, it is simply unlikely that we can predict and
         # write exceptions for every possible exception that could be thrown.
         except Exception as exc:  # pylint: disable=broad-except
-            logging.error("Unexpected exception processing papers: {}".format(arxiv_id_list), exc)
+            logging.exception("Unexpected exception processing papers: {}".format(arxiv_id_list))
```
andrewhead, line 90: I was not aware of this function. Neat!

data-processing/scripts/run_pipeline.py
```diff
@@ -475,7 +476,10 @@ def run_commands_for_arxiv_ids(
                 # digest for a paper cannot be computed once the paper's data is deleted.
                 pipeline_digest.update(digest_for_paper)
             except S2ApiException as e:
-                logging.error("Retryable Failure processing id: {arxiv_id}", e)
+                logging.exception(f"Retryable Failure processing id: {arxiv_id}")
+                exit(RETRYABLE_FAILURE_RETURN_CODE)
+            except FetchFromArxivException as e:
```
andrewhead, line 481: could combine into (not necessary, just possible):

```python
except (S2ApiException, FetchFromArxivException) as e:
```

data-processing/tests/common/commands/test_fetch_arxiv_sources.py
```diff
@@ -0,0 +1,65 @@
...
+@dataclass
+class Args:
+    arxiv_ids: List[str]
+    arxiv_ids_file: Optional[str]
+    source: str
+    s3_bucket: Optional[str]
+
+
+def test_makes_up_to_k_attempts_to_fetch_from_arxiv():
+    with patch("common.commands.fetch_arxiv_sources.time.sleep") as mock_sleep:
```
andrewhead, line 26: When sleep is mocked, does it mean it no longer pauses (i.e., returns instantaneously)?

We have an annotation `@pytest.mark.slow` for slow-running tests that might be appopriate here if there is still some latency associated with the `time.sleep` call.

andrewhead (approved):
lgtm!

data-processing/tests/common/commands/test_fetch_arxiv_sources.py
cmwilhelm: Yeah this overrides the built-in function with a no-op that returns instantly.

data-processing/scripts/run_pipeline.py
cmwilhelm: Didn't know about this -- handy.

data-processing/common/commands/fetch_arxiv_sources.py
```diff
@@ -47,14 +53,34 @@ def load(self) -> Iterator[ArxivId]:
...
             yield arxiv_id
 
-    def process(self, item: ArxivId) -> Iterator[None]:
+    def process(
+        self,
+        item: ArxivId
+    ) -> Iterator[None]:
         if self.args.source == "arxiv":
-            fetch_from_arxiv(item)
+            attempt = 0
```
aschokking, line 61: do we not have a little python util for wrapping things in this kinda retry logic? At BV for instancewe had a python function that took:
- function to call
- optional[exceptions to retry on[]]
- backoff delay

Which made it really easy to throw into places

aschokking (approved):
Looks good 👍

data-processing/common/commands/fetch_arxiv_sources.py
cmwilhelm: We do not have such a utility. It would be good to put in `scholar-data-util`, but that's a private repo we can't depend on here (even though we could copypasta the code).

PR #246, opened 2021-03-04
Non-CRA Webpack config

mjlangan:
I started with the current [skiff app template](https://github.com/allenai/skiff-template/tree/main/ui) and dropped the scholarphi UI code into it, working "backwards" based on how things were configured with create-react-app to get it functional again. Some main changes I'll point out:
- All of the tools involved in building the UI are now version-controlled in package.json. I also followed the new template and stuck everything under `dependencies` since we don't publish this as a library to NPM (which is what `devDependencies` is mainly for). I can split them back up if we want.
- The Skiff template originally used `ts-loader` for typescript and no JS loader. I changed that to use Babel for both TS and JS like we had with the CRA-based stuff.
- Added script `npm run clean` to wipe the build output folder
- Imports from `pdfjs-dist` had to be changed. I'm still not sure why they stopped working as written, but we now have to reach into the library and pull out some of the types. Also I couldn't really find the PDFPageViewport type so I copied it out of the @types/pdfjs-dist package.
- API proxying still works the same as before

You can interact with a version built on this new stack here: https://s2-reader.mjlangan-webpackify.apps.allenai.org/papers


ui/package.json
```diff
@@ -2,109 +2,77 @@
   "name": "scholarphi",
   "version": "0.1.0",
   "private": true,
-  "proxy": "https://s2-reader.apps.allenai.org",
```
mjlangan, line 5: This moves into webpack.config.js

ui/package.json
```diff
@@ -2,109 +2,77 @@
...
-    "@types/react": "^16.8.23",
-    "@types/react-dom": "^16.8.5",
-    "less": "^3.10.3",
-    "less-loader": "^5.0.0",
-    "node-sass": "^4.13.1",
-    "react-scripts": "^3.3.0",
-    "rewire": "^4.0.1",
-    "typescript": "^3.5.3"
-  }
+  "devDependencies": {}
```
mjlangan, line 77: Technically, we don't need to separate the dependencies between `dependencies` and `devDependencies` since there's no intent to distribute this as an NPM package. The Skiff template puts everything in `dependencies` so I kept with that, but I have no strong feelings toward either.

andrewhead: I also do not have a strong opinion either way. Keeping everything in `dependencies` is fine with me.

aschokking: I think the other reason I know of to use dev dependencies is to keep them out of the production image if there's ever a server-side component to things. That said, probably not going to happen anytime soon so I'm okay with collapsing them if you want.

ui/package.json
andrewhead: 👍

aschokking (approved):
Looks good! Thanks for spinning up the test env, with all these moving pieces seems like it would be easy for something to not work live in prod.

ui/src/types/pdfjs-viewer.ts
```diff
@@ -1,4 +1,25 @@
...
+  width: number;
+  height: number;
+  scale: number;
+  transforms: number[];
+
+  clone(options: PDFPageViewportOptions): PDFPageViewport;
+  convertToViewportPoint(x: number, y: number): number[]; // [x, y]
+  convertToViewportRectangle(rect: number[]): number[]; // [x1, y1, x2, y2]
+  convertToPdfPoint(x: number, y: number): number[]; // [x, y]
+}
```
andrewhead, line 22: nit pick: can these two declarations be moved below the block comment below? The block comment below provides exposition describing why the file exists, so it might benefit from continuing to be at the top of the file.

ui/src/types/pdfjs-viewer.ts
```diff
@@ -1,4 +1,25 @@
-import { PDFDocumentProxy, PDFPageProxy, PDFPageViewport } from "pdfjs-dist";
+import { PDFDocumentProxy, PDFPageProxy } from "pdfjs-dist/types/display/api";
+
+interface PDFPageViewportOptions {
```
andrewhead, line 3: It is puzzling that @types/pdfjs-dist changed in how it exports the types! Good to see a fix here :-)

andrewhead (approved):
Agreed with Alex, it is great to see evidence of the build with the live dev server! 👏

ui/src/types/pdfjs-viewer.ts
mjlangan: I'm not really sure what caused this. `pdfjs-dist` also includes typescript definitions itself, and on this stack, the @types packaged seemed like it was being ignored completely. So this import path is where these actual defs from `pdfjs-dist` are located. There's some [confusing communication from the pdfjs devs](https://github.com/mozilla/pdf.js/issues/12384#issuecomment-695071180) about these types only being exported at all for test reasons (yay, JS) but they've kinda become things that all the PDF libraries using PDF.JS expect to work with.

andrewhead: > but they've kinda become things that all the PDF libraries using PDF.JS expect to work with.

Fair enough. I do not have too much concern about these types not necessarily being recommended by the PDF.js team as we update the pdf.js version so rarely (i.e., never to date) that I am not concerned about running into breaking changes related to updated types

andrewhead (commented):
Great, thank you!

ui/src/types/pdfjs-viewer.ts
andrewhead: Great, thank you!

PR #247, opened 2021-03-05
Remove link annotation borders & cache-bust viewer JS and CSS

mjlangan:
No more green/red borders around the PDF links. This should roll out to clients without needing them to manually clear caches, too, as I've added random query param strings for cache-busting.

### DIFF
--- a/ui/public/viewer.css
+++ b/ui/public/viewer.css
@@ -84,6 +84,10 @@
   position: absolute;
 }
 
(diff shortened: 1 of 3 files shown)


aschokking (approved):
Code change looks good 👍  Assume you double checked that citation/symbol ... borders aren't impacted?

andrewhead (approved):
Excellent!

PR #250, opened 2021-03-12
first stab at a new scholarphi db schema

mjlangan:
Draft PR to explore new designs for the ScholarPhi DB schema. 

The schema uses a main `entity` table to assign IDs to all paper entities from the same ID space (one ID to look up any entity, regardless of type). For most purposes outside of the DB layer, this ID is the one we should use when referencing entity data. Values that are common to all entities are kept in this table, and then there are foreign key columns to link out to specific tables for each subtype of entity (e.g. citation, symbol, sentence, etc.).

There are some additional relations to note:
- Symbols can contain other symbols, hence the self-reference in the `symbol` table. One way this is exposed in the reader is users being able to click on a symbol and drill into its constituent parts (if any) and see their definitions as well.
- Symbols appear within Equations and contain a reference back to that equation
- "Terms" and "Symbols" can both have "Definitions"
- "Terms" and "Symbols" both appear within "Sentences"
- All entity types have bounding boxes

Lines and boxes to help paint the picture:
scholarphi2

### DIFF
--- a/api/src/newschema.sql
+++ b/api/src/newschema.sql
@@ -1,0 +1,122 @@
+-- Represents a complete paper within our schema.
+create table paper(
+  id            bigserial primary key,
+  arxiv_id      text,
+  arxiv_version integer,
+  created_at    timestamp without time zone not null, -- necessary?
+  updated_at    timestamp without time zone,  -- necessary?
+  constraint arxiv_id_version_uq unique (arxiv_id, arxiv_version)
+);
+create index paper_arxiv_id_version on paper(arxiv_id);
+
+-- One-to-many relation between papers and S2 IDs for that paper,
+-- since S2 IDs will shift over time.
+create table s2id(
+  paper_id    bigint not null references paper(id),
+  s2_id       text not null,
+  constraint paper_s2id_uq unique (paper_id, s2_id)
+);
+create index s2id_paper_id on s2id(paper_id);
+create index s2id_s2_id on s2id(s2_id);
+
+-- entity data unique to citations
+create table citation(
+  id          bigserial primary key,
+  cites       text not null
+);
+
+-- just assuming this is going to exist?
+create table equation(
+  id              bigserial primary key,
+  text            text, -- would we have this or just the TeX?
+  tex             text
+);
+
+-- entity data unique to symbols
+create table symbol(
+  id                bigserial primary key,
+  normalized_mathml text,
+  tex               text,
+  -- TODO: should these both reference entity(id)?
+  equation_id       bigint not null references equation(id),
+  parent_id         bigint not null references symbol(id)
+);
+create index symbol_equation_id on symbol(equation_id);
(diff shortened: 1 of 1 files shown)


stefanc-ai2:
For reference, this is the current data model. Courtesy of Michael.
IMG_20210311_160444

stefanc-ai2:
@mjlangan for my curiosity, can you point out (in the current code if possible) where the biggest inefficiencies are? I heard something about symbols containing duplicate information.

mjlangan:
> For reference, this is the current data model. Courtesy of Michael.

That's the ERD that informed the new schema design, actually. The current schema looks like this:
scholarphi_schema

Regarding the data duplication, it appears in a few places. The API responses are currently very verbose and contain redundant information when serving up the non-citation entities. For example, if a symbol appears in a sentence, there is an entity for that sentence, and the symbol also contains that sentence as a textual reference (or something of this nature, @andrewhead or @cmwilhelm can probably correct me here). In addition to that, the `entitydata` key-value table allows the same key-value to be written multiple times for a particular entity. An extreme case I found is an entity with 19,520 rows in `entitydata` (this is, 19,520 rows with the same `entity_id`) which actually encoded only about 330 distinct key-value pairs.

kyleclo:
documenting here:  per discussion I had with @mjlangan on 3/30, we settled on these changes to the above proposal:

- consider migrating "text" and "latex" as columns in the entity table, which removes need for separate tables for entities of type sentence or definition.  pretty much every entity can be assumed to have a "text" and "latex" form.
- adding "normalized_mathml" as a field to the equation table.  this is because we need to be able to render these in the application in addition to symbols.
- augmenting the symbol table to have not just a parent lookup, but also a way to lookup all immediate children (not children from deeper layers of parse tree). same thing applies to equation table, though they dont need parents because they are all root nodes.  this is because for the symbol selection feature on the application, when we double-click on a symbol, we want to select the boxes of subsymbols.  
- a entity_cluster table that enables mapping between entity_ids of the exact same string form (e.g. same term or same symbol).  this is for fast behavior on declutter & usages features, where we want to find all other entities to a query entity based on exact string match.   though it's possible that this is no longer needed if all of this is being resolved quickly given the recent API changes.  @mjlangan @cmwilhelm  ?

@andrewhead  can you confirm?

rodneykinney:
I'd like to come to any design review meeting on this. Also happy to give feedback before the review.

andrewhead:
@kyleclo, I am following up on this, hopefully better late than never. Thank you for having this discussion with Michael, and identifying open questions that need to be resolved with regards to the schema design.

Most of these details look good to me. The only place where I was not entirely following is:

> augmenting the symbol table to have not just a parent lookup, but also a way to lookup all immediate children (not children from deeper layers of parse tree).

Why would parent -> child relations be necessary if child -> parent relations are available?

andrewhead:
Got it!

PR #251, opened 2021-03-18
[wip - do not merge] Adding SPP for fuzzy sentence matching (to get bboxes) to pipeline

kyleclo:
Big PR. 

At a high-level, this adds the following steps to the pipeline:

1. Download PDF from arXiv.  You can find this in `data-processing/common/fetch_arxiv.py`
2. Sending the PDF to SPP via HTTP Post & getting back a JSON.  You can find this in `data-processing/entities/sentences_pdf/commands/fetch_spp_jsons.py`
3. Fuzzy matching of sentences with SPP JSON data to get bounding boxes.  You can find this in `data-processing/entities/sentences_pdf/commands/locate_sentences.py`

------

Running the pipeline has changed to:
```
python scripts/run_pipeline.py --arxiv-ids=1601.00978 --entities symbols sentences sentences-pdf
```

Why? Because "sentences" set of commands, which used to be treated like all other entities (e.g. detect in TeX, colorize, detect in images) is no longer our technique for getting sentence bounding boxes.   Now, bounding boxes for sentences depends on SPP JSONs _and_ symbol contexts.  And remember that symbol contexts actually requires detected sentences in TeX.

This means we actually need to do this order:
1. Find sentences in TeX  (this gets their string surface form & start/end char positions)
2. Do all the stuff needed to get full symbols (including finding them in TeX, getting their bboxes, matching to their sentence contexts, etc.)
3. Read in all that stuff & combined w/ SPP JSON to now get the sentence bboxes.

Hence, we've split the sentence commands into 2 steps:  (1) "sentences" entity handles just finding it in TeX.  (2) "sentences-pdf" entity handles getting the bboxes.

------

First, we have to change in `data-processing/entities/sentences/__init__.py`.  We no longer use the template to construct a series of entity extraction commands for sentences.  

----

Next, we add a major new module `data-processing/entities/sentences-pdf/` to the project, where we handle fuzzy matching of sentences found earlier to a SPP token stream data.  These assume SPP JSON already lives in the pipeline somewhere.

----

Finally, we add any pipeline steps that facilitates obtaining SPP JSONs into the pipeline.  This requires fetching PDF from arXiv in `data-processing/common/commands/fetch_arxiv_pdf.py` and shipping that PDF off to SPP in `data-processing/entities/sentences_pdf/commands/fetch_spp_jsons.py`

Note that SPP JSON fetching right now is **not functional**.  We need to set up the SPP endpoint first, hence dont merge this code yet


kyleclo:
TODOs after chatting w/ @cmwilhelm 
[ ] pass the SPP endpoint URL as a CLI arg
[ ] SPP gets PDF from s3 tarball.  check with @ca16

data-processing/entities/sentences_pdf/commands/fetch_spp_jsons.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class SppJsonTask:
+    arxiv_id: ArxivId
+    pdf_path: RelativePath
+    spp_host: str
```
cmwilhelm, line 35: We are now passing in two env vars when the data processor is invoked:

1) `SPP_HOST`
2) `SPP_PORT`

cmwilhelm: From my foggy memory of the `Command` interface, there's a class method you can add to grab env vars for use during command execution.

data-processing/entities/sentences_pdf/commands/fetch_spp_jsons.py
```diff
@@ -0,0 +1,76 @@
...
+        return "Send a PDF to SPP to get back a JSON"
+
+    def get_arxiv_ids_dirkey(self) -> Optional[str]:
+        return 'arxiv-pdfs'
+
+    def load(self) -> Iterator[SppJsonTask]:
+        for arxiv_id in self.arxiv_ids:
+            yield SppJsonTask(
+                arxiv_id=arxiv_id,
+                pdf_path=os.path.join(directories.arxiv_subdir('arxiv-pdfs', arxiv_id=arxiv_id), f'{arxiv_id}.pdf'),
```
cmwilhelm, line 54: So for this, is this what we had chatted about one day, where we let scholarphi compile the Latex->PDF, and we pick up the local PDF here?

data-processing/scripts/commands.py
```diff
@@ -19,12 +20,14 @@
...
 from entities import sentences  # pylint: disable=unused-import
 from entities import symbols  # pylint: disable=unused-import
+from entities import sentences_pdf   # pylint: disable=unused-import
 
 from scripts.pipelines import EntityPipeline, entity_pipelines
 
 # Commands for fetching arXiv sources and preparing for entity processing.
 TEX_PREPARATION_COMMANDS: CommandList = [
     FetchArxivSources,
+    FetchArxivPdf,
```
cmwilhelm, line 30: Per discussion above we probably want to compile the latex and get the pdf from that instead of downloading it from arxiv.

data-processing/entities/sentences_pdf/__init__.py
```diff
@@ -0,0 +1,30 @@
+from common.commands.base import CommandList
```
cmwilhelm, line 1: So just to be clear that I understand -- sentence processing is represented at the meta level by this command list. We are explicitly only using SPP for sentence bbox detection at this time, right?

Separately, is this command list loaded if you ask for `entities="*"`, or only if you ask for `entities="sentences"`?

cmwilhelm (commented):
I think this mostly makes sense to me. I called out a couple things we need to address first, summarized here:

1) Rather than fetching the Arxiv pdfs, we should use the version compiled from the latex in this pipeline.

2) We are passing in the spp host and port as env vars now, so we can leverage them when making the http call.

3) We may want to make this spp based approach opt-in, as there is no publicly available SPP endpoint for people to consume if they are grabbing the OSS distribution of this repo. Alternatively, we could deliberately keep this in a side branch until we have a better story around that. 1) and 2) defo more important.

Let me know how I can help.

PR #253, opened 2021-04-01
Don't repeat the joins once we know the set of entity IDs

mjlangan:
I realized that in our 3 queries to get entities, once we've retrieved all the rows, we continue getting data using the same joins and where clause using `version` and `paperSelector`. This change modifies the latter two queries to select from their respective tables using the set of entity IDs retrieved from the first query. It isn't going to drastically improve performance, but it feels like it should help in some situations.

### DIFF
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -105,14 +105,14 @@ export class Connection {
       FROM paper p
       JOIN entity e on e.paper_id = p.s2_id
       JOIN (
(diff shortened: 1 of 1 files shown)


mjlangan:
This does resolve the massive performance differential between using the arXiv ID in the request URL (fast) and using the S2 ID in the URL (~10x slower)

cmwilhelm (approved):

PR #254, opened 2021-04-01
Deduped payload

cmwilhelm:
Opens a new endpoint to test deduplication of shared data
across instances of a given symbol.

We identify a canonical ID for each symbol, and generate a lookup
table with a single copy of data that are currently duplicated
quadratically at the DB across its instances.

This PR does not change the persistence layer itself, but
defines a desired API response payload shape, and tactically
modified the SQL we are running to build up said payload.

The endpoint can be accessed at:

`v0/papers/<id>/entities-deduped?type=all|symbol`

The response now looks like:

```json
{
    "data": {
        "entities": [ /* as it was minus shared data */ ],
        "sharedSymbolData": {
            someDisambiguatedId: {
                  "defining_formula_equations": [],
                  "defining_formulas": [],
                  "definition_sentences": [],
                  "definition_texs": [],
                  "definitions": [],
                  "snippets": [],
                  "snippet_sentences": [],
                 "sources": [] 
        },
        /* other disambiguated members */
    }
}
```

Symbol entities now include a new member, `disambiguatedId`,
in their attributes to allow for the lookup.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -174,7 +174,7 @@ export const plugin = {
         const entityTypes = request.query.type as EntityType[];
         let res;
         try {
-          res = await dbConnection.getEntitiesForPaper(paperSelector, entityTypes);
+          res = await dbConnection.getEntitiesForPaper(paperSelector, entityTypes, true);
         } catch (e) {
           console.log(e);
         }
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -199,7 +199,7 @@ export const plugin = {
     server.route({
       method: "GET",
       path: "papers/{paperSelector}/entities-deduped",
-      handler: async (request) => {
+      handler: async (request, h) => {
         const paperSelector = parsePaperSelector(request.params.paperSelector);
         // Runtime type-checked during validation.
         const entityTypes = request.query.type as EntityType[];
@@ -208,6 +208,7 @@ export const plugin = {
           res = await dbConnection.getDedupedEntitiesForPaper(paperSelector, entityTypes);
         } catch (e) {
           console.log(e);
+          return h.response().code(500);
         }
         return { data: res };
       },
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -190,6 +190,38 @@ export const plugin = {
       },
     });
 
+    /**
+     * The default implementation above is enormously wasteful in terms of space,
+     * due to a quadratic blowout in identical data.
+     * This route exists as a temporary fixture during development to dedupe data,
+     * accomplished via surgical DB queries rather than exhaustive ones.
+     */
+    server.route({
+      method: "GET",
+      path: "papers/{paperSelector}/entities-deduped",
+      handler: async (request) => {
+        const paperSelector = parsePaperSelector(request.params.paperSelector);
+        // Runtime type-checked during validation.
+        const entityTypes = request.query.type as EntityType[];
+        let res;
+        try {
+          res = await dbConnection.getDedupedEntitiesForPaper(paperSelector, entityTypes);
+        } catch (e) {
+          console.log(e);
+        }
+        return { data: res };
+      },
+      options: {
+        validate: {
+          params: validation.paperSelector,
+          query: Joi.object({
+            type:  validation.apiEntityTypes
+          })
+        },
+      },
+    });
+
+
     server.route({
       method: "POST",
       path: "papers/{arxivSelector}/entities",
@@ -261,7 +293,7 @@ export const plugin = {
           // We don't have version info for this ID, or no citations were extracted so we consider
           // it unsuccessfully processed.
           return h.response().code(404);
-        }    
+        }
       },
       options: {
         validate: {
--- a/api/src/db-connection.ts
+++ b/api/src/db-connection.ts
@@ -431,8 +431,6 @@ export class Connection {
         }
       });
 
-    const entityIds = entityRows.map(e => e.id);
-
     let boundingBoxRows: BoundingBoxRow[];
     try {
       boundingBoxRows = await this._knex("boundingbox")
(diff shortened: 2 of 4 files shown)


api/src/api.ts
```diff
@@ -190,6 +190,38 @@ export const plugin = {
...
+      path: "papers/{paperSelector}/entities-deduped",
+      handler: async (request) => {
+        const paperSelector = parsePaperSelector(request.params.paperSelector);
+        // Runtime type-checked during validation.
+        const entityTypes = request.query.type as EntityType[];
+        let res;
+        try {
+          res = await dbConnection.getDedupedEntitiesForPaper(paperSelector, entityTypes);
+        } catch (e) {
+          console.log(e);
```
mjlangan, line 210: The above implementation doesn't do this, but it's something I've added in my own branch so you might also want to do it here. If you add an `h` argument to this handler function, you can `return h.response().code(500);` in here to make the API server respond with a 500 error instead of an empty set of data.

api/src/db-connection.ts
```diff
@@ -257,6 +261,12 @@ export class Connection {
         ...relationships,
       },
     };
+
+    if (isSymbol(entity)) {
+      (entity as Symbol).attributes.disambiguated_id = entity.attributes.mathml
```
mjlangan, line 266:
```suggestion
      entity.attributes.disambiguated_id = entity.attributes.mathml
```
The type guard in the conditional is doing this cast for you.

cmwilhelm: Didn't realize that. The miracle of TS.

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+  async getDedupedEntitiesForPaper(paperSelector: PaperSelector, entityTypes: EntityType[], version?: number) {
+    if (version === undefined) {
+      try {
+        let latestVersion = await this.getLatestPaperDataVersion(paperSelector);
+        if (latestVersion === null) {
+          return [];
+        }
+        version = latestVersion;
+      } catch (e) {
+        console.log("Error fetching latest data version number:", e);
```
mjlangan, line 419: I feel like we should probably be throwing `e` up the stack here so it can be caught in the API and we actually 500 if this happens.

api/src/api.ts
cmwilhelm: 👍

api/src/db-connection.ts
cmwilhelm: Hm this is a straight copy pasta from the original method. Would it help for me to delineate what's new in this version inline?

api/src/db-connection.ts
mjlangan: Isn't it wondeful?

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+        "entity_id",
+        "source",
+        "key",
+        "value",
+        "item_type",
+        "of_list",
+        "relation_type"
+      )
+      .whereIn("entity_id", entityIds)
+      .whereNotIn("entitydata.key", sharedSymbolFields)
```
cmwilhelm, line 484: Single difference here from OG is this exclusion on the `sharedSymbolFields`

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+            "Invalid entity will not be returned. Error:",
+            validationResult.error
+          );
+          return false;
+        }
+        return true;
+      })
+      .map((validationResult) => validationResult.value as Entity);
+
+    /**
```
cmwilhelm, line 540: Here onward is unique to this deduping method.

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+          "boundingbox.source AS source",
+          "page",
+          "left",
+          "top",
+          "width",
+          "height"
+        )
+        .join("entity", { "boundingbox.entity_id": "entity.id" })
+        .join("paper", { "paper.s2_id": "entity.paper_id" })
+        .where({ ...paperSelector, version })
```
mjlangan, line 451: You can also apply the same look-ma-no-joins trick here

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+    } catch (e) {
+      console.log(e);
+      throw "Error";
```
mjlangan, line 454: Beautiful.

api/src/db-connection.ts
```diff
@@ -389,6 +399,216 @@ export class Connection {
...
+      (dict, row) => {
+        const disambiguatedId = exemplarEntityIdsToDisambiguatedSymbolIds[row.entity_id];
+
+        if (!(disambiguatedId in dict)) {
+          dict[disambiguatedId] = sharedSymbolFields.reduce(
+            (dict, key) => {
+              dict[key] = [];
+              return dict;
+            },
+            {} as {[shitlistedKey: string]: string[]}
```
mjlangan, line 594: ya missed one ;)

api/src/types/api.ts
```diff
@@ -221,6 +222,33 @@ export type GenericRelationships = {
   [key: string]: Relationship | Relationship[];
 };
 
+
```
mjlangan, line 225: nit: extra whitespace line

api/src/types/api.ts
```diff
@@ -152,6 +152,7 @@ export const BASE_ENTITY_ATTRIBUTE_KEYS = [
   "tags",
 ];
 
+
```
mjlangan, line 155: nit: extra whitespace line

api/src/db-connection.ts
```diff
@@ -257,9 +261,15 @@ export class Connection {
         ...relationships,
       },
     };
+
+    if (isSymbol(entity)) {
+      entity.attributes.disambiguated_id = entity.attributes.mathml
```
mjlangan, line 266: nit: `;` (sorry this should really have prettier set up to handle this already...)

api/src/db-connection.ts
```diff
@@ -257,9 +261,15 @@ export class Connection {
...
+    }
+
+    return entity
```
mjlangan, line 269:
```suggestion
    return entity;
```

mjlangan: This is totally optional because I just scrolled ahead and realized that yes auto-formatting would be a very useful thing for this repo 😔

api/src/db-connection.ts
```diff
@@ -389,6 +409,92 @@ export class Connection {
...
+
+    // To provide a deduped copy of shared data between symbol instances, we
+    // identify each symbol by its "disambiguated" id (currently the `mathml` attribute).
+    // An arbitrary symbol entity from within each `mathml` is chosen as an exemplar from
+    // which to look up supporting data that is identical across instances.
+    const disambiguatedSymbolIdsToExemplarEntityIds = entities
+      .filter((row) => isSymbol(row))
+      .reduce(
+        (dict, symbol) => {
+          const disambiguatedId = (symbol as Symbol).attributes.mathml;
```
mjlangan, line 438:
```suggestion
        (dict, symbol: Symbol) => {
          const disambiguatedId = symbol.attributes.mathml;
```
This might also work.

mjlangan (approved):
LGTM

api/src/db-connection.ts
cmwilhelm: 😢 

I've been out of the JS game for a while, can you tell?

api/src/db-connection.ts
cmwilhelm: Sadly it does not. The explicit cast in 438 appears necessary, even though symbol is recognized as a `Symbol` in the signature.

PR #255, opened 2021-04-01
Optionally omit certain fields from entity responses

mjlangan:
This branch changes 3 things about the current entity response:
- `source` fields are removed from entities and bounding boxes. These are sent by the UI when modifying entities, but the UI doesn't really do anything with this information besides applying a CSS style on one component (which is applied for 99% of entities).
- The entity's `version` field is similarly removed because the UI does nothing with it
- Relationships omit the `type` field because they tend to only relate to one type of thing (and it's often in the name of the relationship).

It's all behind a `slim` query param that's `false` by default so it doesn't break the UI.

In a smaller case, it's good for about 27kB savings:
`GET http://localhost:3000/api/v0/papers/arxiv:1601.00978v1/entities?type=all&slim=false` - 188.37kB
`GET http://localhost:3000/api/v0/papers/arxiv:1601.00978v1/entities?type=all&slim=true` - 161.93kB

In a much larger case, it's good for about 14MB savings:
`GET http://localhost:3000/api/v0/papers/arxiv:2101.12018v1/entities?type=all&slim=false` - 58.14MB
`GET http://localhost:3000/api/v0/papers/arxiv:2101.12018v1/entities?type=all&slim=true` - 44.18MB

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -4,12 +4,13 @@ import * as Joi from "@hapi/joi";
 import { Connection, PaperSelector } from "./db-connection";
 import * as s2Api from "./s2-api";
 import {
+  Entity,
   EntityCreatePayload,
   EntityUpdatePayload,
+  EntityType,
   Paper,
   PaperWithIdInfo,
-  Paginated,
-  EntityType
+  Paginated
 } from "./types/api";
 import * as validation from "./types/validation";
 import * as conf from "./conf";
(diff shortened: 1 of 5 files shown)


api/src/api.ts
```diff
@@ -168,23 +169,36 @@ export const plugin = {
...
         const paperSelector = parsePaperSelector(request.params.paperSelector);
         // Runtime type-checked during validation.
         const entityTypes = request.query.type as EntityType[];
-        let res;
+        // pretty sure Joi is actually making this a boolean, despite Hapi saying it's a string
+        const slim = !!request.query.slim;
+        let res: Entity[] = [];
         try {
-          res = await dbConnection.getEntitiesForPaper(paperSelector, entityTypes);
+          res = await dbConnection.getEntitiesForPaper(paperSelector, entityTypes, slim);
```
cmwilhelm, line 180: When you reconcile against my branch you'll want to take a peek at this -- I have a new boolean arg to this method as well.

api/src/db-connection.ts
```diff
@@ -319,17 +320,11 @@ export class Connection {
         [entity_id: string]: BoundingBoxRow[];
       }
     );
-
+    const entityDataColumns = slim
```
cmwilhelm, line 323: This part will conflict with my PR as I had to revive the joins :( :( :( :(

cmwilhelm (approved):
LGTM, nice slimmening.

I called out the spots that I know will require some manual massage on reconciliation with my changes.

cmwilhelm (approved):
LGTM 🚀

PR #256, opened 2021-04-03
 HEDDEx (Definition extractor) joint model integration (v4.1)

nuwandavek:
## Changes
- Use the joint model instead of 3 individual models for each task (termdef, abbrexp, symnick)
- Deleted legacy code in load_data.py

### DIFF
--- a/data-processing/entities/definitions/commands/detect_definitions.py
+++ b/data-processing/entities/definitions/commands/detect_definitions.py
@@ -513,7 +513,7 @@ class DetectDefinitions(
 ):
     """
     Extract terms and definitions of those terms from sentences using a pre-trained definition
-    extraction model. Performs the following stpes:
+    extraction model. Performs the following steps:
     1. Load cleaned sentences from the 'detected-sentences' data directory
     2. Extract features from each sentence (aka featurization)
     3. Load the pre-trained NLP model, load the features, and predict intent and term:definition
@@ -601,8 +601,8 @@ class DetectDefinitions(
             return
 
         # Load the pre-trained definition detection model.
-        prediction_types = ["AI2020", "DocDef2", "W00"]
-        model = DefinitionDetectionModel(prediction_types)
+        prediction_type = "DocDef2+AI2020+W00"
+        model = DefinitionDetectionModel(prediction_type)
 
         definition_index = 0
         features = []
@@ -641,14 +641,8 @@ class DetectDefinitions(
                     # Detect terms and definitions in each sentence with a pre-trained definition
                     # extraction model, from the featurized text.
 
-                    (_, termdef_slots, termdef_slots_confidence,) = model.predict_batch(
-                        cast(List[Dict[Any, Any]], features), "W00"
-                    )
-                    (_, abbrexp_slots, abbrexp_slots_confidence,) = model.predict_batch(
-                        cast(List[Dict[Any, Any]], features), "AI2020"
-                    )
-                    (_, symnick_slots, symnick_slots_confidence,) = model.predict_batch(
-                        cast(List[Dict[Any, Any]], features), "DocDef2"
+                    (_, slots, slots_confidence) = model.predict_batch(
+                        cast(List[Dict[Any, Any]], features)
                     )
 
                     # Package extracted terms and definitions into a representation that's
@@ -665,12 +659,12 @@ class DetectDefinitions(
                     ) in zip(
                         sentences,
                         features,
-                        termdef_slots["W00"],
-                        termdef_slots_confidence["W00"],
-                        abbrexp_slots["AI2020"],
-                        abbrexp_slots_confidence["AI2020"],
-                        symnick_slots["DocDef2"],
-                        symnick_slots_confidence["DocDef2"],
+                        slots["W00"],
+                        slots_confidence["W00"],
+                        slots["AI2020"],
+                        slots_confidence["AI2020"],
+                        slots["DocDef2"],
+                        slots_confidence["DocDef2"],
                     ):
                         # Extract TeX for each symbol from a parallel representation of the
                         # sentence, so that the TeX for symbols can be saved.
--- a/data-processing/entities/definitions/model/configuration.py
+++ b/data-processing/entities/definitions/model/configuration.py
@@ -114,6 +114,24 @@ class DataTrainingArguments:
         default=50, metadata={"help": "TBW"},
     )
     dataconfig_file: Optional[str] = field(
-        default="dataconfig.json", metadata={"help": "dataconfig file (it has intent_label, slot_label dicts, and pos_label list)"},
+        default="dataconfig.json", metadata={"help": 
+        '''
+            The `dataconfig.json` file, that is bundled along with the joint model. 
+            It has intent_label dict, slot_label dict, and pos_label list) for all 
+            the datasets used in to train the model (and thus used during inference).
+            Schema of the file : 
+                {
+                    "intent_label" : {
+                        <dataset_1> : <List of intent labels>,
+                        <dataset_2> : <List of intent labels>,
+                    },
+                    "slot_label" : {
+                        <dataset_1> : <List of slot labels>,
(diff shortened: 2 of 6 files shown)


data-processing/entities/definitions/commands/detect_definitions.py
```diff
@@ -601,8 +601,8 @@ def process(
             return
 
         # Load the pre-trained definition detection model.
-        prediction_types = ["AI2020", "DocDef2", "W00"]
-        model = DefinitionDetectionModel(prediction_types)
+        prediction_type = "DocDef2+AI2020+W00"
+        model = DefinitionDetectionModel(prediction_type)
```
andrewhead, line 605: Does it make sense to still provide prediction_type as a pararmeter to DefinitionDetectionModel if only one prediction type will ever be used?

data-processing/tests/test_extract_definitions.py
```diff
@@ -23,10 +23,11 @@ def model():
 @pytest.mark.slow
 def test_model_extracts_simple_definitions(model: DefinitionDetectionModel):
     prediction_type = "W00"
-    features = model.featurize("Neural networks are machine learning models.")
-    intents, slots, _ = model.predict_batch([features], prediction_type)
+    features = model.featurize("Neural networks are a type of machine learning models.")
```
andrewhead, line 26: What is the reason for the change to this test?

data-processing/entities/definitions/model/configuration.py
```diff
@@ -113,3 +113,7 @@ class DataTrainingArguments:
     max_seq_len: Optional[int] = field(
         default=50, metadata={"help": "TBW"},
     )
+    dataconfig_file: Optional[str] = field(
+        default="dataconfig.json", metadata={"help": "dataconfig file (it has intent_label, slot_label dicts, and pos_label list)"},
```
andrewhead, line 117: Should there be a dataconfig.json file as part of this commit?

Also, what is the expected format of the dataconfig.json file? If it is a standard file format (e.g., documented as part of the Pytorch library or something), then can a link be provided here?

data-processing/entities/definitions/commands/detect_definitions.py
nuwandavek: This is unfortunately an artefact of how the model is trained. The prediction type text is embedded deep in the model (as the keys to model heads). So, I'm afraid we still need to specify this somewhere.

data-processing/tests/test_extract_definitions.py
nuwandavek: The test was failing here. I inspected the logits and could not understand why. I think this was a borderline case, in the earlier model. The text based part of the model is still not as good as the other two heads (but it is actually better than the previous version of the model in aggregate, just not this case surprisingly). So, I rephrased it a little, and it worked. 

But I think we need to somewhere record model test cases (one could argue that's what the dev set is, but I'm not sure where).

data-processing/entities/definitions/model/configuration.py
nuwandavek: No, It's bundled along with the model. Earlier, we had `intent_label.txt` , `pos_label.txt` and `slot_label.txt` files bundled with each model. But since this model was trained on 3 datasets, we would have had to have 9 files with names corrected. So, I made a json file with all the content for better maintenance.  Where do you think is a good place to describe it's schema?

data-processing/entities/definitions/nlp.py
```diff
@@ -62,200 +61,197 @@ def __init__(self, prediction_types: List[str]) -> None:
                 "file": "symnick.zip",
                 "type": "sym-nick",
             },
+            "DocDef2+AI2020+W00": {
+                "baseURL": "https://scholarphi.s3-us-west-1.amazonaws.com/",
+                "file": "joint_abbrexp_termdef_symnick.zip",
+                "type": "joint",
+            },
         }
```
nuwandavek, line 69: @andrewhead I've added the new model location path here

data-processing/entities/definitions/nlp.py
```diff
@@ -62,200 +61,197 @@ def __init__(self, prediction_types: List[str]) -> None:
...
+        self.tokenizer = tokenizer
+        self.model = model
+        self.trainer = Trainer(
+            [
+                training_args,
+                self.model_args,
+                self.data_args,
+            ],
+            self.model,
+        )
```
nuwandavek, line 255: @andrewhead  Every instance where we had a dict of 3 models, 3 tokenizers etc have been replaced by one single model, tokenizer etc.

data-processing/entities/definitions/nlp.py
```diff
@@ -309,18 +305,16 @@ def featurize(self, text: str, limit: bool = False) -> DefaultDict[Any, Any]:
...
         test_dataset, raw = load_and_cache_example_batch_raw(
-            self.data_args[prediction_type], self.tokenizer[prediction_type], data,
+            self.data_args, self.tokenizer, data,
         )
 
         # Perform inference.
-        intent_pred, slot_preds, slot_pred_confs = self.trainer[
-            prediction_type
-        ].evaluate_from_input(test_dataset, raw)
+        intent_pred, slot_preds, slot_pred_confs = self.trainer.evaluate_from_input(test_dataset, raw)
```
nuwandavek, line 318: @andrewhead Also while predicting, we call the args, tokenizer and model directly as opposed to accessing it via a dictionary element.

data-processing/entities/definitions/model/configuration.py
andrewhead: Are there any circumstnades under which we would use something besides "dataconfig.json" as the dataconfig file? If not, then perhaps we should just code this as a constant.

andrewhead: > So, I made a json file with all the content for better maintenance. Where do you think is a good place to describe it's schema?

I would recommend two things:

1. Providing a link to one of the dataconfig.json file on S3 (someone can open this file to see the format themselves)
2. Add documentation of the schema right here as comments (in particular, what are the fields and allowable values)

data-processing/entities/definitions/nlp.py
andrewhead: sounds good!

data-processing/entities/definitions/nlp.py
andrewhead: great!

data-processing/entities/definitions/commands/detect_definitions.py
andrewhead: Great, thank you for the clarification. Maybe add this as a comment for DefinitionDetectionModel if it is not yet there?

data-processing/tests/test_extract_definitions.py
andrewhead: Then let us include the original test case as a skipped test case to document that a future definition detector should probably extract it, as is already done for this test case in the same file https://github.com/allenai/scholarphi/blob/8632c464f58c0b787ba2dd363b5be449f5b1fa98/data-processing/tests/test_extract_definitions.py#L470-L479

andrewhead (approved):
@nuwandavek this looks good. Can you take a look and resolve unresolved comments? Then I think this is ready to merge.

data-processing/entities/definitions/commands/detect_definitions.py
nuwandavek: Done!

data-processing/tests/test_extract_definitions.py
nuwandavek: Done!

data-processing/entities/definitions/model/configuration.py
nuwandavek: It is actually in the zip of the model 😓 . But I shall add a note and the schema

nuwandavek: Done!

PR #261, opened 2021-04-07
Use the deduped entities API for loading the entities

mjlangan:
This change adds an API handler for the new `entities-deduped` endpoint and a glorious un-dedupe function to make its response compatible with the existing code. Not an efficient use of memory within the browser, but it at least allows us to get the perf benefit of the new API without having to change all the UI component code.

Because `entities-deduped` is a much faster API for returning the full set of entity data, I also removed the `&showAll=true` URL param from the reader and made it always request the full set.

A couple related notes:
- With the slimmer API responses, the entities don't include their version numbers (since we only query one version at a time and these are always the same value) or sources (like 97% of these are `tex-pipeline` with a small remainder of `human-annotator`)
- `tags` also were generally empty and so don't get returned, but this change can be undone because I recently realized that tags were added as means of filtering out entities with weird bounding boxes.

I've done some ad-hoc testing to verify that it works for the paper `https://arxiv.org/pdf/1601.00978v1.pdf` but we should test some other papers with symbol content to increase confidence.

### DIFF
--- a/ui/src/api/api.ts
+++ b/ui/src/api/api.ts
@@ -218,6 +218,19 @@ function undedupeResponse(response: DedupedEntityResponse): EntityGetResponse {
   };
 }
 
+
+/**
+ * This API returns a compacted representation of the paper's entities and their relationships.
+ * Certain fields of Symbol entity data are pulled into a separate `sharedSymbolData` map,
+ * which is organized by the Symbols' `attributes.disambiguated_id` value.
+ *
+ * NOTE: Currently, this function passes the response through a transform function to make
+ * it compatible with the existing UI code.
+ * 
+ * @param arxivId arXiv ID of the viewed paper
+ * @param getAllEntities `true` retrieves entities of all types, `false` retrieves only citations
+ * @returns
+ */
 export async function getDedupedEntities(arxivId: string, getAllEntities?: boolean) {
   const params = getAllEntities ? {
     type: ENTITY_API_ALL
--- a/ui/src/api/api.ts
+++ b/ui/src/api/api.ts
@@ -11,7 +11,14 @@ import {
   Paper,
   Paginated,
   PaperIdWithEntityCounts,
+  EntityGetResponse,
+  Citation,
+  Equation,
+  Sentence,
+  Symbol,
+  Term,
 } from "./types";
+import { Entity as DedupedEntity, DedupedEntityResponse, isCitation, isEquation, isSentence, isSymbol, isTerm, toLegacyRelationship } from "./deduped";
 
 const token = cookie.parse(document.cookie)["readerAdminToken"];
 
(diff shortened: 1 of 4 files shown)


ui/src/api/api.ts
```diff
@@ -72,6 +88,175 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
...
+  const params = getAllEntities ? {
+    type: ENTITY_API_ALL
+  } : {};
+  const data = await doGet(
+    axios.get<EntityGetResponse>(
+      `/api/v0/papers/arxiv:${arxivId}/entities-deduped`,
+      {
+        params,
+        //@ts-ignore -- TODO: this pattern works in other projects, is there a version issue somewhere?
+        transformResponse: [].concat(axios.defaults.transformResponse).concat(undedupeResponse)
```
mjlangan, line 253: The TS type checker wasn't happy with this line, complaining that the type for `axios.defaults.transformResponse` didn't match, which seems odd considering that's the default set of response transforms. I used axios on the Beaker project here and used this same pattern without type errors, so I'm wondering if maybe there's some combination of TS/axios versions that is causing the error.

ui/src/api/api.ts
```diff
@@ -72,6 +88,175 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
...
+          tex: deduped.attributes.tex,
+          tex_start: deduped.attributes.tex_start,
+          tex_end: deduped.attributes.tex_end,
+          text: deduped.attributes.text,
+        },
+        relationships: {},
+      };
+      return sentence;
+    } else if (isSymbol(deduped)) {
+      const mathml = deduped.attributes.disambiguated_id;
```
cmwilhelm, line 143: We should probably call this `const disambiguatedId` because I think the connection to `mathml` is likely to change if/when Andrew does real symbol disambiguation in the future temporary.

ui/src/api/deduped.ts
```diff
@@ -0,0 +1,180 @@
...
+  tex: Nullable<string>;
+  type: "identifier" | "function" | "operator"; // TODO: This is null too with the deduped endpoint
+  mathml: Nullable<string>;
+  mathml_near_matches: string[];
+  is_definition: Nullable<boolean>;
+  diagram_label: Nullable<string>;
+  /**
+   * This is the ID to use when getting SharedSymbolData for this entity.
+   */
+  disambiguated_id: string;
```
cmwilhelm, line 55: This is technically nullable right now.

cmwilhelm (approved):
Before merging this, worth pulling up an example paper or two in both versions and verifying all the symbol related stuff in particular looks the same.

mjlangan:
I've manually compared some papers between my branch and what's deployed and things have matched up 100% thus far. This paper in particular seems to exercise a lot of the symbol interactivity: https://scholarphi.semanticscholar.org/?file=https://arxiv.org/pdf/2005.06967v2.pdf&showAll=true

I'm therefore reasonably confident this is working and will merge.

ui/src/api/api.ts
pauls-ai2: I think this is complaining because `[].concat()` takes an array as an argument. But you may as well do something like this:
```
transformResponse: [
  ...axios.defaults.transformResponse,
  undedupeResponse,
]
```

ui/src/api/api.ts
```diff
@@ -72,6 +88,176 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
...
+      };
+      return sentence;
+    } else if (isSymbol(deduped)) {
+       // empty string is a safe default, there shouldn't be sharedSymbolData for empty string.
+      const disambiguatedId = deduped.attributes.disambiguated_id || '';
+      const symbol: Symbol = {
+        id: deduped.id,
+        type: 'symbol',
+        attributes: {
+          bounding_boxes: deduped.attributes.bounding_boxes,
```
andrewhead, line 149: One shorthand that might make this a bit more succinct could be:

```javascript
  attributes: {
    ...deduped.attributes,
    source: 'tex-pipeline',
    tags: []
    definitions: response.sharedSymbolData...
  }
```

ui/src/api/deduped.ts
```diff
@@ -0,0 +1,180 @@
...
+  nickname_sentences: Relationship[];
+}
+
+export interface Equation extends EntityBase {
+  type: 'equation';
+  attributes: EquationAttributes;
+}
+
+interface EquationAttributes extends AttributesBase {
+  tex: Nullable<string>;
```
andrewhead, line 111: Question for my edification: what is the difference between `Nullable<string>` and `string | null`?

ui/src/api/deduped.ts
```diff
@@ -0,0 +1,180 @@
+import { BoundingBox, Relationship as LegacyRelationship } from './types';
+
+export type Nullable<T> = T | null;
+
+/**
+ * Types defined in here are for the entities-deduped endpoint,
+ * and are intentionally less generic.
+ * TODO: Replace existing entity type usages with these
```
andrewhead, line 8: Thank you for this note!

ui/src/api/deduped.ts
```diff
@@ -0,0 +1,180 @@
...
+interface SentenceAttributes extends AttributesBase {
+  text: Nullable<string>;
+  tex: Nullable<string>;
+  tex_start: Nullable<number>;
+  tex_end: Nullable<number>;
+}
+
+interface SharedSymbolData {
+  defining_formula_equations: string[];  // entity id
+  defining_formulas: string[];  // tex-bearing formula text
```
andrewhead, line 128: Checking my understanding: Are these two arrays aligned such that a pair of entries in the two lists are the ID for an equation and the corresponding TeX for the equation with that ID?

ui/src/api/api.ts
```diff
@@ -72,6 +88,176 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
...
+ * it compatible with the existing UI code.
+ *
+ * @param arxivId arXiv ID of the viewed paper
+ * @param getAllEntities `true` retrieves entities of all types, `false` retrieves only citations
+ * @returns
+ */
+export async function getDedupedEntities(arxivId: string, getAllEntities?: boolean) {
+  const params = getAllEntities ? {
+    type: ENTITY_API_ALL
+  } : {};
```
andrewhead, line 247: Checking my understanding: does setting `type` to `{}` results in the API returning the default set of entities?

ui/src/api/api.ts
```diff
@@ -72,6 +88,176 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
...
+  const data = await doGet(
+    axios.get<EntityGetResponse>(
```
andrewhead, line 249: Should this return type be `DedupedEntityResponse`?

andrewhead (commented):
Looks good! Left some questions for my edification, and one small suggestion.

PR #262, opened 2021-04-07
Fix two UI bugs and add tests

mjlangan:
Two bugs I found recently:
1. If one is adding a query param after the PDF URL, and one fat fingers the `&`, pdf.js will ignore the garbage after the `.pdf` extension, but our code matches everything from the arXiv ID to the end of the string (e.g. `2005.06967v2.pdf?showAll=true`). This is a bug because the pdf.js viewer ignores that and properly loads the PDF, but ScholarPhi breaks with no visual indication to the user. Now the regex will correctly pull the arXiv ID out of such a URL.
2. Some math papers on arXiv have IDs that look like `math/0008020v2`. Our API puts the arXiv ID in a URL parameter, so these math IDs actually change the path the API server thinks we're requesting and returns 404 errors for those papers. I noticed we weren't encoding the URI params, so characters like `/` and `?` could cause all kinds of problems, and added the encoding calls. Now those math papers will at least hit the right API endpoint, but it currently returns a 400 because it doesn't decode the encoded characters yet.

I also added Jest and some unit tests around that `extractArxivId` function. Hooray having tests 🎉

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,16 @@ jobs:
           node-version: 12.21.0
       - run: npm install
       - run: npm test
+      # Build the image, so we're sure it still builds.
+      - run: docker build .
+      # We run the build a second time instead of trying to wire the docker SHA from
+      # one step to another. The layers should be cached, so it should be really fast.
+      # We run the build in a separate step above so that there's nice output if something
+      # doesn't work.
+      - run: docker run --name ui --rm -d -p 4000:4000 $(docker build -q .)
+      # Do a simple check that the server returns 200
+      - run: curl --fail http://localhost:4000
+      - run: docker stop ui
   api:
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,15 +10,11 @@ jobs:
         working-directory: ui
     steps:
       - uses: actions/checkout@v2
-      - run: docker build .
-      # We run the build a second time instead of trying to wire the docker SHA from 
-      # one step to another. The layers should be cached, so it should be really fast.
-      # We run the build in a separate step above so that there's nice output if something
-      # doesn't work.
-      - run: docker run --name ui --rm -d -p 4000:4000 $(docker build -q .)
-      # We don't have UI tests yet, so we just make sure the `/` route returns 200.
-      - run: curl --fail http://localhost:4000
-      - run: docker stop ui
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.20.0
+      - run: npm install
+      - run: npm test
   api:
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.20.0
+          node-version: 12.21.0
       - run: npm install
       - run: npm test
   api:
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.20.0
+          node-version: 12.21.0
       - run: npm install
       - run: ./bin/start_local_db.sh
       - run: npm test
--- a/ui/jest.config.js
+++ b/ui/jest.config.js
@@ -1,0 +1,8 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  testPathIgnorePatterns: [
+    "<rootDir>/build/",
+    "<rootDir>/node_modules/"
+  ]
+};
--- a/ui/jest.config.js
+++ b/ui/jest.config.js
@@ -1,8 +1,0 @@
-module.exports = {
-  preset: 'ts-jest',
-  testEnvironment: 'node',
-  testPathIgnorePatterns: [
-    "<rootDir>/build/",
-    "<rootDir>/node_modules/"
-  ]
-};
--- a/ui/package.json
+++ b/ui/package.json
@@ -73,6 +73,7 @@
     "style-loader": "^2.0.0",
     "ts-jest": "^26.5.4",
     "ts-loader": "^8.0.17",
+    "ts-node": "^9.1.1",
     "typescript": "^3.5.3",
     "webpack": "^4.29.5",
     "webpack-cli": "^4.5.0",
--- a/ui/package.json
+++ b/ui/package.json
@@ -7,7 +7,8 @@
     "clean": "rimraf ./build/",
     "lint": "eslint '**/*.{js,ts,tsx,json}' && echo '💫  Lint complete.'",
     "lint:fix": "eslint '**/*.{js,ts,tsx,json}' --fix && npx sort-package-json  && echo '🛠  Lint --fix complete'",
-    "start": "webpack serve --hot --mode development"
+    "start": "webpack serve --hot --mode development",
+    "test": "jest"
   },
   "resolutions": {
     "@typescript-eslint/eslint-plugin": "^4.1.1",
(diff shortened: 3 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -10,15 +10,11 @@ jobs:
...
-      # doesn't work.
-      - run: docker run --name ui --rm -d -p 4000:4000 $(docker build -q .)
-      # We don't have UI tests yet, so we just make sure the `/` route returns 200.
-      - run: curl --fail http://localhost:4000
-      - run: docker stop ui
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.21.0
+      - run: npm install
+      - run: npm test
```
mjlangan, line 17: @codeviking do you think it's worth keeping that "see if it builds" test?

.github/workflows/ci.yml
```diff
@@ -28,7 +24,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.20.0
+          node-version: 12.21.0
```
pauls-ai2, line 27: Noice! Now you match S2!

ui/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  testPathIgnorePatterns: [
+    "<rootDir>/build/",
+    "<rootDir>/node_modules/"
```
pauls-ai2, line 6: Nit: mixing of `'` and `"`

ui/src/utils/ui.test.ts
```diff
@@ -0,0 +1,27 @@
+import { extractArxivId } from './ui';
```
pauls-ai2, line 1: I _LOVE_ that your tests are sitting next to your code. One pattern that FB uses internally when they use Jest is to put them in a `__test__` folder. That way they live next to the code, but don't get in the way. Might be worth considering.

ui/jest.config.js
```diff
@@ -0,0 +1,8 @@
...
+  ]
+};
```
pauls-ai2, line 8: Is it worth setting a glob for test files as `/**/*.test.*`? Or maybe that's default...I can't remember.

ui/jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
```
pauls-ai2, line 1: I _think_ Jest can use `.ts` for configs, which should be typed in your editor...but I could be wrong.

ui/src/api/api.ts
```diff
@@ -83,7 +83,7 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
     type: ENTITY_API_ALL
   } : {};
   const data = await doGet(
-    axios.get(`/api/v0/papers/arxiv:${arxivId}/entities`, { params })
+    axios.get(`/api/v0/papers/arxiv:${encodeURIComponent(arxivId)}/entities`, { params })
```
pauls-ai2, line 86: Looks like `axiso` doesn't intend to even support this for you. Boo on them!
https://github.com/axios/axios/issues/706#issuecomment-280777539

ui/src/utils/ui.ts
```diff
@@ -390,3 +390,12 @@ export function getElementCoordinates(element: HTMLElement) {
     clientHeight: element.clientHeight,
   };
 }
+
+/**
+ * TODO(andrewhead): Handle the case of arXiv publications that have multiple versions. How do we
+ * make sure we're querying for the same version of paper data as the paper that was opened?
+ */
+ export function extractArxivId(url: string): string | undefined {
+  const matches = url.match(/arxiv\.org\/pdf\/(.*)(?:\.pdf).*$/) || [];
```
pauls-ai2, line 399: If you change this to the `.exec()` API, then it won't fail if the string is `undefined`/`null`.
```
const matches = /arxiv\.org\/pdf\/(.*)(?:\.pdf).*$/.exec(url) || [];
```

Also, don't need the `.*$` part in the regex, because `.*` means any chars and `$` means end of the string.

If you have access to the `url` library, you could do this too:
```
import URL from 'url';
...
const urlObj = new URL(urlStr);
const matches = /^\/pdf\/(.*)(?:\.pdf$)/.exec(urlObj.pathname) || [];
```

That would make sure it's a valid URL, and that the pathname ends in `.pdf` (and not `.pdf.zip` or something). It'll strip off the query string for you.

.github/workflows/ci.yml
codeviking: I do, for two reasons:

1. I think some folks work on the code without using a container -- which means they might inadvertently break `docker build`.
2. If it can't build the deploy will fail (it has to run `docker build`) as part of that process. So making sure things build before merging them helps the commit jitter that occurs when someone fixes forward.

ui/src/utils/ui.ts
```diff
@@ -390,3 +390,12 @@ export function getElementCoordinates(element: HTMLElement) {
...
+  return matches[1];
```
pauls-ai2, line 400: (This is a question for me, not about your code)
Is your code prefering `undefined` or `null` for these situations?

Seems like `null` is the right option, because it means you have a result which has no value, while `undefined` means you don't even have a result (as in, you didn't return anything). I know TS can have configs for saying a type can have `null` for a value, or to require it to be declared explicitly (as you have done with `undefined`).

ui/package.json
```diff
@@ -7,7 +7,8 @@
     "clean": "rimraf ./build/",
     "lint": "eslint '**/*.{js,ts,tsx,json}' && echo '💫  Lint complete.'",
     "lint:fix": "eslint '**/*.{js,ts,tsx,json}' --fix && npx sort-package-json  && echo '🛠  Lint --fix complete'",
-    "start": "webpack serve --hot --mode development"
+    "start": "webpack serve --hot --mode development",
+    "test": "jest"
```
pauls-ai2, line 11: Might also be worth adding a something for `jest --watch`, so tests can re-run on save. Or I guess you can do `npm test --watch` if that's what you want.

pauls-ai2 (approved):

ui/src/utils/ui.ts
andrewhead: I believe I authored the first version of this code. Given your explanation, I agree that, were I to write this again, I would have it return `null` if no arXiv ID was present.

ui/package.json
```diff
@@ -68,6 +71,7 @@
     "reselect": "^4.0.0",
     "rimraf": "^3.0.2",
     "style-loader": "^2.0.0",
+    "ts-jest": "^26.5.4",
```
andrewhead, line 74: Is `ts-jest` the only plugin needed to run tests on the `tsx` files (i.e., do we need any additional transpilers to handle the JSX?)

andrewhead (approved):
Looks great, and I am excited to have the first tests in the UI code!!!

ui/jest.config.js
mjlangan: You're right, it does. This was created by their bootstrapping script and I think it just plonks down a .js without seeing what language your project uses. I'll switch 👍

ui/jest.config.js
mjlangan: Yeeeah this project has a mix of quote styles throughout--one reason I'd really like to add Prettier

ui/package.json
mjlangan: Yep, that's it.

ui/package.json
mjlangan: Yeah, `npm test -- --watch` can do this. I'll update the README with a note about this.

ui/src/api/api.ts
mjlangan: They'd need a fancier interface than simply accepting the URL as a string, so I can kinda understand their POV. Still annoying that you have to remember to do it though.

ui/src/utils/ui.ts
mjlangan: There's a fair bit of inconsistency in the code between returning `undefined` or `null` in places, but it would be good to consolidate on one. I'd also go `null`

ui/src/utils/ui.ts
mjlangan: Cool, yeah I've confirmed that the final `.*$` is no longer needed (I think the real bug in the regex must've been making the .pdf extension optional).

ui/package.json
andrewhead: sounds good!

PR #263, opened 2021-04-08
validate math arxiv IDs

mjlangan:
We have a paper with ID `math/0008020v2` in the system. Turns out, the API never had a regex to handle IDs like this. Fortunately, it also turns out that `hapi` is decoding the encoded slash for us (in the URL, this ID looks like `math%2F0008020v2` so it doesn't change the route), so we don't need to manually handle that.

### DIFF
--- a/api/src/types/validation.ts
+++ b/api/src/types/validation.ts
@@ -24,6 +24,7 @@ const s2paperId = Joi.string().pattern(/[a-f0-9]{40}/);
  * https://arxiv.org/help/arxiv_identifier.
  */
 const currentArxivFormat = Joi.string().pattern(/arxiv:[0-9]{2}[0-9]{2}.[0-9]+(v[0-9]+)?/);
+const arxivMathFormat = Joi.string().pattern(/arxiv:math\/[0-9]{5,}(v[0-9]+)/);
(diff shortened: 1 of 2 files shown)


cmwilhelm (approved):

api/src/types/validation.ts
```diff
@@ -24,6 +24,7 @@ const s2paperId = Joi.string().pattern(/[a-f0-9]{40}/);
  * https://arxiv.org/help/arxiv_identifier.
  */
 const currentArxivFormat = Joi.string().pattern(/arxiv:[0-9]{2}[0-9]{2}.[0-9]+(v[0-9]+)?/);
+const arxivMathFormat = Joi.string().pattern(/arxiv:math\/[0-9]{5,}(v[0-9]+)/);
```
ca16, line 27: I think there are arXiv ids in a similar format that start with other things apart from 'math' (looking at this https://arxiv.org/help/arxiv_identifier). (Not objecting to this addition, more of an FYI.)

mjlangan: Oh! So it looks like this is actually a bug in the existing `olderArxivFormat` regex where it's assuming that optional bit after the subject (`.GT`) is always present.

PR #264, opened 2021-04-09
Fixes around fetching citations

mjlangan:
This PR addresses two issues:
1. The list of S2 paper SHAs we pass to the `getPapers` function is not deduplicated, as every instance of a paper citation is its own citation entity linking to the same SHA. The Reader API does dedupe the IDs it receives before fetching the papers from the public S2 API, but the many duplicate IDs emitted from the reader result in more requests to the Reader API being sent than are necessary.
2. If the paper has no citation entities, we never turn off the loading spinner for citations. Whoops.

As a concrete example for item 1, with the paper `https://arxiv.org/pdf/2104.01111v1.pdf`:
- Current prod code results in 8 requests to the Reader API's /papers endpoint
- This branch only makes 4 requests to the Reader API

### DIFF
--- a/ui/src/ScholarReader.tsx
+++ b/ui/src/ScholarReader.tsx
@@ -729,46 +729,44 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
   }
 
   loadDataFromApi = async (): Promise<void> => {
-    if (this.props.paperId !== undefined) {
-      if (this.props.paperId.type === "arxiv") {
-        this.setState({
-          areCitationsLoading: true
-        });
-        const loadingStartTime = performance.now();
-        const entities = await api.getDedupedEntities(this.props.paperId.id, true);
-        this.setState({
-          entities: stateUtils.createRelationalStoreFromArray(entities, "id"),
-        });
+    if (this.props.paperId !== undefined && this.props.paperId.type === "arxiv") {
+      this.setState({
+        areCitationsLoading: true
+      });
+      const loadingStartTime = performance.now();
+      const entities = await api.getDedupedEntities(this.props.paperId.id, true);
+      this.setState({
+        entities: stateUtils.createRelationalStoreFromArray(entities, "id"),
+      });
 
-        const citationS2Ids = entities
-          .filter(isCitation)
-          .map((c) => c.attributes.paper_id)
-          .filter((id) => id !== null)
-          .map((id) => id as string);
-        if (citationS2Ids.length >= 1) {
-          const papers = (await api.getPapers(citationS2Ids)).reduce(
-            (papers, paper) => {
-              papers[paper.s2Id] = paper;
-              return papers;
-            },
-            {} as { [s2Id: string]: Paper }
-          );
-          this.setState({ papers, areCitationsLoading: false });
-        } else {
-          this.setState({ areCitationsLoading: false });
-        }
+      const citationS2Ids = entities
+        .filter(isCitation)
+        .map((c) => c.attributes.paper_id)
+        .filter((id) => id !== null)
+        .map((id) => id as string);
+      if (citationS2Ids.length >= 1) {
+        const papers = (await api.getPapers(citationS2Ids)).reduce(
+          (papers, paper) => {
+            papers[paper.s2Id] = paper;
+            return papers;
+          },
+          {} as { [s2Id: string]: Paper }
+        );
+        this.setState({ papers, areCitationsLoading: false });
+      } else {
+        this.setState({ areCitationsLoading: false });
+      }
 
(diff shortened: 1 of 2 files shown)


ui/src/api/api.ts
```diff
@@ -49,11 +49,17 @@ export async function listPapers(offset: number = 0, size: number = 25) {
  */
 export async function getPapers(s2Ids: string[]) {
   const PAPER_REQUEST_BATCH_SIZE = 50;
-  var s2IdsBatch = [];
-  var promises = [];
+  let s2IdsBatch = [];
+  let promises = [];
+  let dedupedIds: string[] = [];
```
pauls-ai2, line 54: Why no `const`? I guess you are signaling that they are writable?

ui/src/api/api.ts
```diff
@@ -49,11 +49,17 @@ export async function listPapers(offset: number = 0, size: number = 25) {
...
+  s2Ids.forEach(id => {
+    if (dedupedIds.indexOf(id) < 0) {
```
cmwilhelm, line 56: mah n^2

cmwilhelm (approved):
LGTM. Just to be clear, the browser app code doesn't care about order at all right?

ui/src/api/api.ts
```diff
@@ -49,11 +49,17 @@ export async function listPapers(offset: number = 0, size: number = 25) {
...
+      dedupedIds.push(id);
+    }
+  });
 
-  for (let i = 0; i < s2Ids.length; i += PAPER_REQUEST_BATCH_SIZE) {
-    s2IdsBatch.push(s2Ids.slice(i, i + PAPER_REQUEST_BATCH_SIZE));
+  for (let i = 0; i < dedupedIds.length; i += PAPER_REQUEST_BATCH_SIZE) {
```
pauls-ai2, line 61: I have a few recommendations on the code, but I'll let you decide what parts you want to steal.
```
async function getPapers(s2Ids) {
  // Use a set to remove dupes
  const dedupeIds = [...new Set(s2Ids)];
  
  // Use .reduce() to break into chunks
  const s2IdsBatch = dedupeIds.reduce((chunks, s2Id, index) => {
    const chunkIndex = Math.floor(index / PAPER_REQUEST_BATCH_SIZE);
    if (chunks[chunkIndex]) {
      chunks[chunkIndex].push(s2Id);
    } else {
      chunks[chunkIndex] = [s2Id];
    }
    return chunks;
  }, []);
  
  // Await the results within the async function, use async function in map
  const results = await Promise.all(s2IdsBatch.map(async (s2Ids) => {
    const result = await doGet(s2Ids);
    // ...process result
    return result;
  }));
  
  // Use `[].concat(...) to flatten array
  return [].concat(...results);
}
```

ui/src/api/api.ts
```diff
@@ -49,11 +49,17 @@ export async function listPapers(offset: number = 0, size: number = 25) {
  */
 export async function getPapers(s2Ids: string[]) {
```
pauls-ai2, line 50: As written, this function doesn't need to be `async`, since it doesn't `await` anything. It might clean the code up a bit if you did, though.

ui/src/ScholarReader.tsx
```diff
@@ -754,6 +754,8 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
             {} as { [s2Id: string]: Paper }
           );
           this.setState({ papers, areCitationsLoading: false });
+        } else {
```
pauls-ai2, line 757: Is this really **3** levels of nested `if` statements? I suspect it would be worth breaking this function up into smaller functions...but maybe that should be done when adding a store abstraction.

pauls-ai2 (approved):

ui/src/api/api.ts
mjlangan: I was operating on some outdated info that Set still had some browser incompatibilities (it still does, but with IE10 and who cares about that?). Yeah, a Set would be way better than doing what I'm doing here, and your suggestion is way cleaner.

ui/src/ScholarReader.tsx
mjlangan: Well the top two levels should really just be 1...but yes, when we add stores we'll have opportunities to move this code out of the top component & break it up.

ui/src/api/api.ts
mjlangan: Now it does! 🧙

mjlangan:
> LGTM. Just to be clear, the browser app code doesn't care about order at all right?

Correct. It maps the retrieved data by ID for lookup.

PR #265, opened 2021-04-12
Extract font information for tokens

andrewhead:
As part of my work to improve the speed of symbol detection, I needed our equation processor to know:

1. **Each letter / digit (i.e., token) that makes up a symbol**. This is because the new symbol detector needs to find the position of each letter/digit/token in a symbol separately before merging these positions to find the position of the symbol that contains it (e.g., find both an 'x' and 'i' before trying to find the symbol 'x_i'.

2. **What LaTeX fonts are applied to those tokens (e.g., `\mathbf`, `\mathcal`, etc.)**. As the new symbol detector is based on matching images of expected tokens against the document, it needs to know what fonts should be used to format each token when rendering it to an image.

3. **The MathML for each of those tokens**. This can be used as a unique key for tokens, allowing us to extract images for each token (i.e., each letter-font combination) only once.

A related PR is our change to the KaTeX code (see https://github.com/allenai/KaTeX/pull/5) which ensures that, most of the time, symbols are split into single letters and annotated with font informationwhen parsed by KaTeX.

Then, in this PR, the paper processing code was updated to process the output of KaTeX to extract tokens in such a way that tasks 1, 2, and 3 are satisfied. Most of the changes appear in the `parse_equation.py` file.

See proof that MathML and font information is being extracted for tokens from this clipping of the `data/*-detected-equation-tokens/entities.csv` file, generated by running `python scripts/run_pipeline.py --entities symbols --extraction-only --arxiv-ids 1601.00978`:

image

### DIFF
--- a/data-processing/common/parse_equation.py
+++ b/data-processing/common/parse_equation.py
@@ -659,6 +659,7 @@ def _is_atomic_token(element: Tag) -> bool:
         "-",
         "/",
         "*",
+        "∗",
         "∙",
         "⋅",
         "⋅",


data-processing/common/parse_equation.py
```diff
@@ -533,21 +597,39 @@ def _extract_tokens(element: Tag) -> List[Token]:
...
         )
 
     return tokens
 
 
+def _extract_font_macros(element: Tag) -> Tuple[str, ...]:
+    """
+    Get the list of TeX styling macros applied to a token represented by a MathML element. Relies
+    on the MathML having an "s2:font-macros" attribute on the elements for tokens, where
+    each font macro is separated by an '&' character.
```
andrewhead, line 625: Kyle asks if we can add a pointer to the place in the KaTeX code that creates the font-macros property.

kyleclo (approved):
lgtm - comments in person

andrewhead:
Kyle asks if, long term, we make sure there are not issues caused by KaTeX normalizing the macros used for styling (e.g., converting `\tt` to `\texttt`)

data-processing/node/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Utilies for data mining for the Scholar reading application.",
   "main": "index.js",
   "scripts": {
-    "prepare-katex": "git clone https://github.com/allenai/KaTeX.git && cd KaTeX && git submodule init && git submodule update && yarn && npm link",
+    "prepare-katex": "git clone https://github.com/allenai/KaTeX.git && cd KaTeX && git checkout e768b38090fa226895135121e33a2dad97721edb && git submodule init && git submodule update && yarn && npm link",
```
cmwilhelm, line 7: Might be useful to add a git tag to KaTex so you can reference it more succinctly / clearly than a sha here.

cmwilhelm (approved):
Just to clarify, this shouldn't change anything about the data stored at persistence/upload time correct? Only during extraction? 

I ask because we've been using the `mathml` attribute as a disambiguating ID where data is duplicated across multiple instances of the same symbol.

PR #266, opened 2021-04-14
Fix the paper API

mjlangan:
The /papers endpoint in the reader API was adding a bunch of JSON that was being stripped out in the client. I accidentally clobbered the UI side of that with commit https://github.com/allenai/scholarphi/commit/81c0b8f40ad8fe092b45cd80f99289f162199942, which broke the display of citations in the overlays because the API response was not matching the expected shape. Of course, there were no JS errors because ScholarPhi in general returns `undefined` and happily carries on without logging anything, but that's another story.

My change here gets rid of the additional JSON in the response that was just being stripped out in the client, which gets citations working again.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -128,13 +128,8 @@ export const plugin = {
         });
         const data = (await s2Api.getPapers(uniqueIds, config.s2.apiKey))
           .filter((paper) => paper !== undefined)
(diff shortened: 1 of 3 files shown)


pauls-ai2 (approved):
Fix forward!
tenor-152750087

PR #267, opened 2021-04-15
Mjlangan/lazy citations

mjlangan:
`LazyCitationGloss` is essentially a copy-paste of `CitationGloss` but with added code to load the paper data on-demand, so it still has the library and feature voting buttons.

**GET /paper/{s2id}**
I wanted to try implementing an endpoint using `hapi` with error handling, which required bringing in the `@hapi/boom` library. Per the docs, `Boom.internal` exceptions don't return the message in the API response, but they do record it in the server logs. Unfortunately, using `AxiosPromise` as the return type, since this makes it easier for me to work with their errors, means I can't re-use the same paper cache in the current API layer. Since lazy-loading should result in way less requests to the S2 public API (now we're only requesting paper data for a paper when someone clicks it, vs. requesting all papers' data at load time), I think this is an acceptable trade-off, and we can always add it back in if performance or request load become an issue.

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -15,8 +15,6 @@ import {
 } from "./types/api";
 import * as validation from "./types/validation";
 import * as conf from "./conf";
-import { AxiosError } from "axios";
-import { S2ApiError } from "./types/s2-api";
 
 interface ApiOptions {
   connection: Connection;
@@ -120,17 +118,18 @@ export const plugin = {
       path: "paper/{s2Id}",
       handler: async (request) => {
         const s2Id = request.params.s2Id;
-        const paper = await s2Api.getPaperUncached(s2Id, config.s2.apiKey)
-        .then(p => p.data)
-        .catch((error: AxiosError<S2ApiError>) => {
+        let resp = null;
+        try {
+          resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
+        } catch (error) {
           if (error.response?.status === 404) {
             throw Boom.notFound(`not found: ${s2Id}`);
           } else if (error.response?.status === 500) {
             throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
           }
           throw Boom.internal(`Internal Error: ${error}`);
-        });
-        return paper;
+        }
+        return resp?.data;
       },
       options: {
         validate: {
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -1,3 +1,4 @@
+import * as Boom from "@hapi/boom";
 import * as Hapi from "@hapi/hapi";
 import * as HapiAuthBearer from 'hapi-auth-bearer-token';
 import * as Joi from "@hapi/joi";
@@ -14,6 +15,8 @@ import {
 } from "./types/api";
 import * as validation from "./types/validation";
 import * as conf from "./conf";
+import { AxiosError } from "axios";
+import { S2ApiError } from "./types/s2-api";
 
 interface ApiOptions {
   connection: Connection;
@@ -110,6 +113,32 @@ export const plugin = {
         const response: Paginated<PaperWithIdInfo> = { ...papers, ...{ rows: mergedPapers } };
         return response;
       },
+    });
+
+    server.route({
+      method: "GET",
+      path: "paper/{s2Id}",
+      handler: async (request) => {
+        const s2Id = request.params.s2Id;
+        const paper = await s2Api.getPaperUncached(s2Id, config.s2.apiKey)
+        .then(p => p.data)
+        .catch((error: AxiosError<S2ApiError>) => {
+          if (error.response?.status === 404) {
+            throw Boom.notFound(`not found: ${s2Id}`);
+          } else if (error.response?.status === 500) {
+            throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
+          }
+          throw Boom.internal(`Internal Error: ${error}`);
+        });
+        return paper;
+      },
+      options: {
+        validate: {
+          params: Joi.object({
+            s2Id: validation.s2paperId
+          })
+        }
+      }
     });
 
     server.route({
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -118,18 +118,19 @@ export const plugin = {
       path: "paper/{s2Id}",
       handler: async (request) => {
         const s2Id = request.params.s2Id;
-        let resp = null;
         try {
-          resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
+          const resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
+          return resp.data;
         } catch (error) {
-          if (error.response?.status === 404) {
-            throw Boom.notFound(`not found: ${s2Id}`);
-          } else if (error.response?.status === 500) {
-            throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
+          switch(error.response?.status) {
+            case 404:
+              throw Boom.notFound(`not found: ${s2Id}`);
+            case 500:
+              throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
+            default:
+              throw Boom.internal(`Internal Error: ${error}`);
           }
-          throw Boom.internal(`Internal Error: ${error}`);
         }
-        return resp?.data;
       },
       options: {
         validate: {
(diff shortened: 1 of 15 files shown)


api/src/s2-api.ts
```diff
@@ -90,7 +90,62 @@ async function getPaper(s2Id: string, apiKey?: string): Promise<Paper | undefine
...
+  }
+
+  const conf: AxiosRequestConfig = {
+    headers: {
+      "user-agent":
+        "Semantic Reader API Client (https://scholarphi.semanticscholar.org)",
+    },
+    // cast to AxiosTransformer[] or else the base type is never[]
+    // TODO: See if a newer version of typescript is better at resolving empty array types
+    transformResponse: ([] as AxiosTransformer[])
```
mjlangan, line 127: I don't remember having to do stuff like this back on Beaker, which uses TS 3.8, so I think if we took an updated version of TS there would be less of this going on.

ui/src/components/entity/citation/LazyCitationGloss.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <div className="paper-summary">
+        <div className="paper-summary__loading-msg">
+          Error
+        </div>
+        {this.state.error}
+      </div>
+    );
+    
+    // @ts-ignore why is this resolving to null?!
+    const paper: Paper = this.state.paper;
```
mjlangan, line 71: Another weird TS issue. This is typed as `Nullable<Paper>` in state, which means `Paper | null`, but down here TS seems pretty confident it is only `null`. Maybe there's some bug with the react types?

api/src/api.ts
```diff
@@ -112,6 +115,32 @@ export const plugin = {
...
     });
 
+    server.route({
+      method: "GET",
+      path: "paper/{s2Id}",
+      handler: async (request) => {
+        const s2Id = request.params.s2Id;
+        const paper = await s2Api.getPaperUncached(s2Id, config.s2.apiKey)
+        .then(p => p.data)
+        .catch((error: AxiosError<S2ApiError>) => {
```
pauls-ai2, line 125: Why use `.then()` and `.catch()` with `await`? May as well use `try/catch` instead, since it's easier to read.

mjlangan: Old habit 👴 

try/catch would work better, yeah.

api/src/s2-api.ts
```diff
@@ -90,7 +90,62 @@ async function getPaper(s2Id: string, apiKey?: string): Promise<Paper | undefine
...
+ * @returns
+ */
+export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Paper> {
+  if (!apiKey) {
+    console.warn(`
+      WARNING: The S2 Public API Key isn't set. If the backend makes more than 100 requests to
+      the public API in 5 minutes it'll be rate limited and might stop working.
+
+      See the README.md for more information about geting an API key.
+    `);
```
pauls-ai2, line 117: I like having these kind of warnings in development, but this is a large string that won't be able to be shrunk much (gzip might help?) in minifying. Just food for thought, since I know your bundles are huge.

api/src/types/api.ts
```diff
@@ -12,6 +12,8 @@
  * into other projects, all of the types are available to the client code.
  */
 
+export type Nullable<T> = T | null;
```
pauls-ai2, line 15: Oh, so you _are_ using strict nulls! Good. :)

api/src/types/validation.ts
```diff
@@ -18,7 +18,7 @@ export const logEntry = Joi.object({
   data: Joi.object().unknown(true).default(null),
 });
 
-const s2paperId = Joi.string().pattern(/^[a-f0-9]{40}$/);
+export const s2paperId = Joi.string().pattern(/^[a-f0-9]{40}$/);
```
pauls-ai2, line 21: Are uppercase ok?

Also, shouldn't the `p` in `paper` be capitalized?

ui/src/api/api.ts
```diff
@@ -90,6 +91,13 @@ export async function getEntities(arxivId: string, getAllEntities?: boolean) {
   return ((data as any).data || []) as Entity[];
 }
 
+export async function getPaper(s2Id: string): Promise<Nullable<Paper>> {
+  const data = await doGet(
+    axios.get<Paper>(`/api/v0/paper/${encodeURIComponent(s2Id)}`, {
+    }));
```
pauls-ai2, line 97: Super nit: indenting

api/src/api.ts
```diff
@@ -112,6 +113,33 @@ export const plugin = {
...
+        let resp = null;
+        try {
+          resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
+        } catch (error) {
+          if (error.response?.status === 404) {
+            throw Boom.notFound(`not found: ${s2Id}`);
+          } else if (error.response?.status === 500) {
+            throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
+          }
+          throw Boom.internal(`Internal Error: ${error}`);
```
pauls-ai2, line 130: Why not use a switch? It would be a lot more readable that way.
```
switch (error.response?.status) {
  case 404:
    throw Boom.notFound(`not found: ${s2Id}`);
  case 500:
    throw Boom.internal(`S2 API Error accessing paper ${s2Id}`);
  default:
    throw Boom.internal(`Internal Error: ${error}`);
}
```

api/src/api.ts
```diff
@@ -112,6 +113,33 @@ export const plugin = {
...
 
+    server.route({
+      method: "GET",
+      path: "paper/{s2Id}",
+      handler: async (request) => {
+        const s2Id = request.params.s2Id;
+        let resp = null;
+        try {
+          resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
+        } catch (error) {
```
pauls-ai2, line 124: You can return within the try:
```
try { 
  const resp = await s2Api.getPaperUncached(s2Id, config.s2.apiKey);
  return resp?.data;
} catch (error) {
```

Unless the style of this project is putting returns at the bottom, I guess.

api/src/s2-api.ts
```diff
@@ -90,7 +90,62 @@ async function getPaper(s2Id: string, apiKey?: string): Promise<Paper | undefine
   );
   const response = await axios.get<S2ApiPaper>(`${apiOrigin}/paper/${s2Id}`, conf);
   const paper = toPaper(s2Id, response.data);
-  cache.set(s2Id, paper);
+  // only cache if the request is successful to avoid papers being permanent errors
+  if (response.status === 200) {
```
pauls-ai2, line 94: Wouldn't a non-200 throw in the `await` above?

api/src/s2-api.ts
```diff
@@ -90,7 +90,62 @@ async function getPaper(s2Id: string, apiKey?: string): Promise<Paper | undefine
   );
   const response = await axios.get<S2ApiPaper>(`${apiOrigin}/paper/${s2Id}`, conf);
```
pauls-ai2, line 91: Could this change to `getPaperUncached()`? That way you don't have to duplicate as much code?

api/src/types/s2-api.ts
```diff
@@ -17,3 +17,11 @@ interface S2ApiAuthor {
   name: string;
   url: string;
 }
+
+export interface S2ApiError {
+  error: string;
+}
+
+export function isS2ApiError(d: any): d is S2ApiError {
```
pauls-ai2, line 25: Oh wow! Is that a way to assert something about the code? I would have assumed this returned `boolean`, so that's pretty awesome!

pauls-ai2 (approved):
Could comments, but nothing blocking. Consider them optional.

api/src/s2-api.ts
mjlangan: Come to think of it, I think this issue was actually addressed by using the LRU cache in here (it used to be that this was just an in-memory map that would store the response for each paper for all eternity). Going to revert this change.

api/src/types/s2-api.ts
mjlangan: Type guards! https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

It's effectively returning a `boolean`, yeah, but to TypeScript, the boolean tells the type-checker that it can assume exactly the type of the object within that scope. It's usually used like:
```typescript
if (isFoo(foo)) {
  foo.onlyFooThing(); // this is good!
}
(foo as Foo).onlyFooThing(); // casting required here
```

api/src/s2-api.ts
mjlangan: Yeah, I think there's a way to collapse the two now that you mention it.

PR #268, opened 2021-04-15
Use arxiv's compiler

stefanc-ai2:
https://github.com/allenai/scholar/issues/27321

Changes:
1. Use TeXLive from Arxiv instead of from `apt-get`.
2. Change config.ini to point to where TeXLive is installed

Testing: manual testing
- can compile 2 latex files that were previously failed to compile
- can compile 1 latex file that was successfully compiled

Also, the image size balloons to 18GB. We're gonna need a bigger boat, @ca16.

### DIFF
--- a/data-processing/Dockerfile
+++ b/data-processing/Dockerfile
@@ -23,6 +23,7 @@ RUN apt-get install -y wget
 RUN wget https://arxiv-web-static1.s3.amazonaws.com/semantic_scholar_20210412/arXivTeXLive2020.tar.gz
 RUN tar -xf arXivTeXLive2020.tar.gz
 RUN echo "I" | perl 2020/install-tl
+RUN rm -r arXivTeXLive2020.tar.gz 2020
 
 # Install Perl dependencies
 SHELL ["/bin/bash", "-c"]
--- a/data-processing/Dockerfile
+++ b/data-processing/Dockerfile
@@ -53,10 +53,6 @@ RUN sed -i '/pattern="PDF"/s/rights="none"/rights="read | write"/' /etc/ImageMag
 # Install GhostScript
 RUN apt-get install -y ghostscript
 
-# Install pip requireements
-COPY requirements.txt ./
-RUN pip install -r requirements.txt
-
 # Install Node.js dependencies
 WORKDIR /data-processing/node
 RUN apt-get install -y git
@@ -69,6 +65,10 @@ RUN npm run install-katex
 
 # Install vim for when we inevitably want to inspect files
 RUN apt-get install -y vim
+
+# Install pip requireements
+COPY requirements.txt ./
+RUN pip install -r requirements.txt
 
 # Copy over the source code
 WORKDIR /data-processing
--- a/data-processing/Dockerfile
+++ b/data-processing/Dockerfile
@@ -45,11 +45,6 @@ RUN python get-pip.py --force-reinstall
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash
 RUN apt-get install -y nodejs
 
-# Let ImageMagick to read and write PDFs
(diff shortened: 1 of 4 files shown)


data-processing/Dockerfile
```diff
@@ -20,6 +17,19 @@ RUN perlbrew init
 RUN perlbrew --notest install perl-5.22.4
 RUN perlbrew install-cpanm
 
+# Install LaTeX suite
+# This download will take an extremely long time to install, so we install it first
+RUN apt-get install -y wget
+RUN wget https://arxiv-web-static1.s3.amazonaws.com/semantic_scholar_20210412/arXivTeXLive2020.tar.gz
```
stefanc-ai2, line 23: I also have this copied to `s3://ai2-s2-scholarphi-pipeline-prod/arxiv-compiler/` for fun/backup, but getting it from S3 bucket requires AWS creds and cli. Any idea of a simple way to get it from our S3? If not, maybe it's okay to leave it this way?

data-processing/requirements.txt
```diff
@@ -3,7 +3,7 @@ beautifulsoup4==4.8.0
 lxml==4.4.2
 nltk==3.4.5
 numpy==1.17.4
-opencv-python==3.4.8.29
+opencv-python-headless==3.4.8.29
```
stefanc-ai2, line 6: Ran into missing `libSM.so` error. Getting the `headless` version is the solution according to https://stackoverflow.com/a/56738577.

data-processing/Dockerfile
ca16: > for fun

🤣

data-processing/Dockerfile
```diff
@@ -35,35 +45,26 @@ RUN python get-pip.py --force-reinstall
 RUN curl -sL https://deb.nodesource.com/setup_12.x | bash
 RUN apt-get install -y nodejs
 
-# Let ImageMagick to read and write PDFs
-RUN sed -i '/pattern="PDF"/s/rights="none"/rights="read | write"/' /etc/ImageMagick-6/policy.xml
```
ca16, line 39: Do we not need this anymore?

stefanc-ai2: Talked to Andrew and likely we don't need this anymore

ca16 (approved):
Exciting...

data-processing/Dockerfile
stefanc-ai2: Turns out, we still need this. Latex compilation fails without this.

data-processing/Dockerfile
cmwilhelm: Kind of gross but we could parameterize the loc with a build arg for s3 path. Use the one that's hardcoded here as a default but allow us to use our own within our land of credentials (which all boxes building this should have).

cmwilhelm (approved):
Fun fun fun.

Before merge I would recommend a couple things:

1. Verify that the scholarphi-pipeline image can still be built and run correctly when extending this.
2. Run the pipeline on a few test papers before/after to make sure the output is the same.

data-processing/Dockerfile
stefanc-ai2: Ah I forgot that this is an open-sourced project. Changing it to S2's S3 will be problematic.

stefanc-ai2:
> Fun fun fun.
> 
> Before merge I would recommend a couple things:
> 
> 1. Verify that the scholarphi-pipeline image can still be built and run correctly when extending this.
> 2. Run the pipeline on a few test papers before/after to make sure the output is the same.

1. Verified `scholarphi-pipeline` image can still be built and run
2. Compiled latex files from `2009.12303v4` and `2012.09607v2` using old and new latex compilers and the resulting pdfs are identical.

PR #269, opened 2021-04-15
Add warning if references exist, but can't match any with S2 reference data

stefanc-ai2:
For a paper, we extract references from the latex and try to match them with S2's reference data. This is to add a warning if we can't match any reference. This is to get more visibility into `no_citations` error. With this, we can know whether that error is caused by scholarphi not being able to extract any reference vs being able to extract reference but can't match it.

Screen Shot 2021-04-16 at 11 02 38

### DIFF
--- a/data-processing/entities/citations/commands/resolve_bibitems.py
+++ b/data-processing/entities/citations/commands/resolve_bibitems.py
@@ -97,7 +97,7 @@ class ResolveBibitems(ArxivBatchCommand[MatchTask, BibitemMatch]):
         if item.bibitems and ref_match_count == 0:
             logging.warning(f"Could not match any reference for paper {item.arxiv_id}.")
         else:
-            logging.debug(f"Matched {ref_match_count} references for paper {item.arxiv_id}.")
+            logging.info(f"Matched {ref_match_count} references for paper {item.arxiv_id}.")
 
     def save(self, item: MatchTask, result: BibitemMatch) -> None:
         resolutions_dir = directories.arxiv_subdir("bibitem-resolutions", item.arxiv_id)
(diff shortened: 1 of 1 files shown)


data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -52,7 +52,7 @@ def load(self) -> Iterator[MatchTask]:
                     references_path,
                     arxiv_id,
                 )
-                return
+                continue
```
ca16, line 55: Making sure I understand the changes here (the return -> continues)... Before, we would stop yielding MatchTasks as soon as we found an id where we found that it didn't have references or bibitems (not exactly sure what they mean in this context, though a guess is that references are the things we got from the S2 API and bibitems are the things scholarphi pulled from the paper through its pipeline?). But this means we may have lost MatchTasks for other ids after it that do have this information. These changes fix that in when we don't find references or bibitems for an id, we go onto the next id rather than returning immediately. Is that right?

stefanc-ai2: > references are the things we got from the S2 API and bibitems are the things scholarphi pulled from the paper through its pipeline

Yes, that's correct.

I think it's a bug to have `return` instead of `continue`. Though it doesn't matter much now because we've only been processing one paper at a time (through the flag `--one-paper-at-a-time`).

data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -92,6 +94,10 @@ def process(self, item: MatchTask) -> Iterator[BibitemMatch]:
                     bibitem.id_,
                     item.arxiv_id,
                 )
+        if item.bibitems and ref_match_count == 0:
```
ca16, line 97: So we'll only warn if none of them could be matched. But if at least one is matched, we won't warn. Is that right? And I guess, for my own understanding, why is matching none more serious than not matching some?

stefanc-ai2: Good question. I'm also looking to get more visibility into `no_citations` error. With this, we can know whether that error is caused by scholarphi not being able to extract any reference vs being able to extract reference but can't match it. Reading this, I made an update to the code to better suit that. I also updated the PR description.

Screen Shot 2021-04-16 at 11 02 38

ca16 (approved):

PR #271, opened 2021-04-21
Remove "save to library"

mjlangan:
This is implemented using S2 private APIs, which we don't want to be using, so it seems like a good idea to remove it prior to getting a bunch of public eyes on the reader. I wanted to do this a separate change from my previous PR so it could be more easily reverted if we come up with a way to support this on the S2 side without needing the reader to call private APIs.

### DIFF
--- a/ui/src/ScholarReader.tsx
+++ b/ui/src/ScholarReader.tsx
@@ -84,8 +84,6 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
       entities: null,
       lazyPapers: new Map(),
 
-      userLibrary: null,
-
       pages: null,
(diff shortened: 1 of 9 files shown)


pauls-ai2 (approved):
With this set of API calls going away, does that mean we could host scholarphi on it's own domain? Should we consider registering `scholarphi.org` or something for this?

mjlangan:
> With this set of API calls going away, does that mean we could host scholarphi on it's own domain? Should we consider registering `scholarphi.org` or something for this?

I suppose we could. Dunno if there are other reasons we might still want it to be under semanticscholar.org, though.

PR #273, opened 2021-04-23
add back proper SPP service  dependency

kyleclo:


data-processing/entities/sentences_pdf/commands/fetch_spp_jsons.py
```diff
@@ -36,6 +38,10 @@ class SppJsonTask:
 
 
 class FetchSppJsons(ArxivBatchCommand[ArxivId, None]):
+    def __init__(self, args: Any) -> None:
+        super().__init__(args)
+        self._spp_host_url = os.getenv("S2_SPP_HOST_URL", None)
```
cmwilhelm, line 43: We'll want the port as well. The env vars are named:

`SPP_HOST` and `SPP_PORT`

PR #282, opened 2021-05-03
Optional highlighting

cmwilhelm:
Per @kyleclo 's request, an optional query parameters can now 
be passed that will highlight entities that we care about manually evaluating.

Simple append one or more of the following to your URL:

```
&annotateEquations
&annotateSymbols
&annotateCitations
```

Results in things like:

Screen Shot 2021-05-03 at 12 21 49 PM


kyleclo (approved):
lgtm

PR #284, opened 2021-05-03
UI: Disable tooltips for symbols; underline all top-level symbols

andrewhead:
According to current plans, symbol definition tooltips will not be shown in the upcoming ScholarPhi demo. This PR makes two changes to the UI to make the interactions better support declutter-centric interactions with symbols.

1. The definition tooltip (which would not have any definitions to show) is now hidden (toggled off using the new `symbolGlossesEnabled` setting).
2. Underlines are now placed underneath _all_ top-level symbols, and not just the symbols that have definitions available (as technically, none of the symbols will have definitions available anymore). This should make it clearer to readers what symbols they can interact with.

Before:

image

After (no tooltip; `L_CGAN` symbol which did not have a definition before is now underlined): 

image

### DIFF
--- a/ui/src/ScholarReader.tsx
+++ b/ui/src/ScholarReader.tsx
@@ -1115,6 +1115,7 @@ export default class ScholarReader extends React.PureComponent<Props, State> {
                           this.state.citationGlossesEnabled
                         }
                         termAnnotationsEnabled={this.state.termGlossesEnabled}
+                        symbolGlossesEnabled={this.state.symbolGlossesEnabled}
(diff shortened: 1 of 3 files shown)


andrewhead:
Closing this PR in favor of an alternative implementation proposed by @cmwilhelm :

> another option rather than changing the flow is to change the hover text for the tooltip icons to something like "Definitions coming soon"

The alternative implementation has the benefits that it (1) minimizes redesign of interaction (2) reduces the amount of code that needs to be chanced (3) clearly cues what additional functionality will be brought to the system soon

PR #285, opened 2021-05-04
UI: show underlines for all top-level symbols (not just those with de…

andrewhead:
To indicate to readers where it is possible to use "declutter", I swapped a settings flag to make sure that all top-level symbols in a paper are underlined (and hence indicate that they are interactive), instead of just those that have definitions.

Before:

image

After (see additional underlines beneath 'z' and 'y'):

image


cmwilhelm (approved):
agreed now that definitions are gone. let's merge it.

PR #289, opened 2021-05-05
Turns off definitional features.

cmwilhelm:
* hides primer page glossary behind
      flag that is off for CHI
* hides definitional bits of the symbol
      gloss behind new FF
* turns off equation diagrams using existing
      FF

Screen Shot 2021-05-05 at 12 39 45 PM


mjlangan (approved):
LGTM

andrewhead:
LGTM. Might pair nicely with https://github.com/allenai/scholarphi/pull/285, which shows underlines for all symbols for which 'declutter' can be triggered, rather than just the symbols for which definitions are available. Maybe we should consider merging that one too?

andrewhead (approved):
(see comment above this approval)

PR #290, opened 2021-05-05
Primer page copy and links updated.

cmwilhelm:
* updates front page copy based on marketing material for CHI
* links through to CHI marketing page (dead link right now)
* adds google form
  * factors out feedback helpers and creates new feedback link component
* adds link through to arxiv pdf

Screen Shot 2021-05-05 at 2 39 24 PM


ui/src/components/primer/PrimerPage.tsx
```diff
@@ -90,84 +93,66 @@ class PrimerPage extends React.PureComponent<Props> {
     const terms = isEntitiesLoaded(entities) ? glossaryTerms(entities) : [];
     const symbols = isEntitiesLoaded(entities) ? glossarySymbols(entities) : [];
 
-    return ReactDOM.createPortal(
+    const instructions = (
```
mjlangan, line 96: nit: To me, this is a code smell that `instructions` should instead be a new class of component that is imported into the file.

ui/src/utils/feedback.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function createFeedbackLink(
+  paperId?: PaperId | undefined,
+  extraContext?: object
+) {
+  /*
+   * The URL and field ids below are generated by Google. The identifiers
+   * are opaque to us, which means they could change (say, if we modify
+   * the assocaited form). If this proves to be an issue we'll have to think
+   * through alternative feedback mechanisms.
```
mjlangan, line 12: one idea could be making this configurable in the API/UI at deployment time (like the config file used by the API), but that's a later problem and not a now problem.

mjlangan (approved):
LGTM. One minor suggestion that we can probably split out the instructions into a new component to make things a little more concise & clean, but not blocking.

PR #296, opened 2021-05-25
Reduce use of 'master' language

ckarenz:
Use of the word "master" in many technology contexts is broadly accepted to be [racist language](https://www.google.com/search?q=master+racist+language).  This change removes some of these terms in favor of "main".


mjlangan (approved):
LGTM

ckarenz:
Note there are still a few instances around logs and TPU primaries.  These look like they might require migrations, so I addressed only the safest instances.

PR #303, opened 2021-07-08
Add a pod disruption budget, and surge during deploys.

codeviking:
This change does two things:

1. Adds a `PodDisruptionBudget`, which tells Kubernetes that there
   must always be `minAvailable` replicas up. In the case of production,
   this guarantees that there will always be at least one replica to
   handle incoming traffic. The main outcome of this being that
   there shouldn't be downtime anymore during cluster wide Kubernetes
   updates.

2. Updates the deployment specification so that Kubernetes can "surge"
   while deploying changes. This makes deployments faster by quickly
   deploying a full set of new replicas -- currently it does so
   in a rolling fashion. This makes things a little faster, and also
   ensures there's additional capacity during deploys.

### DIFF
--- a/.skiff/webapp.jsonnet
+++ b/.skiff/webapp.jsonnet
@@ -147,6 +147,12 @@ function(
             }
         },
         spec: {
+            strategy: {
+                type: 'RollingUpdate',
+                rollingUpdate: {
+                    maxSurge: replicas // This makes deployments faster.
+                }
+            },
             revisionHistoryLimit: 3,
             replicas: replicas,
             selector: {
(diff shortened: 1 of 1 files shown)


mjlangan (approved):
LGTM, yeah PDBs are handy aren't they? ;)

.skiff/webapp.jsonnet
```diff
@@ -147,6 +147,12 @@ function(
             }
         },
         spec: {
+            strategy: {
+                type: 'RollingUpdate',
```
pauls-ai2, line 151: Is the load balancer stateless? Does this mean during a deploy, two calls to the same API could hit different servers, and therefore different versions? Maybe that doesn't matter for these types of apps.

codeviking: Yup, there's no session stickiness or anything like that.

For our workloads the trade of makes sense. In cases where it matters (say Beaker, LB or the like) we just make sure to maintain backwards compatibility.

PR #315, opened 2021-08-31
WIP: Use the latest UI base image.

codeviking:
...which might have been *just* built.

I'm verifying this as we speak. I'll remove the draft status once I've confirmed it resolves the problem.


codeviking:
Turns out we don't need this, as the application simply split into two. Resolving that *should* fix the problem.

PR #316, opened 2021-09-01
Add preset for papers for 2021 SAB paper review

andrewhead:
This PR adds a new preset to the deployed Semantic Reader UI. When deployed, this will let readers to append the `&preset=sab-2021` parameter to access symbol and abbreviation definition functionality.

None of the underlying functionality of the app is changed. The main purpose of this PR is to provide the Allen AI a switch for the Scientific Advisory board to preview experimental ML features, without turning them on for all readers of Semantic Reader.

### DIFF
--- a/ui/src/settings.ts
+++ b/ui/src/settings.ts
@@ -117,10 +117,12 @@ const PRESETS: Preset[] = [
   {
     key: "demo",
     primerInstructionsEnabled: false,
-    useDefinitionsForDiagramLabels: true,
+    // useDefinitionsForDiagramLabels: true,
     termGlossesEnabled: true,
     declutterEnabled: true,
     equationDiagramsEnabled: true,
+    symbolUnderlineMethod: "defined-symbols",
+    definitionsInSymbolGloss: true
   },
   {
     key: "sab",
--- a/ui/src/settings.ts
+++ b/ui/src/settings.ts
@@ -117,10 +117,16 @@ const PRESETS: Preset[] = [
   {
     key: "demo",
     primerInstructionsEnabled: false,
-    // useDefinitionsForDiagramLabels: true,
+    useDefinitionsForDiagramLabels: true,
     termGlossesEnabled: true,
     declutterEnabled: true,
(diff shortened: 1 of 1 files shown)


ui/src/settings.ts
```diff
@@ -122,6 +122,14 @@ const PRESETS: Preset[] = [
     declutterEnabled: true,
     equationDiagramsEnabled: true,
   },
+  {
```
pauls-ai2, line 125: Now that there are more than a handful of presets, it might be worth adding a comment above each to describe what the preset is and what it's for. That'll help both with readability of this file, and help out who even ends up removing the preset.

pauls-ai2 (approved):

ui/src/settings.ts
```diff
@@ -122,6 +122,14 @@ const PRESETS: Preset[] = [
...
+    key: "sab-2021",
+    primerInstructionsEnabled: false,
+    termGlossesEnabled: true,
+    declutterEnabled: true,
+    symbolUnderlineMethod: "defined-symbols",
```
carolinepaulic, line 130: This is minor, but it could be helpful to add an enum for these `symbolUnderlineMethod` types since they're used a few times throughout this file.

carolinepaulic (approved):

ui/src/settings.ts
andrewhead: Thank you @carolinepaulic ! This is a great idea and in fact there is already a type defined for this property in the settings type declaration:

https://github.com/allenai/scholarphi/blob/b9c50f2b406e65752aad142cf8d44275a11ea662/ui/src/settings.ts#L51

https://github.com/allenai/scholarphi/blob/b9c50f2b406e65752aad142cf8d44275a11ea662/ui/src/components/entity/EntityAnnotationLayer.tsx#L23

This is not an enum per se though it does check the string. Would there be advantage to using an enum type?

ui/src/settings.ts
andrewhead: Done!

PR #318, opened 2021-09-16
Write citations output to file

ca16:
Related to https://github.com/allenai/scholar/issues/28994. 

As part of getting citations info from scholarphi piped through the s2airs system, we want to get citations info into the annotation store. We can adapt the wrapper system (scholarphi-pipeline) to do this, if it has access to the output that the inner scholarphi code produces. 

The simplest way that I found to do this while I was working on https://github.com/allenai/scholar/issues/28297 was to just dump the output as json into a file in a location provided by the caller.

This PR adapts the DatabaseUploadCommand a little, making it more of a 'save output' kind of thing (I haven't renamed it to try to keep this PR small, and avoid changing more than is necessary). The idea is the user can specify whether they want output saved to a db (what we've been doing up until now), saved to a file (what we want for s2airs), or both. To begin with though, only the citations extension of DatabaseUploadCommand will know how to actually save to a file. Other commands extending DatabaseUploadCommand should error out if the user tries to specify that they want output saved to a file for those. To get an idea of how we could extend this to all entities, check out this branch: https://github.com/allenai/scholarphi/compare/chi-2021-demo...chloea-try-producing-json-take-1 (note: it does have some extra stuff around pulling arxiv pdfs that isn't relevant to this, and the saving to file stuff is less neat, and the 'both' option doesn't exist, but hopefully it gets the idea around extending to other entities across).

The plan is that the existing scholarphi-pipeline system would use the 'both' option when processing in citations only mode, so that we can continue to support the current reader, and so that we can put stuff into the annotation store as we build out s2airs.

Testing done:
I added some tests and ran pytest:
```
# pytest --all
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.7.5, pytest-5.3.1, py-1.10.0, pluggy-0.13.1
rootdir: /data-processing, inifile: pytest.ini, testpaths: tests
plugins: cov-2.5.1
collected 214 items                                                                                                                                                                                                                          

tests/test_bounding_box.py ..................                                                                                                                                                                                          [  8%]
tests/test_colorize_sentences.py .....                                                                                                                                                                                                 [ 10%]
tests/test_colorize_tex.py ........                                                                                                                                                                                                    [ 14%]
tests/test_compile.py ........                                                                                                                                                                                                         [ 18%]
tests/test_extract_definitions.py .................ssss                                                                                                                                                                                [ 28%]
tests/test_locate_symbols.py ...                                                                                                                                                                                                       [ 29%]
tests/test_match_symbols.py ......                                                                                                                                                                                                     [ 32%]
tests/test_normalize_tex.py ..............                                                                                                                                                                                             [ 38%]
tests/test_parse_equation.py .........................                                                                                                                                                                                 [ 50%]
tests/test_parse_tex.py ..............................................                                                                                                                                                                 [ 71%]
tests/test_sanitize_equation.py ....                                                                                                                                                                                                   [ 73%]
tests/test_scan_tex.py .....                                                                                                                                                                                                           [ 76%]
tests/test_string.py .............                                                                                                                                                                                                     [ 82%]
tests/test_unpack.py ....                                                                                                                                                                                                              [ 84%]
tests/test_visual_validate.py ......                                                                                                                                                                                                   [ 86%]
tests/common/test_fetch_arxiv.py ......                                                                                                                                                                                                [ 89%]
tests/common/test_upload_entities.py ...                                                                                                                                                                                               [ 91%]
tests/common/commands/test_database.py .........                                                                                                                                                                                       [ 95%]
tests/common/commands/test_fetch_arxiv_sources.py ..                                                                                                                                                                                   [ 96%]
tests/common/commands/test_fetch_s2_data.py ........                                                                                                                                                                                   [100%]

============================================================================================================== warnings summary ==============================================================================================================
/usr/local/lib/python3.7/dist-packages/wandb/util.py:37
/usr/local/lib/python3.7/dist-packages/wandb/util.py:37
  /usr/local/lib/python3.7/dist-packages/wandb/util.py:37: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import namedtuple, Mapping, Sequence

/usr/local/lib/python3.7/dist-packages/wandb/vendor/graphql-core-1.1/graphql/type/directives.py:55
  /usr/local/lib/python3.7/dist-packages/wandb/vendor/graphql-core-1.1/graphql/type/directives.py:55: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    assert isinstance(locations, collections.Iterable), 'Must provide locations for directive.'

entities/definitions/commands/tokenize_sentences.py:111
  /data-processing/entities/definitions/commands/tokenize_sentences.py:111: DeprecationWarning: invalid escape sequence \s
    "EQUATION_DEPTH_0_START\s*(.*?)\s*EQUATION_DEPTH_0_END",

entities/glossary_terms/colorize.py:24
  /data-processing/entities/glossary_terms/colorize.py:24: DeprecationWarning: invalid escape sequence \S
    first_nonspace = re.search("\S", term.tex)

entities/glossary_terms/colorize.py:32
  /data-processing/entities/glossary_terms/colorize.py:32: DeprecationWarning: invalid escape sequence \S
    last_nonspace = re.search("\S(?=\s*$)", term.tex)

tests/test_extract_definitions.py::test_model_extracts_simple_definitions
tests/test_extract_definitions.py::test_model_extracts_simple_definitions
tests/test_extract_definitions.py::test_model_extracts_simple_definitions
tests/test_extract_definitions.py::test_model_extracts_simple_definitions
tests/test_extract_definitions.py::test_model_extracts_simple_definitions
tests/test_extract_definitions.py::test_model_extracts_nickname_before_symbol
tests/test_extract_definitions.py::test_model_extracts_nickname_before_symbol
tests/test_extract_definitions.py::test_model_extracts_nickname_symbol_filter
tests/test_extract_definitions.py::test_model_extracts_nickname_symbol_filter
tests/test_extract_definitions.py::test_model_extract_abbreviation_acronym
tests/test_extract_definitions.py::test_model_extract_abbreviation_acronym
tests/test_extract_definitions.py::test_extract_abbreviation_acronym
tests/test_extract_definitions.py::test_extract_abbreviation_acronym
tests/test_extract_definitions.py::test_model_extract_abbreviation_shortened_word
tests/test_extract_definitions.py::test_model_extract_abbreviation_shortened_word
tests/test_extract_definitions.py::test_extract_abbreviation_shortened_word
tests/test_extract_definitions.py::test_extract_abbreviation_shortened_word
  /usr/local/lib/python3.7/dist-packages/catalogue.py:138: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    for entry_point in AVAILABLE_ENTRY_POINTS.get(self.entry_point_namespace, []):

tests/test_extract_definitions.py::test_model_extracts_simple_definitions
  /usr/local/lib/python3.7/dist-packages/catalogue.py:126: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    for entry_point in AVAILABLE_ENTRY_POINTS.get(self.entry_point_namespace, []):

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================================== 210 passed, 4 skipped, 24 warnings in 187.01s (0:03:07) ===========================================================================================
```

I ran one paper through using the file only format and just citations:
```
root@635a2b2ca566:/data-processing# mkdir hi
root@635a2b2ca566:/data-processing# python scripts/run_pipeline.py --one-paper-at-a-time --arxiv-ids "1601.00978v1" --output-form file --output-dir hi --entities citations
...
root@635a2b2ca566:/data-processing# ls hi/
citations.jsonl
root@635a2b2ca566:/data-processing# cat hi/citations.jsonl 
{"version": "v0", "data": [{"id_": "bandeira_automatic_2010-0", "type_": "citation", "bounding_boxes": [{"left": 0.47549019607843135, "top": 0.30808080808080807, "width": 0.0016339869281045752, "height": 0.008838383838383838, "page": 0, "tex_path": "N/A", "entity_id": "bandeira_automatic_2010-0"}], "data": {"key": "bandeira_automatic_2010", "paper_id": "9769ac9d79a75cfefd3731978b9f319d68345ab5"}, "relationships": null}, {"id_": "bandeira_automatic_2010-1", "type_": "citation", "bounding_boxes": [{"left": 0.4035947712418301, "top": 0.3686868686868687, "width": 0.0016339869281045752, "height": 0.008838383838383838, "page": 0, "tex_path": "N/A", "entity_id": "bandeira_automatic_2010-1"}], "data": {"key": "bandeira_automatic_2010", "paper_id": "9769ac9d79a75cfefd3731978b9f319d68345ab5"}, "relationships": null}, {"id_": "bandeira_automatic_2010-2", "type_": "citation", "bounding_boxes": [{"left": 0.8137254901960784, "top": 0.5404040404040404, "width": 0.0016339869281045752, "height": 0.008838383838383838, "page": 1, "tex_path": "N/A", "entity_id": "bandeira_automatic_2010-2"}], "data": {"key": "bandeira_automatic_2010", "paper_id": "9769ac9d79a75cfefd3731978b9f319d68345ab5"}, "relationships": null}, {"id_": "bandeira_automatic_2010-3", "type_": "citation", "bounding_boxes": [{"left": 0.25326797385620914, "top": 0.5896464646464646, "width": 0.0016339869281045752, "height": 0.008838383838383838, "page": 1, "tex_path": "N/A", "entity_id": "bandeira_automatic_2010-3"}], "data": {"key": "bandeira_automatic_2010", "paper_id": "9769ac9d79a75cfefd3731978b9f319d68345ab5"}, "relationships": null}, {"id_": "urbach_automatic_2009-0", "type_": "citation", "bounding_boxes": [{"left": 0.2957516339869281, "top": 0.3686868686868687, "width": 0.004901960784313725, "height": 0.008838383838383838, "page": 0, "tex_path": "N/A", "entity_id": "urbach_automatic_2009-0"}], "data": {"key": "urbach_automatic_2009", "paper_id": "b4caa57f2321298073d651e95507052ccf2ed66d"}, "relationships": null}, {"id_": "urbach_automatic_2009-1", "type_": "citation", "bounding_boxes": [{"left": 0.23202614379084968, "top": 0.6944444444444444, "width": 0.004901960784313725, "height": 0.008838383838383838, "page": 1, "tex_path": "N/A", "entity_id": "urbach_automatic_2009-1"}], "data": {"key": "urbach_automatic_2009", "paper_id": "b4caa57f2321298073d651e95507052ccf2ed66d"}, "relationships": null}, {"id_": "ding_sub-kilometer_2011-0", "type_": "citation", "bounding_boxes": [{"left": 0.4166666666666667, "top": 0.3686868686868687, "width": 0.006535947712418301, "height": 0.008838383838383838, "page": 0, "tex_path": "N/A", "entity_id": "ding_sub-kilometer_2011-0"}], "data": {"key": "ding_sub-kilometer_2011", "paper_id": "447369e1c509e7e552220fa00eec93e474e1167b"}, "relationships": null}, {"id_": "krizhevsky_imagenet_2012-0", "type_": "citation", "bounding_boxes": [{"left": 0.16830065359477125, "top": 0.5202020202020202, "width": 0.006535947712418301, "height": 0.008838383838383838, "page": 0, "tex_path": "N/A", "entity_id": "krizhevsky_imagenet_2012-0"}], "data": {"key": "krizhevsky_imagenet_2012", "paper_id": "abd1c342495432171beb7ca8fd9551ef13cbd0ff"}, "relationships": null}, {"id_": "krizhevsky_imagenet_2012-1", "type_": "citation", "bounding_boxes": [{"left": 0.1568627450980392, "top": 0.38257575757575757, "width": 0.006535947712418301, "height": 0.008838383838383838, "page": 1, "tex_path": "N/A", "entity_id": "krizhevsky_imagenet_2012-1"}], "data": {"key": "krizhevsky_imagenet_2012", "paper_id": "abd1c342495432171beb7ca8fd9551ef13cbd0ff"}, "relationships": null}]}
```
I ran it with the file only option and a non-citations entity (equations). It failed (as we would expect):
```
python scripts/run_pipeline.py --one-paper-at-a-time --arxiv-ids "1601.00978v1" --output-form file --output-dir hi --entities equations
...
AssertionError: C does not know how to write to a file.
```

Note: this PR is for merging into the chi-2021-demo branch, as that is what the scholarphi-pipeline system is still running off of.


data-processing/common/upload_entities.py
```diff
@@ -309,3 +313,53 @@ def resolve_model(entity_identifier: EntityIdentifier) -> Optional[Entity]:
...
+        json.dump(to_write, output_file)
+
+
+def save_entities(
+    s2_id: S2Id,
+    arxiv_id: ArxivId,
+    entity_infos: List[EntityUploadInfo],
+    data_version: Optional[int],
+    output_details: OutputDetails,
+    filename: str,
```
stefanc-ai2, line 348: How about having `output_path` that includes filename (and not just the dir) in `OutputDetails`? Then you don't need this param.

stefanc-ai2 (approved):
One minor suggestion. Otherwise, looks good!

data-processing/common/upload_entities.py
ca16: Yeah I can see the appeal of that but I think it doesn't quite work...
Each upload command will have one instance of OutputDetails. However, a single upload command could write to more than one file. E.g. if you take a look at the `data-processing/entities/definitions/upload.py` file in the [branch](https://github.com/allenai/scholarphi/compare/chi-2021-demo...chloea-try-producing-json-take-1) that shows how this could be extended to the other scholarphi inner pipelines, the command using `upload_definitions()` writes to two files: `definitions_terms.jsonl` and `definitions_symbols.jsonl`.
(We could probably change that code to collect all the entities for that pipeline before uploading them to the db or writing them to a file, so that we could write them to a single file, but I think I'd rather have this extra argument, and keep it possible to write entities as we go).

stefanc-ai2: Oh I see. Nvm then.

data-processing/common/upload_entities.py
```diff
@@ -1,8 +1,12 @@
+import dataclasses
+import json
 import logging
+import os
 from typing import Any, Dict, List, Optional, Tuple
 
 from peewee import fn
 
+from common.commands.database import OutputDetails
```
andrewhead, line 9: nit pick: the code base tends to only import from `common` into `common.commands` modules, and not vice versa, to avoid cases of circular dependencies. Could this OutputDetails be moved into the 'upload_entities.py' file?

data-processing/common/commands/database.py
```diff
@@ -1,11 +1,39 @@
...
-from typing import Any
+from enum import Enum
+import logging
+from typing import Any, Optional
 
 from common.commands.base import ArxivBatchCommand, I, R
 from common.models import setup_database_connections
 
 
+class OutputForm(Enum):
```
andrewhead, line 11: Style nit pick: would it be possible to make this a `Literal` type instead of an Enum?

data-processing/common/upload_entities.py
```diff
@@ -309,3 +313,53 @@ def resolve_model(entity_identifier: EntityIdentifier) -> Optional[Entity]:
...
+        FORMAT_VERSION,
+    )
+    to_write = {
+        "version": FORMAT_VERSION,
+        "data": [dataclasses.asdict(entity_info) for entity_info in entity_infos],
+    }
+    already_exists_msg = (
+        f"File {output_file_name} already exists. "
+        + "Not overwriting. Entity infos will not be written."
+    )
```
andrewhead, line 335: The default behavior in this repo is to overwrite while printing a warning message. We could do this behavior instead, and in that case I would be curious of your thinking about if this is a better way to do it, in case it makes sense for us to change the behavior elsewhere in the repo.

data-processing/common/upload_entities.py
```diff
@@ -309,3 +313,53 @@ def resolve_model(entity_identifier: EntityIdentifier) -> Optional[Entity]:
                     )
 
     return models
+
+
+def write_to_file(entity_infos: List[EntityUploadInfo], output_file_name: str) -> None:
+    # an attempt to make life easier if we change how we want to format this
+    FORMAT_VERSION = "v0"
```
andrewhead, line 320: Love the versioning! It will be great to be able to distinguish between data created by different versions of the pipeline.

data-processing/common/upload_entities.py
```diff
@@ -309,3 +313,53 @@ def resolve_model(entity_identifier: EntityIdentifier) -> Optional[Entity]:
...
+    entity_infos: List[EntityUploadInfo],
+    data_version: Optional[int],
+    output_details: OutputDetails,
+    filename: str,
+) -> None:
+
+    if output_details.save_to_file():
+        logging.info("Saving to file...")
+        # should always be true, but let's just make sure
+        assert output_details.output_dir is not None, "Expected a defined output dir!"
```
andrewhead, line 354: For these assertions, I am wondering whether they should be created as assertion messages, or as their own exception type. I recall russellr implemented error catchers in the data-pipeline runner code for common types of errors so that they could be surfaced as counters in DataDog. Do you know whether we would be able to count up assertion failures based on their messages from the pipeline runner code?

data-processing/scripts/run_pipeline.py
```diff
@@ -335,8 +338,29 @@ def run_commands_for_arxiv_ids(
         type=str,
         help="Base url for definition model paths."
     )
+
+    parser.add_argument(
+        "--output-form",
+        type=str,
+        choices=[c.value for c in OutputForm],
```
andrewhead, line 345: An alternative way of having users provide this argument could be as a variable-length argument with one word per format (e.g., `--output-form file db` instead of `both`). This might be helpful if we think we are going to add another type of output form in the future. Do you think that is likely?

data-processing/scripts/run_pipeline.py
```diff
@@ -335,8 +338,29 @@ def run_commands_for_arxiv_ids(
...
+    parser.add_argument(
+        "--output-dir",
+        type=str,
+        help="The directory in which to write the outputs."
+    )
+
     args = parser.parse_args()
 
+    # check that output arguments make sense
+    OutputDetails.validate(output_form=args.output_form, output_dir=args.output_dir)
```
andrewhead, line 362: Upon assertion failure, it might make sense to throw an error using Python's [`parser.error`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.error) method if validation fails. Though this is a minor point and not that important.

data-processing/common/commands/database.py
```diff
@@ -1,11 +1,39 @@
...
+        self.output_dir = output_dir
+
+    @staticmethod
+    def validate(output_form: str, output_dir: Optional[str]) -> None:
+        msg = "If the output form is 'file' or 'both', an output dir must also be specified."
+        cond = (output_form in [OutputForm.FILE.value, OutputForm.BOTH.value]) == \
+            (output_dir is not None)
+        assert cond, msg
+
+    def save_to_db(self) -> bool:
```
andrewhead, line 30: Style nit pick: could the name be modified to suggest that it is a check on whether a save can be performed, rather than performing that save (e.g., `can_save_to_db`)?

andrewhead (approved):
Chloe, this looks wonderful! I know this code is pretty tangled, and you did an amazing job of finding all of the places to add in code. As far as I am concerned, this could be merged as is. I left a few comments in places where there are some unspoken style conventions of the repo that might be worth considering, though none of them are blockers.

data-processing/common/commands/database.py
```diff
@@ -14,10 +42,23 @@ class DatabaseUploadCommand(ArxivBatchCommand[I, R], ABC):
...
 
+        if self.output_details.save_to_db():
+            logging.info("Setting up db connection as we expect to upload output to the db.")
+            setup_database_connections(
+                schema_name=args.schema, create_tables=args.create_tables
+            )
+
+        if self.output_details.save_to_file():
+            logging.info("We will be writing output to files.")
+            msg = f"{self.__class__.__name__} does not know how to write to a file."
```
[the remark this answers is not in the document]
ca16: Yep that seems much better. I've made the update.  Now:

```
python scripts/run_pipeline.py --one-paper-at-a-time --arxiv-ids "1601.00978v1" --output-form file --output-dir hi --entities equations
...
2021-09-18 01:17:04,022 [INFO]: We will be writing output to files.
Traceback (most recent call last):
  File "scripts/run_pipeline.py", line 537, in <module>
    filtered_commands, [arxiv_id], args
  File "scripts/run_pipeline.py", line 90, in run_commands_for_arxiv_ids
    command = CommandCls(command_args)
  File "/data-processing/common/commands/database.py", line 34, in __init__
    assert self.can_write_to_file(), msg
AssertionError: upload-equations does not know how to write to a file.
2021-09-18 01:17:04,030 [INFO]: Internal process exited
```

data-processing/common/upload_entities.py
ca16: Yes!

data-processing/common/commands/database.py
ca16: Can you say a little more about this? 

It seems like if I were to use `Literal`, I would want to do something like:
```diff
diff --git a/data-processing/common/upload_entities.py b/data-processing/common/upload_entities.py
index 32d0e12..0403fce 100644
--- a/data-processing/common/upload_entities.py
+++ b/data-processing/common/upload_entities.py
@@ -1,10 +1,9 @@
 import dataclasses
-from enum import Enum
 import json
 import logging
 import os
 from typing import Any, Dict, List, Optional, Tuple
-# from typing_extensions import Literal
+from typing_extensions import Literal
 
 from peewee import fn
 
@@ -21,30 +20,28 @@ from common.types import (
     S2Id,
 )
 
-# OutputForm = Literal["db", "file"]
-
-class OutputForm(Enum):
-    DB = "db"
-    FILE = "file"
+DBOutputForm = Literal["db"]
+FileOutputForm = Literal["file"]
+OutputForm = Literal[DBOutputForm, FileOutputForm]
 
 
 class OutputDetails:
-    def __init__(self, output_forms: List[str], output_dir: Optional[str]):
+    def __init__(self, output_forms: List[OutputForm], output_dir: Optional[str]):
         OutputDetails.validate(output_forms=output_forms, output_dir=output_dir)
-        self.output_forms: List[OutputForm] = [OutputForm(output_form) for output_form in output_forms]
-        self.output_dir: Optional[str] = output_dir
+        self.output_forms = output_forms
+        self.output_dir = output_dir
 
     @staticmethod
-    def validate(output_forms: List[str], output_dir: Optional[str]) -> None:
+    def validate(output_forms: List[OutputForm], output_dir: Optional[str]) -> None:
         msg = "Please include either both 'file' as an output form and an output dir, or neither."
-        cond = (OutputForm.FILE.value in output_forms) == (output_dir is not None)
+        cond = (FileOutputForm in output_forms) == (output_dir is not None)
         assert cond, msg
 
     def can_save_to_db(self) -> bool:
-        return OutputForm.DB in self.output_forms
+        return DBOutputForm in self.output_forms
 
     def can_save_to_file(self) -> bool:
-        return OutputForm.FILE in self.output_forms
+        return FileOutputForm in self.output_forms
 
 
 def get_or_create_data_version(paper_id: str) -> int:
diff --git a/data-processing/scripts/run_pipeline.py b/data-processing/scripts/run_pipeline.py
index 2eb5a73..efdcdc9 100644
--- a/data-processing/scripts/run_pipeline.py
+++ b/data-processing/scripts/run_pipeline.py
@@ -27,7 +27,7 @@ from common.commands.store_pipeline_log import StorePipelineLog
 from common.commands.store_results import DEFAULT_S3_LOGS_BUCKET, StoreResults
 from common.fetch_arxiv import FetchFromArxivException
 from common.make_digest import make_paper_digest
-from common.upload_entities import OutputDetails, OutputForm
+from common.upload_entities import DBOutputForm, OutputDetails, OutputForm
 from common.types import PipelineDigest
 from entities.definitions.commands.detect_definitions import DetectDefinitions
 
@@ -342,10 +342,9 @@ if __name__ == "__main__":
 
     parser.add_argument(
         "--output-forms",
-        type=str,
+        type=OutputForm,
         nargs='+',
-        choices=[c.value for c in OutputForm],
-        default=[OutputForm.DB.value],
+        default=[DBOutputForm],
         help=(
             f"How we want the output to be stored. "
             + "If you include file, you must also provide a value for output-dir."
```
But that doesn't work.

The version I can get to work is something like this:
```diff
diff --git a/data-processing/common/upload_entities.py b/data-processing/common/upload_entities.py
index 32d0e12..e18bb49 100644
--- a/data-processing/common/upload_entities.py
+++ b/data-processing/common/upload_entities.py
@@ -1,10 +1,9 @@
 import dataclasses
-from enum import Enum
 import json
 import logging
 import os
 from typing import Any, Dict, List, Optional, Tuple
-# from typing_extensions import Literal
+from typing_extensions import Literal
 
 from peewee import fn
 
@@ -21,30 +20,26 @@ from common.types import (
     S2Id,
 )
 
-# OutputForm = Literal["db", "file"]
-
-class OutputForm(Enum):
-    DB = "db"
-    FILE = "file"
+OutputForm = Literal["db", "file"]
 
 
 class OutputDetails:
-    def __init__(self, output_forms: List[str], output_dir: Optional[str]):
+    def __init__(self, output_forms: List[OutputForm], output_dir: Optional[str]):
         OutputDetails.validate(output_forms=output_forms, output_dir=output_dir)
-        self.output_forms: List[OutputForm] = [OutputForm(output_form) for output_form in output_forms]
-        self.output_dir: Optional[str] = output_dir
+        self.output_forms = output_forms
+        self.output_dir = output_dir
 
     @staticmethod
-    def validate(output_forms: List[str], output_dir: Optional[str]) -> None:
+    def validate(output_forms: List[OutputForm], output_dir: Optional[str]) -> None:
         msg = "Please include either both 'file' as an output form and an output dir, or neither."
-        cond = (OutputForm.FILE.value in output_forms) == (output_dir is not None)
+        cond = ("file" in output_forms) == (output_dir is not None)
         assert cond, msg
 
     def can_save_to_db(self) -> bool:
-        return OutputForm.DB in self.output_forms
+        return "db" in self.output_forms
 
     def can_save_to_file(self) -> bool:
-        return OutputForm.FILE in self.output_forms
+        return "file" in self.output_forms
 
 
 def get_or_create_data_version(paper_id: str) -> int:
diff --git a/data-processing/scripts/run_pipeline.py b/data-processing/scripts/run_pipeline.py
index 2eb5a73..1cd28e9 100644
--- a/data-processing/scripts/run_pipeline.py
+++ b/data-processing/scripts/run_pipeline.py
@@ -27,7 +27,7 @@ from common.commands.store_pipeline_log import StorePipelineLog
 from common.commands.store_results import DEFAULT_S3_LOGS_BUCKET, StoreResults
 from common.fetch_arxiv import FetchFromArxivException
 from common.make_digest import make_paper_digest
-from common.upload_entities import OutputDetails, OutputForm
+from common.upload_entities import OutputDetails
 from common.types import PipelineDigest
 from entities.definitions.commands.detect_definitions import DetectDefinitions
 
@@ -344,8 +344,8 @@ if __name__ == "__main__":
         "--output-forms",
         type=str,
         nargs='+',
-        choices=[c.value for c in OutputForm],
-        default=[OutputForm.DB.value],
+        choices=["db", "file"],
+        default=["db"],
         help=(
             f"How we want the output to be stored. "
             + "If you include file, you must also provide a value for output-dir."
```
Which seems worse than the current Enum option, because I've got "db" and "file" lots of places (which feels like it would be more prone to typos coming about).

Note: this is the first time I've seen python `Literal`s, so I'd love to know if I'm not using them quite right, and there's a way of using them that does work better!

data-processing/common/upload_entities.py
ca16: I think I tend towards the better safe than sorry route, avoiding overwriting existing things, and placing the responsibility on the user to clean up before running things, if they want output to be stored in the same place.
But I don't feel particularly strongly about this with scholarphi, and I also think that especially with this kind of thing, where things can run for a really long time, it could be quite frustrating if a pipeline completes after a number of hours, only to let you know that you don't get to see any of the output because you forgot to clear the relevant directory or point it somewhere else. I've updated the behavior to print a warning message, then overwrite:
```
...
Saving to file...
2021-09-18 01:01:59,783 [INFO]: About to write 9 entity infos to hi/citations.jsonl (version: v0).
2021-09-18 01:01:59,783 [WARNING]: File hi/citations.jsonl already exists. It will be overwritten.
2021-09-18 01:01:59,785 [INFO]: Finished running command upload-citations
...
```

data-processing/common/upload_entities.py
ca16: Just to make sure we're on the same page, I see this as more about the versioning of the format of what gets written. So even if we updated say the citations extraction code, I wouldn't expect _this_ version to necessarily change (though s2airs would track that through a different path). And the other side of that is that this version could change even if the extraction code doesn't, say if we decide we want our json output to have a different nesting structure, or change some of the key names.

data-processing/common/upload_entities.py
ca16: Yep, I believe the way the pipeline runner code counts things up is by looking at what is printed out (except for a couple of cases where the count is based on exit code), so I think it should be able to count these up as well.

Sidenote: this assertion in particular is something that we shouldn't ever hit (because the way the input argument validation is set up, this should never happen), but sometimes I sprinkle things like this in to guard against code changes potentially changing the expectation that this would never happen.

data-processing/scripts/run_pipeline.py
ca16: I think it's possible. And I think the alternative is cleaner too :) thanks for the suggestion! I've made the change.

data-processing/scripts/run_pipeline.py
ca16: Fine with me! Now:
```
python scripts/run_pipeline.py --one-paper-at-a-time --arxiv-ids "1601.00978v1" --output-form file  --entities citations
wandb: WARNING W&B installed but not logged in.  Run `wandb login` or set the WANDB_API_KEY env variable.
usage: run_pipeline.py [-h] [-v] [--config CONFIG] [--log-prefix LOG_PREFIX]
                       [--entities {sentences-pdf,sentences,symbols,definitions,citations,equations,glossary-terms} [{sentences-pdf,sentences,symbols,definitions,citations,equations,glossary-terms} ...]]
                       [--commands {fetch-arxiv-sources,fetch-s2-metadata,unpack-sources,compile-tex,normalize-tex,compile-normalized-tex,fetch-arxiv-pdf,raster-pages,detect-sentences,extract-bibitems,resolve-bibitems,locate-bounding-boxes-for-citation-fragments,locate-citations,upload-citations,detect-equations,locate-bounding-boxes-for-equations,upload-equations,detect-glossary-terms,extract-contexts-for-glossary-terms,locate-bounding-boxes-for-glossary-terms,upload-glossary-terms,extract-symbols,find-symbol-matches,extract-contexts-for-symbols,locate-bounding-boxes-for-equation-tokens,locate-bounding-boxes-for-symbols-with-affixes,locate-composite-symbols,collect-symbol-locations,upload-symbols,fetch-spp-jsons,locate-sentences,upload-sentences,tokenize-sentences,create-annotation-files,detect-definitions,extract-contexts-for-definitions,locate-bounding-boxes-for-definitions,upload-definitions} [{fetch-arxiv-sources,fetch-s2-metadata,unpack-sources,compile-tex,normalize-tex,compile-normalized-tex,fetch-arxiv-pdf,raster-pages,detect-sentences,extract-bibitems,resolve-bibitems,locate-bounding-boxes-for-citation-fragments,locate-citations,upload-citations,detect-equations,locate-bounding-boxes-for-equations,upload-equations,detect-glossary-terms,extract-contexts-for-glossary-terms,locate-bounding-boxes-for-glossary-terms,upload-glossary-terms,extract-symbols,find-symbol-matches,extract-contexts-for-symbols,locate-bounding-boxes-for-equation-tokens,locate-bounding-boxes-for-symbols-with-affixes,locate-composite-symbols,collect-symbol-locations,upload-symbols,fetch-spp-jsons,locate-sentences,upload-sentences,tokenize-sentences,create-annotation-files,detect-definitions,extract-contexts-for-definitions,locate-bounding-boxes-for-definitions,upload-definitions} ...]]
                       [--start {fetch-arxiv-sources,fetch-s2-metadata,unpack-sources,compile-tex,normalize-tex,compile-normalized-tex,fetch-arxiv-pdf,raster-pages,detect-sentences,extract-bibitems,resolve-bibitems,locate-bounding-boxes-for-citation-fragments,locate-citations,upload-citations,detect-equations,locate-bounding-boxes-for-equations,upload-equations,detect-glossary-terms,extract-contexts-for-glossary-terms,locate-bounding-boxes-for-glossary-terms,upload-glossary-terms,extract-symbols,find-symbol-matches,extract-contexts-for-symbols,locate-bounding-boxes-for-equation-tokens,locate-bounding-boxes-for-symbols-with-affixes,locate-composite-symbols,collect-symbol-locations,upload-symbols,fetch-spp-jsons,locate-sentences,upload-sentences,tokenize-sentences,create-annotation-files,detect-definitions,extract-contexts-for-definitions,locate-bounding-boxes-for-definitions,upload-definitions}]
                       [--end {fetch-arxiv-sources,fetch-s2-metadata,unpack-sources,compile-tex,normalize-tex,compile-normalized-tex,fetch-arxiv-pdf,raster-pages,detect-sentences,extract-bibitems,resolve-bibitems,locate-bounding-boxes-for-citation-fragments,locate-citations,upload-citations,detect-equations,locate-bounding-boxes-for-equations,upload-equations,detect-glossary-terms,extract-contexts-for-glossary-terms,locate-bounding-boxes-for-glossary-terms,upload-glossary-terms,extract-symbols,find-symbol-matches,extract-contexts-for-symbols,locate-bounding-boxes-for-equation-tokens,locate-bounding-boxes-for-symbols-with-affixes,locate-composite-symbols,collect-symbol-locations,upload-symbols,fetch-spp-jsons,locate-sentences,upload-sentences,tokenize-sentences,create-annotation-files,detect-definitions,extract-contexts-for-definitions,locate-bounding-boxes-for-definitions,upload-definitions}]
                       [--arxiv-ids ARXIV_IDS [ARXIV_IDS ...]]
                       [--arxiv-ids-file ARXIV_IDS_FILE] [--days DAYS]
                       [--s3-job-spec S3_JOB_SPEC] [--source {arxiv,s3}]
                       [--s3-arxiv-sources-bucket S3_ARXIV_SOURCES_BUCKET]
                       [--entity-batch-size ENTITY_BATCH_SIZE]
                       [--keep-intermediate-files] [--max-papers MAX_PAPERS]
                       [--one-paper-at-a-time] [--keep-paper-data]
                       [--store-results] [--store-log]
                       [--s3-output-bucket S3_OUTPUT_BUCKET]
                       [--notify-emails NOTIFY_EMAILS [NOTIFY_EMAILS ...]]
                       [--notify-bcc NOTIFY_BCC [NOTIFY_BCC ...]]
                       [--extraction-only] [--dry-run]
                       [--database-schema DATABASE_SCHEMA]
                       [--database-create-tables]
                       [--data-version DATA_VERSION]
                       [--definitions-base-url DEFINITIONS_BASE_URL]
                       [--output-forms {db,file} [{db,file} ...]]
                       [--output-dir OUTPUT_DIR]
run_pipeline.py: error: Please include either both 'file' as an output form and an output dir, or neither.
```

data-processing/common/commands/database.py
ca16: Yes, thanks for the suggestion!

ca16:
@andrewhead thanks for the comments! They were helpful.

I've made a few updates based on your suggestions, and attempted to answer your questions.

I think the one still open question is around using `Literal` instead of `Enum`, from this [comment](https://github.com/allenai/scholarphi/pull/318#discussion_r711057407).

data-processing/common/upload_entities.py
ca16: (I ended up changing the key name from 'version' to 'format_version' to try to make this clearer.)

andrewhead: This sounds good!

data-processing/common/commands/database.py
andrewhead: The second implementation is just the one I was thinking of! Though I think you make a good point:

> Which seems worse than the current Enum option, because I've got "db" and "file" lots of places (which feels like it would be more prone to typos coming about).

When I have used mypy and pylance in VSCode, it has supported autocomplete of literal values, and statically checks in cases where the string provided for the type is not one that is listed in the Literal type, to help rule out typos. Though that said, I think I good argument in favor of sticking with enums is if this support doesn't seem to be consistently in tools, where I think we could rely on good typo-checking support for enums.

andrewhead:
hi @ca16 ! Awesome PR, I am already using these flags!!

Would it make sense to rename the extension of the output from `.jsonl` to `.json`, given that the file contains just one line and can parsed as single JSON object? Sorry, I should have thought to ask this before, though I only realized it when I started working with the file.

ca16:
Yeah I think that's a good idea! I can make the change.

PR #356, opened 2022-06-03
Egork/scholar 32314 citation matching

comorado:
Modifying similarity function for the citation matching


data-processing/entities/citations/types.py
```diff
@@ -26,3 +26,4 @@ class BibitemMatch:
     bibitem_text: str
     s2_id: str
     s2_title: str
+    similarity_score: float
```
ca16, line 29: Curious about why we're adding this here?

data-processing/tests/entities/citations/commands/test_resolve_bibitems.py
```diff
@@ -0,0 +1,39 @@
...
+            bibitems=[Bibitem(id_='one1999two', text='dog cat monkey', context_tex='N/A', start=-1, end=-1, tex_path='N/A',
+                     tex='N/A'),
+                     Bibitem(id_='two2020two', text='one two three', context_tex='N/A', start=-1, end=-1, tex_path='N/A',
+                     tex='N/A')],
+            references=[SerializableReference(s2_id='111', arxivId='2022.04708', doi=None, title='Dog on a walk',
+                                   authors="[{'id': 1, 'name': 'Charles Dickens'}, {'id':2, 'name': 'Ernest Hemingway'}]",
+                                   venue='arXiv', year='2022')]
+        )
+        result = next(resolve_bibitems.process(item))
+        assert result == BibitemMatch(key='one1999two', bibitem_text='dog cat monkey', s2_id='111',
```
ca16, line 37: This match seems like a stretch 😆 Though I think I get why it matches given the current code. 
Do you think the threshold we get from the evaluations will help filter out this kind of thing, or is the purpose of this test to have an example of a kind of mistake we know we're going to make?

data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -118,3 +130,45 @@ def save(self, item: MatchTask, result: BibitemMatch) -> None:
 
         resolutions_path = os.path.join(resolutions_dir, "resolutions.csv")
         file_utils.append_to_csv(resolutions_path, result)
+
+    @staticmethod
+    def similarity_count_vectorizer(s2_return: str, bibliography_reference: str) -> float:
+        """
+        Similarity measure based on CountVectorizer
+        https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text
```
ca16, line 138: Thanks for including this link!

ca16 (commented):
Thanks for looking into this!

I've left a few questions and suggestions in line.

More generally:
1) I think you said an updated threshold value will be coming from evaluations. This seems important as without it, if I'm reading stuff right, it seems like having just one token in common would be sufficient to make a match.
2) My understanding is that once we have the threshold, we will be ready to put this into production. Is that correct?

data-processing/tests/entities/citations/commands/test_resolve_bibitems.py
comorado: Once we pick the threshold (I believe it is going to be large than 0.02 which is the similarity score) the current test will fail. I will add more test cases which are more descriptive and have higher value of the similarity.

data-processing/entities/citations/types.py
comorado: It will help us in future debugging. Without we have no way of getting the score in case we need to look at the similarity match score.

data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -118,3 +130,45 @@ def save(self, item: MatchTask, result: BibitemMatch) -> None:
...
+            return 0.0
+
+        vectorizer = CountVectorizer(binary=True, ngram_range=(1, 5), stop_words='english')
+        try:
+            vectorizer.fit([s2_return])
+        except ValueError as e:
+            print(e)
+            return 0.0
+
+        if len(vectorizer.get_feature_names()) > 3:
```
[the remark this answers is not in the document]
comorado: Yes that is right I have a section in the [doc](https://docs.google.com/document/d/1qbu3_3WOTCyPqeFPh_znHY8lVhxE5zygJNQeNjExLbU/edit#bookmark=id.xs6bjya1vntv) on that. If the vocabulary size is small < 3 similarity score is quite high and has only 2 value 1 and ~0.4. We want to avoid such S2 entries as probably we will make mistakes matching to them.

comorado:
To close the loop on the threshold I have a new section in the [doc](https://docs.google.com/document/d/1qbu3_3WOTCyPqeFPh_znHY8lVhxE5zygJNQeNjExLbU/edit#bookmark=id.lpx8qg2j8cyt) we decided to use score_threshold = 0.2
Here is the plot of the score vs match rate and score vs coverage of bib entries. Full lines correspond to match rate vs score, dashed lines correspond to coverage vs score. Three set of results are presented:
- 3 gram approach (current approach used in scholarPhi), score threshold value = 0.5 which corresponds to ~88% of match rate and 7% of items not assigned to s2_id. Results are plotted in magenta color
- countVectroizer approach which uses extracted text from the bib entry plotted in green at score = .2 with match rate 98% and 25% of items not getting s2 id assigned
- countVectroizer approach which uses full bib entry plotted in yellow at score = 0.2 with match rate of 98.4% and 19% of items not getting s2 id assigned

image

kelseym-ai2:
Much improved AUC! Given that we have a mechanism to fallback to unmatched reference strings in raw form, I think that this very high precision and lower recall is a better user experience. We're only showing users the data that we're confident in, instead of having 12% erroneous cards.

kelseym-ai2 (approved):

ca16 (approved):
Totally agree. Would much rather have a few more unmatched entries, but this lower match error rate. I'm excited to run the papers that we know have match errors through!

Wrt this PR, there are a few comments left that would ideally get addressed before this merges, though they aren't deal breakers.

Next steps: once this merges, my understanding is that this is ready to go to prod. If so, I'll update the downstream systems to run this instead. Probably after the tech debt sprint.

ca16:
Thanks for the changes!

PR #357, opened 2022-06-07
Egork/scholar 32314 citation matching full bib entry

comorado:


data-processing/entities/citations/types.py
```diff
@@ -19,6 +19,9 @@ class Bibitem(SerializableEntity):
     text: str
     " Plaintext extracted for bibitem. "
 
+    bibitem_code: str
+    ' Bibitem code'
```
ca16, line 23: Tiny thing: is there a meaning behind the description of the field above being enclosed in double quotes, and this description being enclosed in single quotes? If not, can we have them be consistent with each other (otherwise I think it'll make people think they're different for a reason).

data-processing/entities/citations/commands/resolve_bibitems.py
```diff
@@ -99,6 +99,7 @@ def process(self, item: MatchTask) -> Iterator[BibitemMatch]:
                     most_similar_reference.s2_id,
                     most_similar_reference.title,
                     max_similarity,
+                    bibitem.bibitem_code,
```
ca16, line 102: For debug purposes also?

ca16 (approved):
One suggestion and a question. 

Apart from that, assuming the tests still pass, and the evaluations point to this option being best, LGTM.

data-processing/entities/citations/types.py
ca16: (TBH  both of these actually confuse me - I would have expected a comment, but I trust that you and the writer of the other one know what you're doing... Either way though, I think consistency between the two would be great unless there's a reason they're different.)

PR #377, opened 2023-02-01
Save Corpus ids from PublicApi references matched to PDF Bib Entries

stupendousC:
* From 4th checkbox in parent task [[pipe corpus ids through scholarphi to s2airs] dev/local end to end](https://github.com/allenai/scholar/issues/35326#top) #35326
  * Parent task has 3 PRs in these 3 repos: `scholarphi` --> `scholarphi-pipeline` --> `s2airs`
  * This PR is for that first repo.
    * Where scholarphi takes the bib entries from a PDF, and the list of references from Public API. 
    * Title-matching done on these 2 groups so we know which bib entry corresp to which paper in our corpus.
    * The new thing here is that we also want to save the corpus IDs.

FILES I TOUCHED
-----------
* The files I touch in this PR came from studying the[ flow in Scholarphi](https://github.com/allenai/scholar/issues/35326#issuecomment-1412952683).  
  * The most important thing is the updated PublicAPI endpoint that provides me with the corpus ids to begin with.
  * [Example](https://api.semanticscholar.org/graph/v1/paper/arXiv:1512.02595?fields=references.authors,references.title,references.externalIds,references.venue,references.year) of what this looks like: 
  * Screen Shot 2023-02-03 at 12 16 53 PM
* Back when I erroneously submited this PR to the `main` branch, I had a bunch of comments showing in the 1st commit "initial thoughts" for what goes where, but now that I've switched base to `chi-2021-demo`, they're gone.   Oh well, no big deal.
* Recommended viewing order (following the flow mentioned above):
  * `fetch_s2_data.py` & its `test`
    * Public Api now returns references' corpus IDs, which is saved in `references.csv` on L152
  * `common/types.py`
    * Data classes now store corpus IDs.
    * Also added new string types to make `CitationData` more understandable.
  * `resolve_bibitems.py` & its `test`
    * Now a bib item is matched to an S2 corpus paper, I want to save that corpus ID.
  * `entities/citations/types.py`
  * `upload_citations.py`
    * Didn't realize the final "citations.json" is the goal, instead of the intermediate various CSVs.  Anyway, CP ids should be in there now.
  * `process_pdf.py`
    * Not sure why this is still there, it's not being used anymore, esp since the import on L14 refers to something that was [deleted in July 2020](https://github.com/allenai/scholarphi/commits/main/data-processing/entities/citations/utils.py).
    * Added TODO to consider deleting it later.

TESTING 👍 && 👍 
-------
* **BOTH approaches below need to be green/thumbsUp before merging!!**
* IDEA 1 = [E2E testing](https://github.com/allenai/scholar/issues/35326#issuecomment-1431998184) of scholarphi --> scholarphi-pipeline --> s2airs (as seen in [reference PR](https://github.com/allenai/scholarphi/pull/362))  👍 
  * **DEFINITELY** will do this, but need to change stuff in the other 2 repos to prove this.  
  * Make sure the entire assembly line works before I merge any PRs for any of the 3 repos.
* IDEA 2 = Test just this scholarphi portion 👍 
  * Will involve the scholarphi/data-processing container on my machine
    * `docker pull <latest official image>` ([image](https://github.com/allenai/scholarphi-pipeline/blob/main/scholarphi_pipeline/processors/Dockerfile#L1))
    * from data-processing/... `docker run --rm -it -v $(pwd):/data-processing 60bb4741 /bin/bash`
      * mount my file changes to override behavior
  * IDEA 2A = `pytest --all` as seen in  ([readme](https://github.com/allenai/scholarphi/blob/main/data-processing/README.md)) <-- 👎  but it's out of my hands and nothing to do with me
    * Failed 7 tests (even without my changes), I saw code 404/NotFound.
      * I think the [S3 bucket it mentioned ](https://github.com/allenai/scholarphi/blob/chi-2021-demo/data-processing/README.md#a-note-on-tests-around-definitions) doesn't even exist anymore. 
      * Seems like these [3 files](https://github.com/allenai/scholarphi/blob/chi-2021-demo/data-processing/entities/definitions/nlp.py#L53-L67) are what the tests want.
      * So I went digging in S3 and see what buckets we have, I think it's either [`scholarphi-public` ](https://s3.console.aws.amazon.com/s3/buckets/scholarphi-public?region=us-west-2&prefix=model/v4/&showversions=false) or [`scholarphi4nlp`](https://s3.console.aws.amazon.com/s3/buckets/scholarphi4nlp?prefix=model/v4/&region=us-west-2)
      * So I went with scholarphi-public S3, and I subbed it in the code, cleaned `rm -rf ./cache/AI2020_model`, and then `pytest --all`
        * Now I see 7x code 403/Forbidden... but its' a public bucket 😖 
        * My PR doesn't touch the Definitions stuff anyway, so it's unlikely I broke it.  Still... it'd be nice to see it pass.
        * TRY: maybe do the scholarphi4nlp bucket url instead.  Will re-attempt if I feel like it later.
  * IDEA 2B = `pytest` is just a quick test, that doesn't deal with the Definitions stuff seen above.  It totally passed.  👍 
  * IDEA 2C = Run pipeline for just 1 paper 👍 
    * Also seen in readme, but don't follow those directions. 
    *  Instead I used the following, with arxiv_id=1512.02595...
      ```
      # from inside the docker container:
      python scripts/run_pipeline.py \
         --one-paper-at-a-time \
         --arxiv-ids 1512.02595 \
         --output-forms file \
         --output-dir data \
         --entities citations
      ```
    * Full log printout here for fun:
      ```
      root@6cf004dd1cc2:/data-processing# python scripts/run_pipeline.py    --one-paper-at-a-time    --arxiv-ids 1512.02595    --output-forms file    --output-dir data --entities citations
      wandb: WARNING W&B installed but not logged in.  Run `wandb login` or set the WANDB_API_KEY env variable.
      2023-02-07 03:28:21,145 [INFO]: The following commands will be run, in this order: ['fetch-arxiv-sources', 'fetch-s2-metadata', 'unpack-sources', 'compile-tex', 'normalize-tex', 'compile-normalized-tex', 'raster-pages', 'extract-bibitems', 'resolve-bibitems', 'locate-bounding-boxes-for-citation-fragments', 'locate-citations', 'upload-citations']
      2023-02-07 03:28:21,146 [INFO]: Running pipeline for paper 1512.02595
      2023-02-07 03:28:21,148 [INFO]: Launching command fetch-arxiv-sources
      2023-02-07 03:28:32,695 [INFO]: Finished running command fetch-arxiv-sources
      2023-02-07 03:28:32,698 [INFO]: Launching command fetch-s2-metadata
      2023-02-07 03:28:32,699 [INFO]: Issuing request to S2 @ api.semanticscholar.org
      2023-02-07 03:28:36,295 [INFO]: Finished running command fetch-s2-metadata
      2023-02-07 03:28:36,298 [INFO]: Launching command unpack-sources
      2023-02-07 03:28:36,423 [INFO]: Finished running command unpack-sources
      2023-02-07 03:28:36,424 [INFO]: Launching command compile-tex
      2023-02-07 03:28:39,286 [INFO]: Finished running command compile-tex
      2023-02-07 03:28:39,288 [INFO]: Launching command normalize-tex
      2023-02-07 03:28:39,410 [INFO]: Finished running command normalize-tex
      2023-02-07 03:28:39,412 [INFO]: Launching command compile-normalized-tex
      2023-02-07 03:28:43,432 [INFO]: Finished running command compile-normalized-tex
      2023-02-07 03:28:43,433 [INFO]: Launching command raster-pages
      2023-02-07 03:28:44,960 [INFO]: Finished running command raster-pages
      2023-02-07 03:28:44,962 [INFO]: Launching command extract-bibitems
      2023-02-07 03:28:45,930 [INFO]: Finished running command extract-bibitems
      2023-02-07 03:28:45,932 [INFO]: Launching command resolve-bibitems
      /usr/local/lib/python3.7/dist-packages/sklearn/utils/deprecation.py:87: FutureWarning: Function get_feature_names is deprecated; get_feature_names is deprecated in 1.0 and will be removed in 1.2. Please use get_feature_names_out instead.
        warnings.warn(msg, category=FutureWarning)
      2023-02-07 03:28:46,219 [WARNING]: Could not find a sufficiently similar reference for bibitem ModernGPU of paper 1512.02595
      2023-02-07 03:28:47,044 [WARNING]: Could not find a sufficiently similar reference for bibitem Elsen2015 of paper 1512.02595
      2023-02-07 03:28:48,347 [WARNING]: Could not find a sufficiently similar reference for bibitem NervanaGPU of paper 1512.02595
      2023-02-07 03:28:49,428 [WARNING]: Could not find a sufficiently similar reference for bibitem szegedy2014googlenet of paper 1512.02595
      2023-02-07 03:28:49,754 [INFO]: Paper has 70 references, able to match 66 reference(s) to S2 data.
      2023-02-07 03:28:49,755 [INFO]: Finished running command resolve-bibitems
      2023-02-07 03:28:49,757 [INFO]: Launching command locate-bounding-boxes-for-citation-fragments
      2023-02-07 03:29:39,475 [INFO]: Finished running command locate-bounding-boxes-for-citation-fragments
      2023-02-07 03:29:39,476 [INFO]: Launching command locate-citations
      2023-02-07 03:29:39,706 [INFO]: Finished running command locate-citations
      2023-02-07 03:29:39,708 [INFO]: We will be writing output to files.
      2023-02-07 03:29:39,708 [INFO]: Launching command upload-citations
      2023-02-07 03:29:39,728 [WARNING]: Missing S2 match for bibitem with key ModernGPU for paper 1512.02595
      2023-02-07 03:29:39,729 [WARNING]: Missing S2 match for bibitem with key Elsen2015 for paper 1512.02595
      2023-02-07 03:29:39,730 [WARNING]: Missing S2 match for bibitem with key NervanaGPU for paper 1512.02595
      2023-02-07 03:29:39,731 [WARNING]: Missing S2 match for bibitem with key szegedy2014googlenet for paper 1512.02595
      2023-02-07 03:29:39,731 [WARNING]: Missing S2 match for bibitem with key Elsen2015 for paper 1512.02595
      2023-02-07 03:29:39,732 [WARNING]: Missing S2 match for bibitem with key szegedy2014googlenet for paper 1512.02595
      2023-02-07 03:29:39,732 [WARNING]: Missing S2 match for bibitem with key NervanaGPU for paper 1512.02595
      2023-02-07 03:29:39,733 [WARNING]: Missing S2 match for bibitem with key ModernGPU for paper 1512.02595
      2023-02-07 03:29:39,733 [INFO]: Saving to file...
      2023-02-07 03:29:39,734 [INFO]: About to write 184 entity infos to data/citations.json (version: v0).
      2023-02-07 03:29:39,755 [INFO]: Finished running command upload-citations
      ```
      * Log printouts make sense with each other, 4 of those bib entries did not get matched to a reference seen in Public API.
    * OMG YES!!!!!!!
      * Screen Shot 2023-02-03 at 12 27 48 PM


data-processing/pdf/process_pdf.py
```diff
@@ -207,17 +213,16 @@ def get_citations(self) -> CitationData:
...
         upload_citations(citations, 'pdf-pipeline')
         symbols = self.get_symbols()
-        upload_symbols(symbols,'pdf-pipeline')
+        upload_symbols(symbols, "pdf-pipeline")
 
 if __name__ == '__main__':
     logging.basicConfig(level=logging.INFO)
 
-    pdf_hashes = ['8c53cd64e46cf382bf81ce2c4e0087ef31351919']
+    pdf_hashes = []
```
stupendousC, line 225: Rodney left that in by accident.  Cleaning up now.

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -102,10 +105,20 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
                     authors = []
                     for author_data in reference_data["authors"]:
                         authors.append(Author(author_data["authorId"], author_data["name"]))
+
+                    external_ids = reference_data["externalIds"]
+                    if external_ids:
+                        corpus_id = external_ids.get("CorpusId")
+                        arxiv_id = external_ids.get("ArXiv")
+                        doi = external_ids.get("DOI")
```
stupendousC, line 113: the shape of data is different in the legacy PublicAPI endpoint vs this new one now, see the screenshot in the main comment.

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -102,10 +105,20 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
...
+                    else:
+                        corpus_id = arxiv_id = doi = None
+
                     reference = Reference(
                         s2_id=reference_data["paperId"],
-                        arxivId=reference_data["arxivId"],
-                        doi=reference_data["doi"],
+                        corpus_id=corpus_id,
```
ca16, line 119: Small suggestion: maybe let's put corpus_id at the end, in case there's anything that relies on the order that these fields appear in? Having said that, if you're confident you know all the usages, and there's nothing that does rely on the order that would get messed up, it's probably find to do it this way...

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -102,10 +105,20 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
...
+                    if external_ids:
+                        corpus_id = external_ids.get("CorpusId")
+                        arxiv_id = external_ids.get("ArXiv")
+                        doi = external_ids.get("DOI")
+                    else:
+                        corpus_id = arxiv_id = doi = None
+
                     reference = Reference(
                         s2_id=reference_data["paperId"],
-                        arxivId=reference_data["arxivId"],
```
ca16, line 107: Making sure I follow... This would also result in an optional arXiv id and optional doi, because in the old endpoint's response shape, when we don't have these for a reference, the key still exists, and the value is `null`. Is that right?

data-processing/common/types.py
```diff
@@ -92,6 +94,7 @@ class Author:
 @dataclass(frozen=True)
 class Reference:
     s2_id: S2Id
+    corpus_id: S2CorpusId
```
ca16, line 97: Wondering why the definition of this type isn't an optional string... Actually wondering the same thing about doi and arxiv_id...

data-processing/entities/citations/commands/upload_citations.py
```diff
@@ -104,30 +104,31 @@ def _entity_data_from_text_and_matches(
...
-                arxiv_id,
-            )
+            log_missing("bibitem text")
 
         if bibitem_key in key_s2_ids:
-            entity_data["paper_id"] = key_s2_ids[bibitem_key]
+            reference_s2_id = key_s2_ids[bibitem_key]
+            entity_data["paper_id"] = reference_s2_id
+            if reference_s2_id in s2_data:
+                entity_data["corpus_id"] = s2_data[reference_s2_id].corpus_id
```
ca16, line 127: So s2_data is reference info keyed by the legacy paper id of the reference? So s2_data only contains info about matched references?

data-processing/pdf/process_pdf.py
```diff
@@ -1,243 +0,0 @@
-import logging
```
ca16, line 1: I do get the reasoning behind deleting this... I would suggest though maybe not including it in this PR to keep this PR smaller and to make it easier to revert the deletion if it turns out we want this file for whatever reason. Especially because we're so rarely in this code, it feels a little scarier to me to pull in unrelated maintenance stuff. But I leave it up to you!

ca16 (commented):
Thanks for doing this! 

I left mostly questions, and some initial feedback. Like we kinda talked about (and will probably talk about more later), I would recommend doing some E2E testing before merging, now that it looks like you've got the scholarphi side mostly figured out...

data-processing/common/commands/fetch_s2_data.py
stupendousC: Yup.

With [OLD API]((https://api.semanticscholar.org/v1/paper/arXiv:1512.02595))
-------
We can see references that has a null ArXiv and/or null DOI.  Here's one with both being null:
Screen Shot 2023-02-06 at 1 44 00 PM

So the old code was making `Reference` obj like this: 
``` 
reference = Reference(
    s2_id=reference_data["paperId"], <-- actual value
    arxivId=reference_data["arxivId"], <-- None
    doi=reference_data["doi"], <-- None
    ... )
```


With [NEW API](https://api.semanticscholar.org/graph/v1/paper/arXiv:1512.02595?fields=references.authors,references.title,references.externalIds,references.venue,references.year)
-------
That same reference item shows up as:
Screen Shot 2023-02-06 at 1 50 37 PM

So the new code was making `Reference` obj the same way:
```
reference = Reference(
    s2_id=reference_data["paperId"],  <-- actual value
    corpus_id=corpus_id, <-- None
    arxivId=arxiv_id, <-- None
    doi=doi, <-- None
    ...)
```

And for a reference item that has null under `externalIds`, such as this:
Screen Shot 2023-02-06 at 1 55 07 PM
... we'd still be making the `Reference` obj the same way above, except this time s2_id is also None.

data-processing/common/commands/fetch_s2_data.py
stupendousC: oh yeah, i only put that in that order so I can keep `corpus_id` close to `s2_id` so it's easier on the eyeballs.  I think it'll be fine to leave as is though.

data-processing/common/types.py
stupendousC: hmm yeah IDK... on L90 for `Author` they actually did make the authorId an `Optional`...

Does it not matter? confused...

data-processing/entities/citations/commands/upload_citations.py
stupendousC: * `ResolveBibItems` make matches by comparing title strings of the PDF bib entry VS the reference titles given by Public API.  
  * INTERESTING!!!  The old API **ONLY** provides references that has a paper_id/SHA.  
    * but **NOT** the new API!
    * So given that not all the reference items from Public API actually have paper_id/SHA or any other ids,
      * ...it's entirely possible for the `ResolveBibItems` step to make a `BibitemMatch` for a bib entry to such a paper!
      * But we shouldn't even allow such a match to occur in the first place bc there's no point!  
      * OK, I'm gonna need to change soemthing in `ResolveBibItems` to exclude from matching on these deadbeat references from PublicAPI!!!
  * These `BibitemMatch` results are saved to the intermediate `resolution.csv`
* So this next step `UploadCitations` grabs that `resolutions.csv`
  * L125 could actually be None, oh no!  (Hence the fix to exclude garbage above!)
  
STAY TUNED!!! still thinking/tinkering...

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -99,13 +102,28 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
                 data = resp.json()
                 references = []
                 for reference_data in data["references"]:
+
+                    if reference_data["paperId"] is None:
+                        # can be legit references not in our corpus, or bad references from bogus extractions
+                        continue
```
stupendousC, line 108: * As mentioned in this [comment](https://github.com/allenai/scholarphi/pull/377/files#r1097999664)...
  * New public API endpoint gives us _all_ the references, even for those that failed Citation Matching (and that's why both paperID/SHA and corpus ID are blank).  
  * Old API already ignore those, so here we're making sure the behavior is not changed.
* PEOPLE MIGHT ASK... ok so the new API does provide legit references which just happen to not be in our corpus, wouldn't it be good to include those anyway?
  * No... currently this group of id-less failed citation matches ALSO includes a lot of garbage from bogus extractions.  We can't tell (without human intervention) which is which.  So right now it's just safer to exclude this noise from getting passed into the Reader.  
  * Besides, the inline mentions should (irrelev to this ticket) already show Citation Cards with general error msgs.  
  * Even if we do pass in these id-less papers, it's not like there's a valid PDP to send users to anyway.  YAGNI.

data-processing/common/commands/fetch_s2_data.py
stupendousC: It is entirely possible to get `arxiv_id` and/or `doi` to be `None`, just like before.

data-processing/scripts/process_pdf.py
```diff
@@ -1,3 +1,6 @@
+# TODO: CONSIDER REMOVING ENTIRE FILE, doesn't look like it's even being used anymore.
+#  the import upload_citations from entities.citations.util has been deleted in July 2020.
```
stupendousC, line 2: THIS IS THE ONLY CHANGE I made in this file.

ca16: Looks like it maybe also got renamed?

data-processing/scripts/process_pdf.py
```diff
@@ -64,10 +67,10 @@ def union(self, bbox1: BoundingBoxModel, bbox2: BoundingBoxModel):
     @staticmethod
     def should_combine(bbox1: BoundingBoxModel, bbox2: BoundingBoxModel):
         return bbox1.page == bbox2.page and (
-            abs(bbox1.top - bbox2.top) < 4 # Same y-coordinate
+                abs(bbox1.top - bbox2.top) < 4 # Same y-coordinate
```
ca16, line 70: Plus some spacing changes?

data-processing/scripts/process_pdf.py
stupendousC: that's what i thought at first, but the existing `load_located_citations()` in `entities.citations.util.py` has nothing to do with this.  Looks like there was a whole bunch of refactoring going on for that [commit](https://github.com/allenai/scholarphi/commit/a4f8285d6f443f2c3b97dc12f9098794c89d79f3) over 17 files anyway.

data-processing/scripts/process_pdf.py
stupendousC: ooops, that happened somehow from copy/paste.

data-processing/common/commands/fetch_s2_data.py
stupendousC: Ok, so I reran the testing (as seen in IDEA 2C from this PR's main comment), and the resulting `citations.csv` is still the same, as expected.

The only difference is that I can see from my temporary print statements (green) that those 3 unwanted no-id references from the NEW public API endpoint have been excluded, yay.
Screen Shot 2023-02-06 at 5 33 36 PM

stupendousC:
PARKING this as a DRAFT
----------
Just for now.  So I don't accidentally try to merge this.   

Will flip back to review status once the E2E testing across scholarphi --> scholarphi-pipelines --> s2airs is done.

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -99,13 +102,28 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
                 data = resp.json()
                 references = []
                 for reference_data in data["references"]:
+
+                    if reference_data["paperId"] is None:
```
ca16, line 106: Mostly a curiosity thing - why have this case instead of having just `if reference_data["paperId"]: do stuff`?

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -99,13 +102,28 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
...
+                        # can be legit references not in our corpus, or bad references from bogus extractions
+                        continue
+
+                    external_ids = reference_data["externalIds"]
+                    if external_ids:
+                        corpus_id = external_ids.get("CorpusId")  # always present (scholar/citations/MatchResult.scala)
```
ca16, line 112: Looking back at this comment... I wonder if it might have more potential to confuse someone than to help, because of all the intermediate steps between citation matching and what the public API returns, as well as the fact that this is an open source library so not everyone will have access to the code you're pointing at... What do you think? Maybe we could just say every canonical paper has a corpus id, so you shouldn't be able to have a matched reference without one?

data-processing/common/types.py
ca16: Hmmm that seems potentially more confusing than not having types at all. But also not your PR's problem.

data-processing/tests/common/commands/test_fetch_s2_data.py
```diff
@@ -14,6 +14,9 @@
 )
 from scripts.commands import run_command
 
+references_fields = ["authors", "title", "externalIds", "venue", "year"]
```
ca16, line 17: What do you think of maybe calling these `expected_references_fields` and `expected_fields_query`?

data-processing/common/commands/fetch_s2_data.py
stupendousC: oh lol, yeah it's just my old habit of throwing out unwanted stuff in the beginning, then everything afterwards can just be the big chunk of code where I work on stuff I actually want.

```
if (unwanted condition):
  skip this one

# huge chunk of code specifically for ones I want
# do this
# do that
# etc...
```

Whereas I suppose this would be cleaner:
```
if (wanted condition):
  # do this
  # do that
  # etc...
```

Will change!!

data-processing/common/commands/fetch_s2_data.py
stupendousC: oh yeah!  lol I forgot scholarphi is open source.  Yes, I will change the comment to `# every canonical paper will always have a CorpusId`

ca16:
A more general question - just want to double check that for `Reference`, `SerializableReference` and `BibItemMatch`, to which this PR adds new fields, all their usages have been accounted for?

ca16 (approved):

PR #381, opened 2023-03-28
Switch to new public API endpoint, and deprecate partner subdomain

stupendousC:
* Parent task = [Switch scholarphi to new public API endpoint #32158](https://github.com/allenai/scholar/issues/32158#top)
  * As seen in [comment/question](https://github.com/allenai/scholar/issues/32158#issuecomment-1456799584), we have 2 goals:
    * GOAL 1 = use the latest `graph/v1/` endpoint instead of legacy `v1/` endpoint
    * GOAL 2 = use the regular `api.semanticscholar.org` subdomain, instead of an optional `partner.semanticscholar.org` subdomain.
      * Question: before, users with `S2_PARTNER_API_TOKEN` env were calling the partner domain for higher rate limit.  Can we use the same token on the regular `api.semantic.org` subdomain?   
        * [Rodney says yes](https://github.com/allenai/scholar/issues/32158#issuecomment-1487492642).

* THIS PR is for the `main` branch.
  * Goal 1 was done in the `chi-2021-demo` branch as seen in this PR: #377 
    * So we'll be using that as a reference, MINUS the changes made for adding corpus IDs of citations. 
  * Goal 2 was done in the `chi-2021-demo` branch as seen in this PR: #382 
    * So we used that as a reference as well.

### DIFF
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -118,13 +118,9 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
     ),
   };
   if (apiKey) {
-    conf.headers['x-api-key'] = apiKey;
+    conf.headers['x-api-key'] = apiKey; // TODO: should I keep this around?  Asking Rodney https://github.com/allenai/scholar/issues/32158#issuecomment-1487492642
   }
-  const apiOrigin = (
-    apiKey
-      ? "https://partner.semanticscholar.org/v1"
-      : "https://api.semanticscholar.org/v1"
-  );
+  const apiOrigin = "https://api.semanticscholar.org/graph/v1";
 
   return axios.get<Paper>(`${apiOrigin}/paper/${s2Id}`, conf);
 }
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -118,7 +118,7 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
     ),
   };
   if (apiKey) {
-    conf.headers['x-api-key'] = apiKey; // TODO: should I keep this around?  Asking Rodney https://github.com/allenai/scholar/issues/32158#issuecomment-1487492642
+    conf.headers['x-api-key'] = apiKey;
   }
   const apiOrigin = "https://api.semanticscholar.org/graph/v1";
 
--- a/data-processing/common/commands/fetch_s2_data.py
+++ b/data-processing/common/commands/fetch_s2_data.py
@@ -48,7 +48,7 @@ class S2ReferencesNotFoundException(S2MetadataException):
 class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
     def __init__(self, args: Any) -> None:
         super().__init__(args)
-        # self._partner_api_token = os.getenv("S2_PARTNER_API_TOKEN", None) # TODO: can this be used on regular api.ss.org subdomain?
+        self._partner_api_token = os.getenv("S2_PARTNER_API_TOKEN", None)
 
     def _mk_api_request(self, arxivId: ArxivId) -> requests.Response:
         # XXX(andrewhead): S2 API does not have versions of arXiv papers. I don't think this
@@ -58,8 +58,8 @@ class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
         base_url = "api.semanticscholar.org"
         headers = None
 
-        # if self._partner_api_token:   # TODO: idk, might be useful to keep for rate limits?
-        #     headers = {"x-api-key": self._partner_api_token}
+        if self._partner_api_token:
+            headers = {"x-api-key": self._partner_api_token}
 
         references_fields = ["authors", "title", "externalIds", "venue", "year"]
         fields_query = "fields=references." + ",references.".join(references_fields)
@@ -102,27 +102,28 @@ class FetchS2Metadata(ArxivBatchCommand[ArxivId, S2Metadata]):
                 references = []
                 for reference_data in data["references"]:
 
-                    arxiv_id = doi = None
                     if reference_data["paperId"]:
                         external_ids = reference_data["externalIds"]
                         if external_ids:
                             arxiv_id = external_ids.get("ArXiv")    # may be None
                             doi = external_ids.get("DOI")           # may be None
+                        else:
+                            arxiv_id = doi = None
 
-                    authors = []
-                    for author_data in reference_data["authors"]:
-                        authors.append(Author(author_data["authorId"], author_data["name"]))
+                        authors = []
+                        for author_data in reference_data["authors"]:
+                            authors.append(Author(author_data["authorId"], author_data["name"]))
 
-                    reference = Reference(
-                        s2_id=reference_data["paperId"],
-                        arxivId=arxiv_id,
-                        doi=doi,
-                        title=reference_data["title"],
-                        authors=authors,
-                        venue=reference_data["venue"],
-                        year=reference_data["year"],
-                    )
-                    references.append(reference)
+                        reference = Reference(
+                            s2_id=reference_data["paperId"],
+                            arxivId=arxiv_id,
+                            doi=doi,
+                            title=reference_data["title"],
+                            authors=authors,
+                            venue=reference_data["venue"],
+                            year=reference_data["year"],
+                        )
+                        references.append(reference)
 
                 if not references:
                     # References are required to process citations, mark job as failed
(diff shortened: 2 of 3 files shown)


api/src/s2-api.ts
```diff
@@ -120,11 +120,7 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = (
-    apiKey
-      ? "https://partner.semanticscholar.org/v1"
-      : "https://api.semanticscholar.org/v1"
-  );
+  const apiOrigin = "https://api.semanticscholar.org/graph/v1";
```
codeviking, line 123: Is the API key that's in place now specific to the partner site? Do we need to provision a new key for ScholarPhi to use with the new graph API?

stupendousC: nope!  [Checked with Rodney and he says the same partner API key can be used for either subdomains.](https://github.com/allenai/scholar/issues/32158#issuecomment-1487527699)

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -99,19 +101,29 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
...
-                        arxivId=reference_data["arxivId"],
-                        doi=reference_data["doi"],
-                        title=reference_data["title"],
-                        authors=authors,
-                        venue=reference_data["venue"],
-                        year=reference_data["year"],
-                    )
-                    references.append(reference)
+
+                    if reference_data["paperId"]:
```
kyleclo, line 105: what happens in case missing `paperId`?

kyleclo: can we define a new error type complementing `S2ReferencesNotFoundException` that handles case when `if reference_data["paperId"]` is `False`?

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -59,13 +59,15 @@ def _mk_api_request(self, arxivId: ArxivId) -> requests.Response:
         headers = None
 
         if self._partner_api_token:
-            base_url = "partner.semanticscholar.org"
             headers = {"x-api-key": self._partner_api_token}
 
+        references_fields = ["authors", "title", "externalIds", "venue", "year"]
+        fields_query = "fields=references." + ",references.".join(references_fields)
```
codeviking, line 65: This took me a second to understand, mainly because of the leading `fields=references+` bit. What do you think about this?

```python
fields = ["references." + f for f in [ "authors", "title", "externalIds", "venue", "year" ] ]
```

Also I'm omitting the query string argument name. More on that below.

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -59,13 +59,15 @@ def _mk_api_request(self, arxivId: ArxivId) -> requests.Response:
...
+
         logger.info(f"Issuing request to S2 @ {base_url}")
 
         return requests.get(
-            f"https://{base_url}/v1/paper/arXiv:{versionless_id}",
+            f"https://{base_url}/graph/v1/paper/arXiv:{versionless_id}?{fields_query}",
```
codeviking, line 70: I'd suggest using the `params` parameter instead of building the query string yourself. That'll take care of escaping things that need to be escaped, and probably eliminates a few lines. There's probably nothing to escape here, but it's one less thing to think about.

Combined with the change above I think that'd produce something like this:

```python
return requests.get(
    "https://{base_url}/graph/v1/paper/arXiv:{versionless_id}",
    params={"fields": ",".join(fields) },
    headers=headers
)
```

Here's the relevant docs for the `requests` library:
https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

data-processing/common/commands/fetch_s2_data.py
codeviking, line 105: I don't see this check in the existing code. Is there a reason we're doing it now?

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -99,19 +101,29 @@ def process(self, item: ArxivId) -> Iterator[S2Metadata]:
...
-                    )
-                    references.append(reference)
+
+                    if reference_data["paperId"]:
+                        external_ids = reference_data["externalIds"]
+                        if external_ids:
+                            arxiv_id = external_ids.get("ArXiv")    # may be None
+                            doi = external_ids.get("DOI")           # may be None
+                        else:
+                            arxiv_id = doi = None
```
codeviking, line 111: I think this should have the same effect:

```suggestion
                        arxiv_id = reference_data.get("externalIds", {}).get("ArXiv")
                        doi = reference_data.get("externalIds", {}).get("DOI")
```

data-processing/tests/common/commands/test_fetch_s2_data.py
```diff
@@ -46,7 +50,7 @@ def test_makes_request_over_partner_api_when_token_present():
             command = FetchS2Metadata(Args(arxiv_ids=['fakeid']))
             command._mk_api_request("fakeid")
             mock_requests.get.assert_called_with(
-                "https://partner.semanticscholar.org/v1/paper/arXiv:fakeid",
+                f"https://api.semanticscholar.org/graph/v1/paper/arXiv:fakeid?{expected_fields_query}",
                 headers={"x-api-key": "some_token"}
             )
```
codeviking, line 55: I'd make the same changes noted above here too, by using the `requests` lib for constructing query string arguments.

Also I'd just copy the line for producing the list of reference related fields and inline it with each invocation here. Sure, you could use a single variable that's hosted to the top of the file -- but that's more indirection. Keeping things inline makes the code a bit easier to read, as there's no need to jump around.

codeviking (approved):
I left a few ideas that might improve the code a bit, but there's nothing functionally wrong that I can see.

Once you've merged this TAL at https://marina.apps.allenai.org/a/s2-reader to see the resulting deploys.

data-processing/common/commands/fetch_s2_data.py
stupendousC: Kyle asked the same question, here's the [response](https://github.com/allenai/scholarphi/pull/381#discussion_r1151235345)

stupendousC: > can we define a new error type complementing S2ReferencesNotFoundException that handles case when if reference_data["paperId"] is False?

I'm not sure if that's going to be useful?  Esp since it's very possible that the unmatched reference turns out to be bogus.  For example, I've seen "references" that were obviously from bad extractions, where the "title" is the caption from a graph, or where the "title" is essentially a repeat of another legit reference (this happens bc different Sourced Papers can have slightly differently worded reference titles, depending on what extraction source each Sourced Paper come from).

Since the unmatched references was not a case that was a concern back when we used the legacy endpoint, I figured that we probably don't need to care about them here anyway.

codeviking: I see a case for raising an exception, in that it lets the caller decide if they care.

But I also see the case for not raising an exception, as it means we have to further digest the code we're changing (which we haven't changed in awhile). We have to figure out where the associated code is called, and handle the exception we're adding (to preserve the current behavior).

I favor the latter because it's less work. But I'd consider adding a log message whenever we skip a reference because of this, so that we know something is being filtered out at runtime. That way if someone is investigating something that's related, the log message might help them deduce why quickly. I'd also make a note explaining this. You could even link to this discussion thread.

FWIW it's surprising to me that the API returns references that aren't in-corpus. But that's probably a lot more context that plays into that decision. I wonder if we could add a parameter to the API that lets people filter the response to only include in-corpus things `?discard_nonexistant=1` or something.

stupendousC: > FWIW it's surprising to me that the API returns references that aren't in-corpus. But that's probably a lot more context that plays into that decision. I wonder if we could add a parameter to the API that lets people filter the response to only include in-corpus things ?discard_nonexistant=1 or something.

interesting, yeah IDK the new Public API endpoint does that, I think we didn't realize the difference at the time.  But yeah, I will check with that team and see if it's soemthing they're interested in

stupendousC: > I favor the latter because it's less work. But I'd consider adding a log message whenever we skip a reference because of this, so that we know something is being filtered out at runtime. That way if someone is investigating something that's related, the log message might help them deduce why quickly. I'd also make a note explaining this. You could even link to this discussion thread.

Oh yeah... a log would be useful, will add!

kyleclo (approved):
lgtm - minor comment about the logging

data-processing/common/commands/fetch_s2_data.py
wammar: @kyleclo do you mind adding suggestions for the public API [here as a feature request from S2 research](https://docs.google.com/document/d/1Qtc-vdMv5RA6RRXtyriegLg6u21OiSgz7k8osIUykJU/edit#bookmark=id.f227a6gjy21k) so I can track and prioritize it as needed for the following quarters? I already added a placeholder but it'd be great if someone who experienced the user pain around this can elaborate.

PR #382, opened 2023-03-28
Deprecate partner subdomain

stupendousC:
* Parent task = [Switch scholarphi to new public API endpoint #32158](https://github.com/allenai/scholar/issues/32158#top)
  * As seen in [comment/question](https://github.com/allenai/scholar/issues/32158#issuecomment-1456799584), we have 2 goals:
    * GOAL 1 = use the latest `graph/v1/` endpoint instead of legacy `v1/` endpoint
    * GOAL 2 = use the regular `api.semanticscholar.org` subdomain, instead of an optional `partner.semanticscholar.org` subdomain.
      * Question: before, users with `S2_PARTNER_API_TOKEN` env were calling the partner domain for higher rate limit.  Can we use the same token on the regular `api.semantic.org` subdomain?   
        * [Rodney says yes](https://github.com/allenai/scholar/issues/32158#issuecomment-1487492642).

* THIS PR is for the `chi-2021-demo` branch.
  * Goal 1 is already done in this PR: #377 
  * Goal 2 is the change we're making here.

-------
also plz ignore the branch name `gh35326_chi2021demo` where it referenced the wrong parent ticket lol

TESTING 👍 
------
* Deploy locally, then run pipeline on an example paper.  
  * Basically the same steps seen in IDEA 2C in this [ticket](https://github.com/allenai/scholarphi/pull/377))
  * **Do this WITHOUT a partner API key**, logs should show the same.
    * This is prob not necessary but I'm gonna test it out just in case. 
    * Yup, I see the same results as the reference ticket above! (The log starting with `>>>` is temporary and is not in PR)
    * Screen Shot 2023-03-29 at 12 19 38 PM
  * **Do this WITH a partner API key**, logs should show that it's being sent to the `api.semanticscholar.org` subdomain.
    * Lots of existing partner API keys available, I'm just going to borrow `[s2-partner-conference-api`](https://us-west-2.console.aws.amazon.com/apigateway/home?region=us-west-2#/api-keys/m6l2m75vif) for now. 
    * Yup, I see that it's no longer calling the partner subdomain!
    * Screen Shot 2023-03-29 at 12 23 46 PM
    * Looked at the `data-processing/data/citations.json` produced and it's legit!

### DIFF
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -120,11 +120,7 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = (
-    apiKey
-      ? "https://partner.semanticscholar.org/v1"
-      : "https://api.semanticscholar.org/v1"
-  );
+  const apiOrigin = "https://api.semanticscholar.org/graph/v1";
 
   return axios.get<Paper>(`${apiOrigin}/paper/${s2Id}`, conf);
 }
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -120,11 +120,7 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = (
-    apiKey
-      ? "https://partner.semanticscholar.org/v1"
-      : "https://api.semanticscholar.org/v1"
-  );
+  const apiOrigin = "https://api.semanticscholar.org/v1";
 
   return axios.get<Paper>(`${apiOrigin}/paper/${s2Id}`, conf);
 }
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -120,7 +120,11 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = "https://api.semanticscholar.org/graph/v1";
+  const apiOrigin = (
+    apiKey
+      ? "https://partner.semanticscholar.org/v1"
+      : "https://api.semanticscholar.org/v1"
+  );
 
   return axios.get<Paper>(`${apiOrigin}/paper/${s2Id}`, conf);
 }
--- a/api/src/s2-api.ts
+++ b/api/src/s2-api.ts
@@ -120,7 +120,11 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = "https://api.semanticscholar.org/v1";
+  const apiOrigin = (
+    apiKey
+      ? "https://partner.semanticscholar.org/v1"
+      : "https://api.semanticscholar.org/v1"
+  );
 
   return axios.get<Paper>(`${apiOrigin}/paper/${s2Id}`, conf);
 }
--- a/data-processing/common/commands/fetch_s2_data.py
+++ b/data-processing/common/commands/fetch_s2_data.py
@@ -25,7 +25,7 @@ class S2ApiException(Exception):
 
 class S2ApiRateLimitingException(S2ApiException):
     """
-    Caller has been rate-limited by S2's Public/Partner API.
+    Caller has been rate-limited by S2's Public API.
     """
 
 class S2MetadataException(S2ApiException):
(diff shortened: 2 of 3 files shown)


api/src/s2-api.ts
```diff
@@ -120,11 +120,7 @@ export function getPaperUncached(s2Id: string, apiKey?: string): AxiosPromise<Pa
   if (apiKey) {
     conf.headers['x-api-key'] = apiKey;
   }
-  const apiOrigin = (
```
ca16, line 123: I wouldn't change anything in the `api` directory. I don't have a great idea of what goes on here, and we (reader data side) don't really maintain it - we just focus on the data-processing dir.

data-processing/common/commands/fetch_s2_data.py
```diff
@@ -59,7 +59,6 @@ def _mk_api_request(self, arxivId: ArxivId) -> requests.Response:
         headers = None
 
         if self._partner_api_token:
-            base_url = "partner.semanticscholar.org"
```
ca16, line 62: Don't we need to replace this with something?

ca16 (commented):
A couple of comments, but more broadly wondering about the plans for testing...

data-processing/common/commands/fetch_s2_data.py
stupendousC: no, it'll just stay as the regular "api.semanticscholar.org" on L58.

api/src/s2-api.ts
stupendousC: k, i'll revert this change!

stupendousC: OR... maybe @mjlangan can shed some light here?

mjlangan: No idea what that's used for.

stupendousC: Michael says:
Screen Shot 2023-03-28 at 3 20 45 PM

stupendousC:
Will absorb the totally cool code changes Sam suggested in PR #381 .  Will also not leave `s2-api.ts` alone after all since the `main` owners are cool with it, it's not like anyone actually uses it anyway, but it's probably best to keep both branches as consistent with each other as possible?

ca16:
If you're excited to do the api changes too, feel free! Two things about that:
1) I wonder if it would be better to do a separate PR for data-processing code, and a separate PR for api code? What do you think?
2) I don't think I'm equipped to review the api changes. Sounds like maybe it doesn't matter for this case given the review on the PR for main, and the fact that I'm not sure anyone uses the chi-2021-demo branch's api code, but just wanted to make sure that was clear...

ca16:
Oh also! Still wondering about the testing plan for the data-processing changes...

stupendousC:
> Oh also! Still wondering about the testing plan for the data-processing changes...

Yup, will get on that!  stay tuned

stupendousC:
> 1. I wonder if it would be better to do a separate PR for data-processing code, and a separate PR for api code? What do you think?
> 2. I don't think I'm equipped to review the api changes. Sounds like maybe it doesn't matter for this case given the review on the PR for main, and the fact that I'm not sure anyone uses the chi-2021-demo branch's api code, but just wanted to make sure that was clear...

hmmm yeah, I see what you're saying.  Ok i'll move the `s2-api.ts` changes to another PR so it'll be easier to figure out if anything goes wrong on the API side.

ca16 (commented):
Looking pretty good! Two more questions:
1. How does the output produced from your runs on this branch compare to output for the same paper produced from a run on the chi-2021-demo branch as it is rn?
2. Did running pytest go alright? (just a reminder there's also that TC job that will run it for you)

stupendousC:
> 1. How does the output produced from your runs on this branch compare to output for the same paper produced from a run on the chi-2021-demo branch as it is rn?
> 2. Did running pytest go alright? (just a reminder there's also that TC job that will run it for you)

| | Current `chi-2021-demo` branch  | This branch, WITH API key | This branch, WITHOUT API key |
| ------------- | ------------- | ------------- | ------------- |
| 1. compare resulting `citations.json` | 🎯  |match 👍   |match 👍   |
| 2A. `pytest`  | 🆗   | 🆗  | 🆗 |
| 2B. `pytest --all` locally | fail but 🆗 | fail but 🆗 | fail but 🆗 |
| 2C. `pytest --all` via [TC](http://s2build.inf.ai2/buildConfiguration/SemanticScholar_Contin_Data_ScholarphiCi/2386392) | 🆗  | WIP  | <-- |

* Comparing resulting `citations.json` is a slight headache, bc each obj therein may not be present in the exact same order.  
* 2B. `pytest --all` failed same 7 `definition` tests on here, but this is a known issue (as [previously seen](https://github.com/allenai/scholarphi/pull/377)), nothing to do with me.
* 2C.  Oops I went with the [wrong TC job](http://s2build.inf.ai2/buildConfiguration/SemanticScholar_SparkCluster_Timo_ScholarPhi_BuildNewScholarphiImage/2386250) at first. Thanks @ca16 !!

ca16 (approved):

